PR #1, opened 2019-01-21
Schema schema!

warpfork:
The title should say it all -- herein we have a self-describing schema :)

There's a lot of commits in between the start and this result, almost all with long commit messages that are worth reading if you want to follow the intermediate decision making in detail.  I won't reproduce all of them here -- the commit log exists, in its wonderful decentralized form -- just want to draw the reader's attention to the fact there is in fact text there, and it's meant for consumption.

Short highlights:

- This is a schema!
  - [There's already some docs in another file over at `doc/schema.md`](https://github.com/ipld/go-ipld-prime/blob/schema-schema/doc/schema.md).
- It's designed around "structural typing" -- being aimed at serial data, that is our defacto reality.
  - (But take note: you can get "nominative typing"-style behaviors by using Unions!)
- The basic kinds of types are those from the IPLD Data Model -- strings, ints, bytes, maps, lists, etc.
- We add "struct" (product types) and "union" and "enum" (sum types) kinds to the mix in addition to the IPLD Data Model kinds.
  - These all define "representation" properties -- which bidirectionally map them into one of the basic Data Model kinds.
- There are limits around nullability and optionality of fields.  The defaults are conservative: nulls are *not* allowed and all fields in structs are *required* unless otherwise declared.

Included in these files are two sets of examples:

1) a short example (using mostly anonymous nested types, to prove those out)
2) the schema itself!

... and each in two formats:

1) a JSON document, conforming to the schema schema
2) an equivalent DSL (which as you can see tends to be much, much shorter).

The DSL-form schema schema file contains a large amount of docs; **read those for the meat of what's going on here**.  (The actual declarations are only about 112 lines; so that's another 220 or so of pure docs.)

We should probably add fields for "`documentation optional String`" to each of the types (or, hey, speaking of which, maybe a mixin system for when lots of types have the same property like that?); I didn't do so here, mostly because it wouldn't be a pleasant effect on the readability of the JSON files.

---

There was a jaunt in the middle where a couple of different formats for representing nullable and optional fields were trialed; the ones that weren't selected have been merge-ignored.  This image shows how each of the options would've shaken out:

schema-side-by-side2

Long story short, I went for the representation that yields the document which is both shortest and least-deeply-nested.  (It's also a particularly amusing image for showing how compact the DSL can be compared to the fully-normalized JSON equivalent.)  See the per-commit messages for additional detail.

---

A great deal of the implementation for wiring this up to real working code is already present in the `typed/system` package.  The first take on coding `typed/declaration` is currently out of sync with where this schema doc came to rest, and needs updating (specifically, to support TypeTerm and recursive inline definitions), but should be short work.

The real dream, of course, would be to take this schema; run codegen on it, to get nice native types in golang code which match the schema semantics... and use those as essentially the entirety of all the AST code of the `typed/declaration` package.  But that'd be a bit of a bootstrapping cycle issue ;) so we'll have to do a *little* more turning of the crank manually to get this all landed.

### DIFF
--- a/typed/declaration/examples.ipldsch
+++ b/typed/declaration/examples.ipldsch
@@ -1,0 +1,11 @@
+
+type ExampleWithNullable map {String : nullable String}
+
+type ExampleWithAnonDefns struct {
+	fooField optional {String:String}
+	barField nullable {String:String}
+	bazField {String : nullable String}
+	wozField {String:[nullable String]}
+} representation map (
+	fooField: alias="foo_field"
+)
--- a/typed/declaration/schema-schema.ipldsch
+++ b/typed/declaration/schema-schema.ipldsch
@@ -46,7 +46,7 @@ type SchemaMap map {TypeName:Type}
 ## }
 ## ```
 ##
-type Schema union = {
+type Schema union {
 	| SchemaMap "schema"
 } representation keyed
 
@@ -75,7 +75,7 @@ type Schema union = {
 ## }
 ## ```
 ##
-type Type union = {
+type Type union {
 	| TypeBool "bool"
 	| TypeString "string"
 	| TypeBytes "bytes"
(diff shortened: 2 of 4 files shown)


warpfork:
Incidentally, I just love trying to describe in English what's going on here.

> "The 'representation' property of the UnionRepresentation type (which a 'union' kind) is itself a 'keyed'-representation union."

Whew!

warpfork:
A couple of other ideas that came up while doing the final update to the schema-schema json...

- We could make `TypeStruct.representation` an optional field.
	- It's *clearly* "map" if you don't specify something in the DSL.  Might as well be considered so clear in the JSON.
	- Notice how *every one* of the instances so far has been a terse `"representation": {"map": {}}`.  The total size of the schema-schema JSON goes from 300 lines down to 260 if these are elided.
- We could make TypeStruct.fields optional.
	- This comes up less, but... a struct of no fields is certainly valid.
- We could have a terser encoding of "unit" type (e.g. the type with no properties).
	- We could call it `unit` in the DSL.
	- We could either continue to internally represent it as struct no fields and properties,or we could represent it as a type called 'unit'.  Is there any substantive importance in the distinction?  (I can't think of one, other than arguably clarity of intent.)

There's also a big open review question about three-valued logic versus optionality on struct fields: namely, what if I want an `optional Bool` field, but still want two-valued (rather than three-valued) logic?  This requires more definition of how exactly we handle "undefined" as a value, which is currently... rather undefined.  Or at least under-defined.  (Example: how do I suggest to a `typed.Node` instance that I want a field to be undefined?)

These are all things I'm happy to consider further later, though.  What's pushed here should be enough to iterate on for a while.

typed/declaration/schema-schema.ipldsch
```diff
@@ -0,0 +1,358 @@
...
+## The values of the list have some specific type of their own.
+##
+type TypeList struct {
+	valueType TypeTerm
+	valueNullable optional Bool
+}
+
+## TypeLink describes a hash linking to another object (a CID).
+##
+## REVIEW: this currently has no details... but possibly it should have a
```
lanzafame: I am not aware of the implications of adding a type hint but from work I have done in the past in with [FHIR](http://hl7.org/fhir/) which similarly can link/embedded many different types with others, a type hint would have been very beneficial.

typed/declaration/schema-schema.ipldsch
```diff
@@ -0,0 +1,358 @@
...
+##   "schema": {
+##     "MyFooType": {
+##       "type": "string"
+##     }
+##   }
+## }
+## ```
+##
+type Schema union {
+	| SchemaMap "schema"
```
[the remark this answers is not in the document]
warpfork: Yup.  Pipe is a visual cue that we're in a union or enum.

(A lot of the syntactical choices thus far are fairly arbitrary; the parse could even be unambiguous without this.  It's more or less there because I liked the drafts better that made sum types (union, enum) look strongly visually distinct.  And probably, subconsciously, the Pipe character is coming from either Haskell, [OCaml variants](https://v1.realworldocaml.org/v1/en/html/variants.html), or [GraphQL unions](https://graphql.org/learn/schema/#union-types).)

typed/declaration/schema-schema.ipldsch
warpfork: :+1: :+1: :+1:  I think we should definitely have a type hint here... I'm only leaving it blank this far for lack of a fully finished idea of how exactly that should work.

Is it as simple as adding a `targetType optional TypeName` property?  Or should it be `TypeTerm` (so it can link directly to anonymous map types)?  Are there any *more* properties that could be needed?

In the DSL, this could perhaps be like `type Foo struct { fieldName optional Link<TargetType> }` -- spending (hopefully the last of) our sigil complexity budget for those angle brackets.

typed/declaration/examples.ipldsch
```diff
@@ -0,0 +1,11 @@
+
+type ExampleWithNullable map {String : nullable String}
```
whyrusleeping: One request, feel free to tell me its a terrible idea, but It might be nice if the schema could always be compacted down into a single line.

typed/declaration/schema-schema.ipldsch
whyrusleeping: +1 to type hints on links. I think that’s super important.

typed/declaration/examples.ipldsch
warpfork: I think that's slightly nuts but if you want to imagine a regexp that replaces newlines with semicolons, I guess sure.

(Kidding.  I mean "parser that regards semicolons and linebreaks as the same".  Don't quote html+zalgo at me.)

warpfork:
I think I'm going to mark this as "merged" and keep iterating.  Thanks for all the comments so far (both on and off github), keem 'em coming :dancer: 

There's still *definitely* imperfections and issues in this: in particular, I think "optional" as currently specified is outright *wrong* and there's actually two closely related features that it should be split out into (namely: optionality, which *does* do membership-cardinality-plus-one; and defaults (or perhaps "representAs"), which *don't*; and the former is part of the type definition, while the latter is part of the representation definition).  But that's something for handling in another round of diffs.

warpfork:
Anyone still using this PR link as your entry point for design review, some updates on Optionals -- previously, there were some open questions:

> There's also a big open review question about three-valued logic versus optionality on struct fields: namely, what if I want an `optional Bool` field, but still want two-valued (rather than three-valued) logic? This requires more definition of how exactly we handle "undefined" as a value, which is currently... rather undefined. Or at least under-defined. (Example: how do I suggest to a `typed.Node` instance that I want a field to be undefined?)
> 

This is now addressed by splitting the two concepts: we can have `optional` declared on a field, which adds an "undefined" value to considerations; and a `default` property defined as part of the representation, which doesn't.

`optional` is considered a property of the type definition; whereas `default` values are defined in the *representation* rather than the *type* because of the (lack of) impact on type membership cardinality.

More details in this commit with that work: https://github.com/ipld/go-ipld-prime/commit/4e3c8def80f6be413980d0b5f7be6be3eb9268e2

PR #3, opened 2019-03-12
Docs

warpfork:
Docs.  They're important.

Posting as PR so the diff can be linked easily (but will probably be merging this fairly promptly because it's mostly backfill for stuff that's already pretty settled.  Comments still welcome; I'll address them in more patches if they come in after I merge).


doc/big-picture.md
```diff
@@ -0,0 +1,67 @@
...
+IPLD Schema bring common concepts like "structs" with named "fields", as well
+as "enums" and even "unions" -- and make clear definitions of how to map all of
+these concepts onto the basic serializable types of the Data Model.
+
+You can use IPLD at any of these levels you choose.  We recommend having a
+schema for your data, because it makes validity checking, sanitization,
+migration, and documentation all easier in the long run -- but the Data Model
+is also absolutely usable stand-alone.
+
+IPFN is mentioned here as an intent to design something between "dependent types"
```
rvagg: I think this is the only mention of this here (aside from the list at the top), is there a link or an expansion on the acronym you can provide to give some clarity? It's not clear to me what "not to expect" from these docs from this paragraph alone.

doc/big-picture.md
```diff
@@ -0,0 +1,67 @@
...
+
+Three (or four) big picture groups of concepts you'll want to keep in mind when
+reading the rest of the docs are nodes, operations, encoding, and linking.
+
+Nodes: every piece of data in IPLD can be handled as a "node".
+Strings are nodes; ints are nodes; maps are nodes; etc.
+[Nodes have their own full chapter of the documentation](./nodes.md).
+
+Operations: some basic computation we can do *on* a node is an operation.
+Traversals and simply updates have built-in functions; more advanced operations
```
rvagg: *simple?

doc/big-picture.md
warpfork: Um, I see your point and will take suggestions for wording.

"IPFN" isn't a... thing that... exists.  There's been some floating discussion of "what if we made a functional programming language" and "what if WASM" and sometimes it floats into territory I'd file as "what if dependent types".  I wanted to give a tag to all that mostly to say "this ain't it".

It's possible this just isn't a good place in the docs for that.  Maybe the whole '3.' should be dropped and s/several/two/ ?

warpfork: There's some mention of non-turing-completeness and not-dependent-types in the schema page already, so that would also make this seem more droppable.

doc/big-picture.md
warpfork: Thaaank you!  precise eyes :heart:

doc/big-picture.md
rvagg: yeah, I'd say drop it, it's quite a tangent and maybe deserves a more complete write-up of the ideas in a different place

warpfork: sold :+1:

PR #4, opened 2019-03-16
Linking redux

warpfork:
Massive rewrite of all components having to do with linking.

Linking is now primarily defined by a set of interfaces in the root ipld package: `Link`, `LinkBuilder`, `Loader`, and `Storer`.  These interfaces cover the general concept of what it means to link things and be referencable by link.  (Older generations of LinkLoader, LinkBuilder, and LinkContext types which lived in either the 'repose' or 'traversal' packages are now gone, but their concepts should be recognizable in the new stuff).

Cid itself disappeared from almost all concrete uses.  For example, traversal.TraversalProgress.LastBlock now uses Link instead of Cid.  `go-ipld-prime//linking/cidlink` provides implementations of `ipld.Link` which use `cid.Cid` (and `ipld.LinkBuilder` is roughly equal to `cid.Prefix`, etc).  You can grep for "cid" throughout the project now, and (aside from comments) the only concrete references are bounded to the `linking/cidlink`, `encoding/dagcbor`, and `encoding/dagjson` packages.  This is neat.

And `encoding/dagcbor` and `encoding/dagjson` now actually parse and emit links in the special forms of those formats!  Of course, these concretely reference `cid.Cid` since that's what they parse.

A general `encoding` feature set remains in that package, and works... except it doesn't know how to use Links of any kind.  But it does allow use of go-ipld-prime to do serialization work for anything *not* using Links *without* importing the `cid` packages, which is pretty cool.  (Future hope: we'll also be able to introduce other kinds of linking features which are defined by *schemas*, which we will be able to then use in these general encodings.  But that's a fairly long shot into the future.)

The 'repose' package pretty much got nuked in the course of all this.  It was a half baked first draft that accumulated a junkdrawer of badly segmented concepts.  Good riddance.

Multicodecs are now handled via a registry system which is confined in scope to the cidlink package.  (It's global, but I think in practice this will be fine: it's a plugin system, and there's no good cause for allowing variations in how those magic bytes of cids are interpreted at runtime: only whether or not a plugin is built into a program or not as a whole.)


warpfork:
Multicodec registration is done as a package init function.  I'm not exactly thrilled by using such a mechanism, but I debated doing explicit user-controlled registration instead, and it seems strictly worse.  As much as I dislike side-effecting package imports, it does have the upside of making it valid for a program to contain two different packages which each use the import trigger and they won't conflict nor double-register.

hannahhoward:
A few questions about this:

1. How to I build an ipld.Link from just a CID (no block present)? 
2. And how do I get a CID back from an ipld.Link? 

Unfortunately go-graphsyncs wire protocol is for sure CID based.

I guess I see the value of an abstrack link type, but it presents several additional challenges in terms working with it for large systems (go-filecoin / go-ipfs) that have CID as the link type as an assumption throughout the codebase. I'm not opposed but for migration not to be really tough I feel like we need an easy CID -> ipld.Link -> CID layer. (with, sadly, the requisite typecast unfortunately)

hannahhoward:
Ignore comment... I now see all these interfaces and fields are public.

PR #9, opened 2019-04-11
Fix schema examples

rvagg:
Testing out my parser on your example, but I think it's out of sync with the schema so I guess it was an older iteration. This doesn't need to be merged but I have questions, inline.

### DIFF
--- a/typed/declaration/examples.ipldsch
+++ b/typed/declaration/examples.ipldsch
@@ -1,4 +1,3 @@
-
 type ExampleWithNullable map {String : nullable String}
 
 type ExampleWithAnonDefns struct {
@@ -6,6 +5,4 @@ type ExampleWithAnonDefns struct {
 	barField nullable {String:String}
 	bazField {String : nullable String}
 	wozField {String:[nullable String]}
-} representation map (
-	fooField: alias="foo_field"
-)
+}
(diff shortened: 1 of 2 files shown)


typed/declaration/examples.ipldsch
```diff
@@ -1,11 +1,8 @@
-
 type ExampleWithNullable map {String : nullable String}
 
 type ExampleWithAnonDefns struct {
 	fooField optional {String:String}
 	barField nullable {String:String}
 	bazField {String : nullable String}
 	wozField {String:[nullable String]}
-} representation map (
-	fooField: alias="foo_field"
```
rvagg: I gather aliases aren't something that landed in the latest schema, also the `:` and `=` characters on this line seem out of place, or was there some partly developed ideas here?

typed/declaration/examples.ipldsch.json
```diff
@@ -4,53 +4,51 @@
 			"kind": "map",
 			"keyType": "String",
 			"valueType": "String",
-			"valueNullable": true
```
rvagg: I think I have this wrong, TypeMap uses "valueNullable" so I guess a "nullable" becomes "valueNullable" in that case

typed/declaration/examples.ipldsch.json
```diff
@@ -4,53 +4,51 @@
...
+			"nullable": true,
+			"representation": {
+				"map": {}
+			}
 		},
 		"ExampleWithAnonDefns": {
 			"kind": "struct",
 			"fields": {
 				"fooField": {
-					"valueType": {
```
rvagg: as a struct, I think this should be "type" as "valueType" is for maps isn't it? coming from StructField->type

PR #11, opened 2019-06-17
Fix doc

matthiasbeyer:
The iteration can stop if the `Done()` method returns true, not false.

---

I just noticed that when scrolling the docs, I did not look at code for this.


warpfork:
Sharp eyes :)  Thank you!

PR #12, opened 2019-06-23
Selector implementation!

warpfork:
This is a first pass at some selector implementation...!  (At long last.)

Corresponding updates to the descriptions of selectors in the ipld/specs repo will be coming soon; the schema has changed somewhat significantly.  (I've got drafts of this change as well, but need to finish updating all the examples.)

Comments in the body should be fairly sufficient to grok if you already grok schemas and you've seen some of the more recent stuff -- e.g. the `ParseSelector` function is basically a manually-rolled implementation of unpacking a "keyed union" representation.

So far there's just outlines of implementations of a few of the basic ones.  The Selector interface as a whole sprouted a few more methods better distinguish from exploration vs matching, and to provide a way for selectors to hint that they're only going to need to traverse *some* specific fields rather than *all* (which may be consequential for large maps).  More implementations of other specific selectors still needed, but hopefully the interface is close to the mark now.

This has already been hanging out for a while now, as I've been bouncing around on other priorities, so I'm basically just going to merge it -- the tests so far pass! -- and then expect that when it's picked up later, there might be some significant refactors, and that's fine.

### DIFF
--- a/path.go
+++ b/path.go
@@ -67,6 +67,16 @@ func (p Path) Join(p2 Path) Path {
 	return p
 }
 
+// AppendSegment is as per Join, but a shortcut when appending single segments.
+func (p Path) AppendSegment(ps string) Path {
+	l := len(p.segments)
+	combinedSegments := make([]string, l+1)
+	copy(combinedSegments, p.segments)
+	combinedSegments[l] = ps
+	p.segments = combinedSegments
+	return p
+}
+
 // Parent returns a path with the last of its segments popped off (or
 // the zero path if it's already empty).
 func (p Path) Parent() Path {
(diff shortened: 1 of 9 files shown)


warpfork:
Also caveat to the reader: there may be a mangle of inconsistent terminology about "selector" vs "matcher" vs "explorer".  Those terms and concepts have been evolving throughout the code.  I haven't done a final pass again lately to make sure all references here are up to date.

The latest is: selector is a union of all explorers and a matcher; explorers just explore (in other words,  their "Decide" method just flatly returns false -- and now that I think about it, maybe we should make a small mixin struct that just Does That); and matcher is a single thing that's a tad like the comparable concept in regexps.  (And if an application wants mid-tree matches *and* to continue exploration, one does so with an exploreUnion.)

Simple, right?  :woman_shrugging:

warpfork:
Mergin'!  (But like I said, that doesn't mean this is "done" -- just that I wanna keep the commit graph from fanning out too far.)

PR #13, opened 2019-07-07
Add Dependency Management Via Go Modules, Adds CI (Travis, CodeCov), Satisfy Go Vet

postables:
Add Dependency Management Via Go Modules, Adds CI (Travis, CodeCov), Satisfy Go Vet. Also removes the git submodules since they aren't needed with go modules :D


hannahhoward:
+1 can we merge this @warpfork ? (I guess I have merge ability but I'm still deferring to you)

warpfork:
This is a lot of things at once.

- I'm more than happy to have `go mod` support -- so long as someone's actually going to exercise it.  (For better or for worse, I simply don't, and it would be improper for me to claim otherwise.)  Sounds like someone will, so... :+1:   (Side note: it's not actually clear to me why downstream projects can't just... do the thing that go mod usually does with transitive dependencies?)

- I'm not happy to remove git submodules.  They work, and I use them, and I expect to continue to use them.  Git submodules and a minimal GOPATH setup have worked since go 1.0.  I *like* things with a breakage halflife measurable in decades.  Their nature as a standard composable merkle tree also makes them play well with some other tools that are fairly essential to my sanity.  I am happy to support Go Modules but please leave the existing dependency management tooling, as it works well and in any case can co-exist happily with Go Modules.

- I don't like most of the changes `go vet` proposed.

- I'm happy to have travis CI config added.  :bowing_man: 

- I'm indifferent to codecov, but would certainly merge it.  (Having it is nice; having automatic comments on github is...  ehh.  It's, uh, some kind of a UX, I guess.  But maybe it's better than not having it!)

I don't think I'd merge this as-is.  But I'd merge several of the individual constituents in a heartbeat!

hannahhoward:
closing @postables but some of your commits are included in #15 . Pretty sure we will not remove .gopath and we aren't putting go vet in CI. But we've adding module support + CI + codecov. Thanks for your contribution.

postables:
@warpfork @hannahhoward all good point, thanks for adding module support it makes it much easier to use in other projects :rocket:

PR #17, opened 2019-07-16
Update to selector spec, minus recursive selectors

hannahhoward:
# Goals

Bring implementation of selectors in line with what is listed at https://github.com/ipld/specs/blob/master/selectors/selectors.md - minus ExploreRecursive and ExploreRecursiveEdge (plus Conditions which are still in flux)

# Implementation

- Modify SelectFields to ExploreFields, and update to spec
- Modify SelectTrue to Matcher, and add parsing
- Introduce ExploreIndex and ExploreRange, along with parsing
- Modify SelectAll to ExploreAll, update to spec, add parsing
- Modify SelectUnion to ExploreUnion, update to spec, add parsing, add implementation of Explore
- Add unit tests for all implemented selectors

# Not Covered (Future PRs)

- Link loading
- Updating TraverseInformatively to handle Lists and respect limited values returned by Interests()
- SelectorBuilder (want to implement for convenience of making selectors)
- ExploreRecursive + ExploreRecursiveEdge

# For Discussion

- Are unit tests too verbose / exhaustive (making them potentially brittle to change?)
- Note check on validity of ExploreRange end must be > start -- do we want to enforce this?
- Note checks on list type of node for Explore method of ExploreIndex & ExploreRange -- is this the right behavior? Does explore validate that there will be a returned selector in the context of the node? Or just the PathSegment? If the former (which I assume) I wonder if this needs range checks as well. (i.e. the index of the path segment does not go past the length of the list in the node?)
- Should Union selectors attempt to dedup interests? (assuming no high-cardinality)?
- Should Union selectors attempt to dedup returned selectors (i.e. [Matcher{}, Matcher{}])?


rvagg:
> Are unit tests too verbose / exhaustive (making them potentially brittle to change?)

NO! Never!

traversal/selector/exploreAll.go
```diff
@@ -0,0 +1,44 @@
...
+	ipld "github.com/ipld/go-ipld-prime"
+)
+
+// ExploreAll is similar to a `*` -- it traverses all elements of an array,
+// or all entries in a map, and applies a next selector to the reached nodes.
+type ExploreAll struct {
+	next Selector // selector for element we're interested in
+}
+
+// Interests for ExploreAll is empty (meaning traverse everything)
```
warpfork: This is a bit nitty of me, but I like to be very precise about saying "nil" and "zero" (or "empty" for lists and maps, I suppose, but then meaning "not nil").  This one being nil, of course.

warpfork:
One thought on the testing is that you could actually just write a bunch of JSON, for both the data fixtures and for the test selectors.  

Those might be nice fixtures for cross-language test fixtures later, too.

The tests you've written are awesome though.  Just suggesting it as an option.

traversal/selector/exploreIndex.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+
+	ipld "github.com/ipld/go-ipld-prime"
+)
+
+// ExploreIndex traverses a specific index in a list, and applies a next
+// selector to the reached node.
+type ExploreIndex struct {
+	next     Selector      // selector for element we're interested in
+	interest []PathSegment // index of element we're interested in
```
warpfork: This is... a slice that's always of length one?

It's a bit wizard, but we might actually declare this as `interest [1]PathSegment`, and then have the `Interests()` func return `s.interest[:]` -- the latter syntax skooshes it into a slice, while the fixed length declaration means it can get inlined into the struct rather than needing another malloc.

(Might not matter too much.  But I've become a bit of a malloc hawk whenever it's not too much of a stretch to nab the wins.)

warpfork:
For a while, I was going to comment that maybe we can get rid of `nil` Selector values entirely -- everything should terminate in a `Matcher` nowadays; there are no nullable Selectors in the AST schema!

But on further reading, I see they're still being used in a couple of other cases, like to say "nope!" when an index selector is applied to non-list data.  And that does seem to make sense.  :+1: 

Maybe would be nice squeeze a reminder comment in about that somewhere.  Seems like the kind of detail I'd be prone to forget and then scratch my head and question again 6mo from now (even though it makes perfect sense after a long enough look).

traversal/selector/exploreIndex.go
warpfork: (... I'm not actually sure if that will survive escape analysis, but I think it's got a good shot.  Even if not, the fixed length is a nice bit of self-documenting, IMO.)

traversal/selector/exploreRange.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, err
+	}
+	x := ExploreRange{
+		selector,
+		make(map[int]struct{}, endValue-startValue),
+		make([]PathSegment, 0, endValue-startValue),
+	}
+	for i := startValue; i < endValue; i++ {
+		x.interest = append(x.interest, PathSegmentInt{I: i})
+		x.hasSelector[i] = struct{}{}
```
warpfork: Ok, so we've kinda screwed for the interest set on this one by the interface (owch), but I'd definitely go for just storing the two ints and doing the range check again in the Explore method rather than a map.

traversal/selector/exploreUnion.go
```diff
@@ -0,0 +1,95 @@
...
+func (s ExploreUnion) Interests() []PathSegment {
+	// Check for any high-cardinality selectors first; if so, shortcircuit.
+	//  (n.b. we're assuming the 'Interests' method is cheap here.)
+	for _, m := range s.Members {
+		if m.Interests() == nil {
+			return nil
+		}
+	}
+	// Accumulate the whitelist of interesting path segments.
+	// TODO: Dedup?
```
warpfork: I'd go a step farther and say they probably ought to be sorted.  Which makes the deduping easier; also, though, because if we're then seeking over a large map that's sorted, we'll tend to get some wins thusly.

I'm fine with merging this without that, though.

traversal/selector/exploreRange.go
hannahhoward: ok. fair. is that a problem for ExploreUnion, which pretty explicitly says "assumes interests isn't a big computation?"

traversal/selector/exploreUnion.go
hannahhoward: Yea I'd like to put dedup on the TODO list.

traversal/selector/exploreRange.go
hannahhoward: This comment -- https://github.com/ipld/go-ipld-prime/pull/17/files/c590adc105460366fffaf07318b86b03f528cb67#diff-0f5c839144c4c664fcef0ebf75f3eb79R25

traversal/selector/exploreIndex.go
hannahhoward: I like it. will do.

traversal/traverse_test.go
```diff
@@ -86,10 +92,18 @@ func TestTraverse(t *testing.T) {
...
-				mb.Insert(knb.CreateString("nested"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
-					mb.Insert(knb.CreateString("f"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
-						mb.Insert(knb.CreateString("nonlink"), vnb.CreateBool(true))
+				mb.Insert(knb.CreateString("f>"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
+					mb.Insert(knb.CreateString("foo"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
+						mb.Insert(knb.CreateString("."), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {}))
+					}))
+					mb.Insert(knb.CreateString("nested"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
+						mb.Insert(knb.CreateString("f"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
+							mb.Insert(knb.CreateString("f>"), vnb.CreateMap(func(mb fluent.MapBuilder, knb fluent.NodeBuilder, vnb fluent.NodeBuilder) {
```
warpfork: Hm.  I think you commented on this in another medium already, but you're right: this stutter kind of sucks, here.

I wonder if we should amend the schema over over in the specs repo to say just `type ExploreFields {String:Selector}` instead of `type ExploreFields struct { fields {String:Selector} (rename "f>") }`.  It doesn't seem like we're going to come up with any other properties to put into that struct.

But this is a comment for the schema.  This code DTRT :+1:

traversal/selector/exploreRange.go
warpfork: yeah, that's what I mean by "we're kinda screwed for the interest set" -- I agree there's not an obvious way to avoid pregenerating that slice.

The map is pretty easy to ditch though, I think?

warpfork:
This is... awesome.  :rocket:   I went through it with a pretty fine-toothed comb and my only comments are tiny nits.

This is good to merge whenever you want :D

PR #19, opened 2019-07-17
Builder for selectors

hannahhoward:
# Goals

Add a way to quickly assemble mostly type safe selectors until IPLD schema code generation is finished.

# Implementation

Adds a builder class to quickly assemble selectors nodes, that enforces the underlying
schema. This allows quick assembly of selector nodes. The builder makes a SelectorSpec, which can be converted to an ipld.Node or an actual parsed Selector.

This will likely be removed once IPLD Schema code gen is finished, but for the time being, given this library may be in use very shortly, it doesn't make a lot of sense to wait, and I am anxious about handing the schema over to folks not familiar with selectors and telling them to use it!

Notes: I realize this PR contains a brand new interface and style which means it might be open to more debate. If it doesn't work, I can keep it outside the go-ipld-prime package potentially, and maybe throw it in go-graphsync or filecoin itself.

Also: I'm using "typesafe" euphemistically. I realize we're really just generating "valid" selectors, not type checking them.


warpfork:
I have "Hmm" feelings towards this.

Would it work to put it in a sub-package?  It would be nice to make it clear that these builders aren't part of the selector core spec per se.

If that trips on exporting of fields, can we make it work via serial round-trip?  While in some cases that might strike me as a code smell, here it actually seems perfectly reasonable because we want those paths to be exercised anyway; we *want* the serial form to be the natural home of the logic.

warpfork:

PR #20, opened 2019-07-18
More complete selector traversal

hannahhoward:
# Goals

Improve selector traversal you can cross link boundaries and more effectively traverse over lists and limited cardinality selectors

# Implementation

- Fix a bug in decoding of simple nodes to/from json/cbor -- this came up when I tried to use the link loader with some of the fixture data that wasn't otherwise used in the focus test
- Add link loading to traverseInformatively -- mostly just copied from Focus
- Modify iterator in traverseInformatively -- uses a new "SegmentIterator" concept that basically traverses by PathSegment, abstracting out the underlying ListIterator/MapIterator based on node type -- see selector.go
- factor out a traverseChild method for DRI and cause otherwise the traverseInformatively method would be even longer than it already is (though, I wonder if this is the best place to factor out a function cause it's inside the loop)

# For discussion

The basic goal here was to get to basically feature complete on selectors... even if there's probably some tech debt in here. I will go back and fix previous PRs and rebase on top of them shortly.

### DIFF
--- a/encoding/dagcbor/roundtrip_test.go
+++ b/encoding/dagcbor/roundtrip_test.go
@@ -43,3 +43,19 @@ func TestRoundtrip(t *testing.T) {
 		Wish(t, n2, ShouldEqual, n)
 	})
 }
+
+func TestRoundtripSimple(t *testing.T) {
+	simple := fnb.CreateString("applesauce")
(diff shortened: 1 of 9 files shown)


warpfork:

PR #21, opened 2019-07-22
Schema and codegen fun

warpfork:
Lots of stuff.  Lots more to go; this is just a checkpoint.

- Several new pieces of codegen, working towards generating structs.  Most of the accessors for Node are there; NodeBuilders still todo.
- Figured out a reasonable looking way to reduce the number of very similar methods in our codegen -- in the codegen source that is, not in the output.  The output contains a lot of redundancy primarily for the reason of unique contents in error messages.  (Weighed alternative approaches to this which would produce less redundant output, but am deciding that good error messages are more important than saving some bytes.)
- There are now exported builder methods for `schema.Type` values.  Note the cautionary comments about their temporariness (and seealso the full commit message starting with "Cut gordian knot").
- `schema.Kind` introduced, and relationship to `ipld.ReprKind` now more established.
- coreward: several new well-typed errors introduced.  (More refactors required to use those everywhere.  Want to do that asap in several more small branches off master after landing this one.)
- coreward: `ipld.Node` now has a `IsDefined() bool` property (as has been alluded to in commented-out docs for quite a while).
- added `ipld.Null` and `ipld.Undef` consts for convenience usage.


warpfork:
Belated "well duh"-realization while re-reading the message of 6898ed2 -- we can very likely not worry about the idea of typed null for the application of that Node being able to return a relevant NodeBuilder, because there's really no usecase which passes control flow directly through there:

- if building typed data with application logic, that logic certainly knows what type it's about to create to replace that field, and it just *does so*;
- if in the middle of a big unmarshal or other generic tree traversing, the logic should be using the NodeBuilder of parent objects to suggest the NodeBuilder for children and fields *anyway*, and that handles this as well.  (This is already the case because it's the only way to make systems like reflective `bind.Node` work even when they don't have schematic type information; and also looking like it's going to be true even for schematically typed node building because we'll want to carry along addition information about parser order strictness, which means keeping recursion purely on the builder side.)

So.  Probably we can do the simplest possible things with untyped nulls.  (Whew.)

PR #23, opened 2019-07-27
fix(encoding): match go-ipld-cbor cid encoding

hannahhoward:
go-ipld-cbor encodes a 0 byte ahead of cid bytes for multibase support - to maintain compatibility
of nodes encoded to CBOR, add this to encoder of go-ipld-prime.


warpfork:
~I wonder if this is suitably documented (ideally with rationale)~ No, I'm wrong, it's right here: https://github.com/ipld/specs/blob/8d7dbe757a1b42c5bb3704d29cf2fffdf8639d4d/block-layer/codecs/DAG-CBOR.md#link-format

Alright.  Yep.

PR #25, opened 2019-07-30
Fix interaction between ExploreRecursive and ExploreUnion

hannahhoward:
# Goals

Fix a bug where ExploreRecursive would not properly replace ExploreRecursiveEdge (and cause a
panic when traversed) if ExploreUnion was part of the recursion tree.

Essentially, the problem is an ExploreRecursiveEdge could end up a member of an ExploreUnion after an ExploreUnion is explored (remember, ExploreUnion returns another ExploreUnion if more than one member's Explore is not nil), which would not get detected by the existing checks in ExploreRecursive's Explore, which checked only if the next selector is an ExploreRecursiveEdge.
 
# Implementation

This is a possible fix till I can think of something better. Suffice to say... I'm not super happy about it. In short, it adds an additional check if the returned type is ExploreUnion, and if so, recursively replaces any ExploreRecursiveEdge members that are part of that explore union or of any children of the ExploreUnion that are also ExploreUnions (side node -- probably something should be done here about collapsing ExploreUnions under ExploreUnions -- cause I think that is a collapsable thing)

# For Discussion

Ack... now we've started special cases. What if we add another selector that is weird? I dunno quite what the general case solve is here.

Also, there are a bunch of selectors in the spec that aren't directly called out as invalid, but are weird and would break our current code. For example:

```json
{
  "exploreRecursive": {
    "maxDepth": 100,
    "sequence": {
      "exploreRecursiveEdge": {
      }
    }
  }
}
```

The above seems like a terrible idea but technically the spec doesn't say it's not ok. It would cause a panic on the interests method. Wonder if that should get updated, and marked as invalid in the parser.

### DIFF
--- a/traversal/selector/exploreRecursive.go
+++ b/traversal/selector/exploreRecursive.go
@@ -45,17 +45,58 @@ func (s ExploreRecursive) Interests() []PathSegment {
 // Explore returns the node's selector for all fields
 func (s ExploreRecursive) Explore(n ipld.Node, p PathSegment) Selector {
 	nextSelector := s.current.Explore(n, p)
+	maxDepth := s.maxDepth
 	if nextSelector == nil {
 		return nil
 	}
-	_, ok := nextSelector.(ExploreRecursiveEdge)
-	if !ok {
-		return ExploreRecursive{s.sequence, nextSelector, s.maxDepth}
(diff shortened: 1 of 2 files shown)


warpfork:

PR #26, opened 2019-08-06
Feat/recursive selectors

hannahhoward:
# Goals

An initial implementation of recursive selectors that offers a way to traverse them. Not covered is
validating ExploreRecursiveEdge being under an ExploreRecursive selector

# Implementation

Basically, this works in the sense that it supports recursion, including nested recursion and multiple ExploreRecursiveEdge's per ExploreRecursive

It does include one type assertion in ExploreRecursive's Explore method that enables everything to work. I wanted to check if this feels ok, and if this approach in general feels ok.

I haven't quite worked out how to validate that ExploreRecursiveEdge only appears under ExploreRecursive during selector parsing, or that every ExploreRecursive has at least one ExploreRecursiveEdge, and I wanted to verify if those validations are important to support during parse

reopens #18

### DIFF
--- a/traversal/selector/exploreAll.go
+++ b/traversal/selector/exploreAll.go
@@ -28,7 +28,7 @@ func (s ExploreAll) Decide(n ipld.Node) bool {
 }
 
 // ParseExploreAll assembles a Selector from a ExploreAll selector node
-func ParseExploreAll(n ipld.Node) (Selector, error) {
+func ParseExploreAll(n ipld.Node, selectorContexts ...SelectorContext) (Selector, error) {
 	if n.ReprKind() != ipld.ReprKind_Map {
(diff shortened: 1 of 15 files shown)


rvagg:
`ParseContext{}.Parse*()` feels pretty weird, is that just simplified for the tests and will likely be part of more complex construction in realistic scenarios or is this intended usage? I get the need for state with the parent stack but pushing that state management onto the API user feels a bit odd.

warpfork:
Might be improved by a constructor, yeah.  It's not always obvious that the zero value of something is a good place to start a call chain.

A top level `Parse()` that creates the first ParseContext and then does all the rest without ever exposing the ParseContext seems like it would fit nicely here.

We can stack that on top though.  I'm getting a little enthusiastic to start merging!

PR #34, opened 2019-10-22
feat(selectors): support union limit type

hannahhoward:
# Goals

Support https://github.com/ipld/specs/pull/207, now merged

# Implementation

In explore recursive, internally use a depth parameter and a hasDepth boolean. For limit type depth, depth = depth limit & hasDepth = true. For limit type none, depth = n/a & hasDepth = false

### DIFF
--- a/traversal/selector/exploreRecursive.go
+++ b/traversal/selector/exploreRecursive.go
@@ -34,7 +34,8 @@ import (
 type ExploreRecursive struct {
 	sequence Selector // selector for element we're interested in
 	current  Selector // selector to apply to the current node
-	maxDepth int
+	depth    int
+	hasDepth bool
 }
 
 // Interests for ExploreRecursive is empty (meaning traverse everything)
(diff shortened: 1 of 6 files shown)


traversal/selector/exploreRecursive.go
```diff
@@ -34,7 +34,8 @@ import (
 type ExploreRecursive struct {
 	sequence Selector // selector for element we're interested in
 	current  Selector // selector to apply to the current node
-	maxDepth int
+	depth    int
+	hasDepth bool
```
warpfork: This does all look about right, but I might consider extracting one more type here in the go code to stay topologically closer to the schema, and embed it here... like:

```
type RecursionLimit_Mode uint8

const (
    RecursionLimit_None RecursionLimit_Mode = 0
    RecursionLimit_Depth RecursionLimit_Mode = 1
)

type RecursionLimit struct {
    mode RecursionLimit_Mode
    depth int
}
```

... and then embed that, and most of the rest of the code stays the same, except a bunch of `if` blocks change to `switch` (which seems more correct if we do ever imagine adding more members to the limit union).

(Admittedly, it's always kind of frustrating to decide which of the 14(?) different kinds of sort-of-acts-like-a-union non-native construct in Go to use to implement a conceptual union.  I picked one at semirandom here.)

PR #35, opened 2019-10-24
FYC: add int, bytes, and link types to code-gen

hannahhoward:
# Goals

Totally selfish, I want to try to code gen protobuf nodes. I needed Link, Bytes, and Int to make it work.

# For Discussion

You do not need merge this. At the same time, it seemed like a lot of code busy work that might save you time. I get the sense you are still iterating on the core abstractions and focusing on the hardest part, but if this saves you time once you get there, here you go.


warpfork:
"still iterating on the core abstractions" is right for sure.  But all this looks great, and now that it's written, I think I'm happy to fold it in and keep it in sync on master!

PR #37, opened 2019-10-25
feat(schema/gen): make emit methods public

hannahhoward:
# Goals

I'm using some early code gen in https://github.com/ipld/go-ipld-prime-proto (DagPB for go-ipld-prime!).

Making these methods public seemed like the least intrusive way to allow a preliminary method for generating code outside this module.  I'm not asking for stability on these APIs -- just a commit in master, that go-ipld-prime-proto can be built on till things stabilize (and then it would be updated obviously)


warpfork:
Sold.

PR #38, opened 2019-11-07
fix(schema/gen): return value not reference

hannahhoward:
Map/List should return an actual node type, not a pointer to a node. Otherwise, breaks when nested
in other map/list type.

We use a pointer to a node when we're building nodes, for maps (so we can mod it easily I assume). However, when we actually build it, we should dereference it -- because that's the type anyone using the node expects-- and in fact things will error if you build a map inside a map.


warpfork:
Good eye.  I'm mildly surprised this even compiled...

PR #39, opened 2019-11-08
Feat/add code gen disclaimer

hannahhoward:
One more add -- a disclaimer about files being code gen'd which will cause most linters to ignore the file, avoiding lots of lint warnings :)


hannahhoward:
also includes last PR

warpfork:
dangit my email filters are _still_ not good enough, I missed this until now

TIL also!   Thanks!

PR #41, opened 2020-01-30
Add a typed link node to allow traversal with code gen'd builders across links

hannahhoward:
# Goals

Mostly, I want to support fast Graphsync :)
But concretely, when performing a traversal, if my schema contains a hint about what the node on the other side of the link is, and I have a custom builder for that type, I should utilize that information to determine what node builder to use in NodeBuilderChooser.

This also unlocks a lot of customization of NodeBuilderChooser without having to modify the default function.

There will still need to be some graphsync work to handle starting with the right root node builder, but this feels like it would unlock a lot of behaviors for code-gen

# For Discussion

I explored putting this method on the generated link itself. This turned out to have several problems -- specifically, links vary according to how they reference their data and how they load that data. Currently, the only link type is CIDs. It turned out to not be able to continue supporting an abstract link type and a concrete CID link type while also varying a link type according to where it is typed.


schema/gen/go/genKindLinkNode.go
```diff
@@ -98,6 +98,16 @@ func (gk generateKindLink) EmitTypedNodeMethodRepresentation(w io.Writer) {
 	`, w, gk)
 }
 
+func (gk generateKindLink) EmitTypedLinkNodeMethodReferencedLinkBuilder(w io.Writer) {
```
warpfork: s/ReferencedLinkBuilder/ReferencedNodeBuilder/ ?

hannahhoward: awesome, done, and then I'm gonna merge

PR #47, opened 2020-02-23
Path clarifications

warpfork:
Several significant clarifications to Path docs.

Primarily, being committal and direct about how special values, erm... *aren't* -- namely, "/" is a valid segment, and so is empty string -- and confessing how much tricky work that makes.

Several methods are now more upfront about how much they *do not do good things* on such values.  I'd like to fix these methods, but to do that, we should formally specify an escaping system and canonical string encoding for paths.  That work should start in the IPLD Specs repo, and involve fixtures.  Once that is tackled, I will (very!) gladly start fixing these methods to operate accordingly, and get rid of all the sharp-edges warnings.

Several methods also now include comments about whether they will enforce string character encoding or normalization on their function boundaries.  They won't.  (This is certainly the norm in Go -- the stance of "string is really just bytes" is ubiquitous in the gopher community -- so it may seem not worthy of wasting docs on to other gopher readers.  However, since IPLD is all about cross-language consistency, and we may have people from other programming language ecosystems reading our docs, it seems good to be as clear and explicit as possible.)  I suspect this *will* remain the case (it's turned out tricky and work-makey to try to specify anything else; subtracting the countably-infinite domain of non-normalized strings from the countably-infinite domain of all possible byte sequences makes for some very nasty and pernicious expressibility and round-tripping issues, long story short).

Thanks to @ribasushi for the kick in the shins that these docs needed work and clarifications.  The previous docs were written a fair while ago when we knew less things less confidently about how the system and specs would evolve, so these clarifications were badly needed.

### DIFF
--- a/path.go
+++ b/path.go
@@ -4,37 +4,99 @@ import (
 	"strings"
 )
 
-// Path is used in describing progress in a traversal;
-// and can also be used as an instruction for a specific traverse.
+// Path describes a series of steps across a tree or DAG of Node,
+// where each segment in the path is a map key or list index
+// (literaly, Path is a slice of PathSegment values).
+// Path is used in describing progress in a traversal; and
+// can also be used as an instruction for traversing from one Node to another.
+// Path values will also often be encountered as part of error messages.
 //
+// (Note that Paths are useful as an instruction for traversing from
+// *one* Node to *one* other Node; to do a walk from one Node and visit
+// *several* Nodes based on some sort of pattern, look to IPLD Selectors,
+// and the 'traversal/selector' package in this project.)
+//
+// Path values are always relative.
+// Observe how 'traversal.Focus' requires both a Node and a Path argument --
+// where to start, and where to go, respectively.
+// Similarly, error values which include a Path will be speaking in reference
+// to the "starting Node" in whatever context they arose from.
+//
+// The canonical form of a Path is as a list of PathSegment.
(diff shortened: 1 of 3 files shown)


path.go
```diff
@@ -4,37 +4,103 @@ import (
...
-// Paths are representable as strings.  When represented as a string, each
-// segment is separated by a "/" character.
-// (It follows that path segments may not themselves contain a "/" character.)
-// (Note: escaping may be specified and supported in the future; currently, it is not.)
+// A string of "/" is a valid PathSegment.
+// (As with empty strings, this is unfortunate (in particular, because it
+// very much doesn't match up well with expectations popularized by UNIX-like
+// filesystems); but, as with empty strings, maps which contain such a key
+// certainly exist, and it is important that we be able to regard them!)
 //
```
ribasushi: Perhaps something extra like ( just a suggestion )

```suggestion
//
// Lastly a string starting, ending, or otherwise containing the NUL (\x00) byte
// is also a valid PathSegment. This is of marked concern for languages implementing
// "C-style nul-terminated strings". When working with IPLD in such environments
// an implementation is required to accommodate NUL-containing paths
```

warpfork: Thanks, turned that into another paragraph.

(Kinda love the github "suggest" feature, too, thanks for using that.  Really gets things rolling.)

PR #52, opened 2020-04-23
Codegen update -- Assemblers, and many new representations

warpfork:
Warning: large diff.

This is a big update and rework of codegen.

Codegen was briefly left behind entirely by the winter update to the core interfaces which moved us to "assembler"-oriented operation -- wherein we mostly assign data into already-bulk-allocated memory -- over the previous "builders".  As a result, this is very nearly a total reboot.  The use of textual templating as the primary mechanism remains the same; and the overall pattern of {node, assembler, reprNode, reprAssembler} remains the same; but most of the actual code has churned.

Some of the highlight:

- There's now working proof of a type with various representation strategies!  Check out structs: they have a representation as a map, and also a representation of "stringjoin", which behaves *very* differently.
- There's now an "AdjunctConfig" system, and this supports customization of anything that's goland-and-this-code-generator-specific (as opposed to anything schema-specified).
  - Among the things that can be done with this is overriding the text used in creating symbols in the generated code.  This isn't quite wired up all the way to user-configurability, but the infra should all be in place to do so.
  - You can configure whether "MaybeT" types -- generated and used whenever something is either `nullable` or `optional` -- use pointers internally, or not.  (This might be something an advanced user legitimately wants to tune for performance reasons.)
- Everything is (of course) now generating NodeAssembler interfaces.
  - These get nested and embedded in each other.  So, whole deep trees of data can be created and even the assemblers get amortized into a single allocation.
- There's test infrastructure that's actually satisfying!
  - You can run 'go test' on the gen package and it will...
    1. exercise and test the generation itself
    2. ensure that the generated code actually compiles
    3. link the compiled objects back into the test process and run functional tests over them!
  - There's some caveats to this: unfortunately, some of the systems I used here don't work on windows, and they also require (blegh!) some cgo, and thus C compilers on your $PATH.  I'm not thrilled with this, but it has much better outcomes than the alternatives -- you can see the commit messages in the development process for more detail, including fully worked attempts at alternatives.
- ... much more.

This still has a decent number of TODOs, FIXMEs, and REVIEWPLZ comments in it, but may nonetheless already be an interesting read.

There's also a *lot* of code which is not yet even remotely "DRY".  I've been intentionally not DRY'ing things out on the first pass: it's been reliably non-obvious so far what parts are actually going to turn out reusable and at what granularity and boundaries, so it seems better (and less frankly brain-melting) to do things with high redundancy now, and factor it back out later once it's clear what works and what doesn't.

Despite the roughness, I might land this back on master soon.  It's still not recommended for incautious general use... but it's getting done enough to kick it around if you're feeling experimental, and it's certainly closer to useful than what's on master presently, so it's probably more useful to move forward than to twiddle thumbs.


schema/gen/go/genMinima.go
```diff
@@ -3,34 +3,40 @@ package gengo
...
+		package `+packageName+`
+
+		import (
+			"github.com/ipld/go-ipld-prime/schema"
+		)
+
+		// Code generated go-ipld-prime DO NOT EDIT.
+
+		const (
+			// The 'Maybe' enum does double-duty in this package as a state machine for assembler completion.
```
warpfork, line 23: If trying to review and understand how the whole child-completion signalling system works, this couple of comments is kinda key.

TODO(self): this comment block probably shouldn't be emitted into the *output*, though...

schema/gen/go/genStructReprStringjoin.go
```diff
@@ -0,0 +1,246 @@
...
+			}
+			{{- end}}
+			return nil
+		}
+	`, w, g.AdjCfg, g)
+}
+func (g structReprStringjoinReprBuilderGenerator) EmitNodeAssemblerType(w io.Writer) {
+	// - 'w' is the "**w**ip" pointer.
+	// - 'z' is used to denote a null (in case we're used in a context that's acceptable).  z for **z**ilch.
+	// - 'fcb' is the **f**inish **c**all**b**ack, supplied by the parent if we're a child assembler.
```
warpfork, line 191: FIXME(self): this is both out of date and clearly not covered by tests :/

Current tests for structs with stringjoin representation cover type-level assemblers but apparently not representation-level assemblers

schema/gen/go/genMap.go
```diff
@@ -0,0 +1,631 @@
...
+	// Generate a speciated Lookup as well as LookupMaybe method.
+	// The LookupMaybe method is needed if the map value is nullable and you're going to distinguish nulls
+	//  (and may also be convenient if you would rather handle Maybe_Absent than an error for not-found).
+	// The Lookup method works fine if the map value isn't nullable
+	//  (and should be preferred in that case, because boxing something into a maybe when it wasn't already stored that way costs an alloc(!),
+	//   and may additionally incur a memcpy if the maybe for the value type doesn't use pointers internally).
+	// REVIEW: is there a way we can make this less twisty?  it is VERY unfortunate if the user has to know what sort of map it is to know which method to prefer.
+	//  Maybe the Lookup method on maps that have nullable values should just always have a MaybeT return type?
+	//   But then this means the Lookup method doesn't "need" an error as part of its return signiture, which just shuffles differences around.
+	doTemplate(`
```
warpfork, line 54: Dear Reviewer, may I direct your attention to this area especially?  There's quite a lot of ergonomics questions here.  I really haven't sorted out what I even *want* this to look like, and there seems to be a tug-o-war between pleasantness/consistency versus making it obvious what's performant and natural to prefer the usage that is fastest.

schema/gen/go/genMap.go
```diff
@@ -0,0 +1,631 @@
...
+
+func (g mapGenerator) EmitNodeTypeAssertions(w io.Writer) {
+	doTemplate(`
+		var _ ipld.Node = ({{ .Type | TypeSymbol }})(&_{{ .Type | TypeSymbol }}{})
+		var _ schema.TypedNode = ({{ .Type | TypeSymbol }})(&_{{ .Type | TypeSymbol }}{})
+	`, w, g.AdjCfg, g)
+}
+
+func (g mapGenerator) EmitNodeMethodLookupString(w io.Writer) {
+	// What should be coercible in which directions (and how surprising that is) is an interesting question.
```
warpfork, line 128: Dear Reviewer, may I request your attention to this bit especially?

There's some _very_ interesting implications for PathSegments, how map keys work, and systemically emergent limitations that we need to think about.

schema/gen/go/genMap.go
```diff
@@ -0,0 +1,631 @@
...
+			{{- end}}
+		}
+	`, w, g.AdjCfg, g)
+}
+
+func (g mapGenerator) EmitNodeMethodLookup(w io.Writer) {
+	// LookupNode will procede by cast if it can; or simply error if that doesn't work.
+	//  There's no attempt to turn the node (or its repr) into a string and then reify that into a key;
+	//   if you used a Node here, you should've meant it.
+	// REVIEW: by comparison structs will coerce anything stringish silently...!  so we should figure out if that inconsistency is acceptable, and at least document it if so.
```
warpfork, line 165: This is the second half of the questions in the area of https://github.com/ipld/go-ipld-prime/pull/52/files#r413877995 .

node/realgen/tString.go
```diff
@@ -0,0 +1,229 @@
...
+func (String) AsBool() (bool, error) {
+	return mixins.String{"realgen.String"}.AsBool()
+}
+func (String) AsInt() (int, error) {
+	return mixins.String{"realgen.String"}.AsInt()
+}
+func (String) AsFloat() (float64, error) {
+	return mixins.String{"realgen.String"}.AsFloat()
+}
+func (n String) AsString() (string, error) {
```
willscott, line 102: thoughts on also adding the [`String()`](https://golang.org/src/fmt/print.go#L62) method so values are auto-converted by fmt?

warpfork: Uncertain!  What do you think?

There's also the possibility of types having a `String() string` method (no error return, like `AsString` has) when their kind is in fact `string`.  But I wonder if that should be reserved as an interface-castable indictation of string kinds?

Perhaps considering that an indicative interface is pretty foolish considering `fmt.Stringer` already has the contract it does, though.

warpfork: There's also the question of whether `String() string` should exist on things such a struct type with stringjoin representation.  On the representation node, the answer is definitely 'yes'.  On the type-level node, I guess it would be {insert answer to whether we just always emit a fmt.Stringer}.

warpfork:
> It looks like the current approach will end up feeling like there's quite a bit of 'bloat'? - i guess the main eye here should be towards ergonomic godoc and autocomplete, with a secondary hope of minimizing exported methods / compile size.

Yep, this is the case.  In general I'm going for the priority order of Correct > Fast > Readable/Autocomplete-friendly/Doc-friend > Small-Compile-obj > Small-GSLOC.

I don't really know how to do a monomorphizing interface without quite a lot of "nope" method stubs, unfortunately.  Some of them could be grouped up behind one sub-interface (e.g. all the `Lookup*` methods could go on a `MapNode` interface), but this doesn't help overall all that much -- the part I can't think of a way to get around is that there'd still be at least one method per kind -- and would make most user code call trees into a longer and less pleasant chain.

node/realgen/tString.go
willscott: I think, as long as there's a story, be it implementing `Formatter`, or one of the other interfaces that allow the various common debugging methods to display the value in a sane way I don't have strong attachments to `Stringer` in particular.

I see your point about helping programmers do good defensive programming by forcing them to consider the error, and at the same time see why that's often not ergonomic in Go. I would probably provide common implementer of interfaces to allow that final value to be directly passed into recipient functions with one less extraction of error.

warpfork:
FWIW, I think the binary object size for most of these stubs should be on the order of a few dozen bytes at most.

I don't have the infrastructure to measure and test this conveniently set up yet, but my vague intuition based on previous drags from the assembly faucet is that all the errors should compile down into a word of reference into the global string table for the type name text, and then a few opcodes worth of setting it in a struct, and that's about the end of it.  (The function calls to the types in the `mixins` package that do the rest of the error value construction *do* get inlined and smashed down; that at least I did check at some point.)

When we do get around to scripting measurements of this, there will probably be room for improvement, but I'm expecting it to be by some constant factor.  It might be a locally sizable multiplier, but it's unclear how big it will be in a bigger picture.  Probably "nice to have" territory; probably also dwarfed by the size change one can get by stripping DWARF symbols.

But we'll see; these are just my preliminary bets.  More measurements wanted!

warpfork:
I got nerdsniped for a moment thinking about what it would be like to do more granular sub-interfaces of Node to try to reduce the number of stub methods, so, some quick notes on where that train of thought took me:

First of all, I definitely considered this a long time ago, but discarded it.  I don't have good notes from the time.  So maybe it's not a bad idea to re-open it for consideration.  (And maybe something learned since then could change the consideration, too.)

I'm also a little worried that the more-interfaces approach would turn out worse in performance, though.  Multiple interface satisfaction checks to figure out what Kind something is def slower than switching on the Kind enum... So we'd still have that switch in any generic code... And then we'd have unconditional casts to the interface in question, which may or may not be optimizable (need to check -- probably partially, but also, the compiler can't "know" that the enum-based switch is supposed to indicate that the interface cast should (and boy, there's a big asterisk on "should" about lawful implementors of the interface, too) be statically unable to fail, so it'll still have to generate error paths)?  Or if we don't make the user do the casts, then the cast wrapping method call, which hopefully doesn't obstruct inlining in whatever context it's used (but who can say, since it's inlining -- it probably bumps a counter that contributes to inline abort if nothing else)?

And it's also just not at all clear that more interfaces would make the API more fun to use.  `more.dotted.calls` per single node traversed tends not to feel better.

So, there's lots of reasons to be wary of that direction.  Would be interesting to try it!  But to really know, you'd have to build a whole library that way, take it to production, and see how it behaves end to end.

The r&d cost of actually executing on these kinds of alternative core interfaces is unfortunately astronomical.  Sometimes I gotta pick one approach and stick with it.  I hope I've picked well...

And if not?  Well, I'm perfectly comfortable with the idea of alternative competitive IPLD libraries being written.  There's more than one golang library for protocol buffers, too, after all; and that's kind of awesome, really.  Different strokes for different user stories and development styles.

warpfork:
Oh, also, other alternative that you might be thinking of just to reduce the GSLOC size -- can we embed something that saves us from generating all those stubs?

Well, no.

At least, not without making one of two other unacceptable choices: either not having type name info in the error (boo usability), or, having the embedded thing use actual memory to story what's functionally a constant for that text (boo performance: totally useless resident memory size increase).

If we wanted to reduce the volume of code nicely while still having the type name be available to the error construction, we'd want some sort of feature in Golang that's like "embedding" in that it brings methods, but has no state, but also can take some compile-time tag that adds some info to the methods.  But... ah, this is inventing macros :sunflower: 

If you view what codegen does as if it was a macro system, it feels a lot better.  It's just a macro system that puts the intermediate result in plain view.  By comparison, lots of other language's macro systems would let you whimsically generate far more bloated code than this... it would just be harder to notice it.

warpfork:
Okay -- I think it's about time to sync this with master and merge it.

It's still far from perfect, but 99 commits of work is far enough divergence.  Work will continue on new (hopefully mostly _smaller_) branches after this.

Issue #54, opened 2020-04-30
Rename NodeStyle to NodePrototype


warpfork:
Like it says on the tin.  Just rename the thing.

Both names for this comment had been in the running originally.  I chose to roll with NodeStyle previously because it's just plain character-count shorter.  But as time passed and more feedback is accumulated, NodePrototype appears to be the preferred.

Discussion with @hannahhoward preferred NodePrototype, with the only condition being that if it's confusable with prototypical inheritance concepts, that might be problematic.  Acausally, @rvagg had also expressed a preference for NodePrototype, and as he's pretty deeply steeped in the javascript world, I'm going to take that as a pretty solid point that the potential confusion with prototypical inheritance is *not* a huge problem in this case.

We're currently tagged at `v0.0.3` version; I'll probably make this change just before whatever becomes `v0.0.4`.  It should be a "sed-refactor" (trivial find-and-replace) when it lands, so hopefully not too disruptive to downstreams.

---

Additionally: it may be worth noting that in the realm of schemas: we have a `Type` interface, which provides access to more info about a type.  _These will probably be the **same** as the Prototype_ for nodes that have them.

Or at least, conditionally.

If you have some `foo.ipldsch` and you do codegen with it: it'll emit a bunch of code and all those nodes will have a `Type()` method, and the value it returns will be pointer-equal to what the  ~`Style()`~ `Prototype()` method returns.

If you have some `foo.ipldsch` and parse it and do runtime wrapper nodes to get the typed behaviors: they'll also have `Type()` and ~`Style()`~ `Prototype()` methods which return pointer-equal things.

And those values will also be *content*-wise equal to what you get from the codegen.  But not pointer-equal.

It's hoped that the fact that the "type" and "prototype" are the same thing is reasonably comprehensible.  Or at least, I hope it's more comprehensible than that the "type" and the "style" are the same thing.  Alternate vocab for this is still open to discussion if someone has ideas that seem better.

rvagg:
Dirty secret of the JavaScript world: a small portion of JS programmers fully understand the prototypical inheritance system, and that's why we got the `class` keyword eventually. Even though a `class` is just a shorthand for what you'd do with prototypes, it's way less confusing for most programmers so it now rules. They remain an obscure concept in CS even with JS's widespread adoption. You're safe with this one and I think it actually makes a ton of sense in this instance.

warpfork:
This happened already!  Closing.

PR #59, opened 2020-06-29
API updates for v0.5: the renamening

warpfork:
Several API naming changes that have been buffered and planned for a while are now getting applied.  :D

The CHANGELOG diff says it as well, but for recap in github's PR UI:

- Renamed: `NodeStyle` -> `NodePrototype`.
	- Reason: it seems to fit better!  See https://github.com/ipld/go-ipld-prime/issues/54 for a full discussion.
	- This should be a "sed refactor" -- the change is purely naming, not semantics, so it should be easy to update your code for.
	- This also affects some package-scoped vars named `Style`; they're accordingly also renamed to `Prototype`.
	- This also affects several methods such as `KeyStyle` and `ValueStyle`; they're accordingly also renamed to `KeyPrototype` and `ValuePrototype`.
- Renamed: `(Node).Lookup{Foo}` -> `(Node).LookupBy{Foo}`.
	- Reason: The former phrasing makes it sound like the "{Foo}" component of the name describes what it returns, but in fact what it describes is the type of the param (which is necessary, since Golang lacks function overloading parametric polymorphism).  Adding the preposition should make this less likely to mislead (even though it does make the method name moderately longer).
	- This should be a "sed refactor" -- the change is purely naming, not semantics, so it should be easy to update your code for.
- Renamed: `(Node).Lookup` -> `(Node).LookupNode`.
	- Reason: The shortest and least-qualified name, 'Lookup', should be reserved for the best-typed variant of the method, which is only present on codegenerated types (and not present on the Node interface at all, due to Golang's limited polymorphism).
	- This should be a "sed refactor" -- the change is purely naming, not semantics, so it should be easy to update your code for.  (The change itself in the library was fairly literally `s/Lookup(/LookupNode(/g`, and then `s/"Lookup"/"LookupNode"/g` to catch a few error message strings, so consumers shouldn't have it much harder.)
	- Note: combined with the above rename, this method overall becomes `(Node).LookupByNode`.
- Renamed: `ipld.Undef` -> `ipld.Absent`, and `(Node).IsUndefined` -> `(Node).IsAbsent`.
	- Reason: "absent" has emerged as a much, much better description of what this value means.  "Undefined" sounds nebulous and carries less meaning.  In long-form prose docs written recently, "absent" consistently fits the sentence flow much better.  Let's just adopt "absent" consistently and do away with "undefined".
	- This should be a "sed refactor" -- the change is purely naming, not semantics, so it should be easy to update your code for.

*All* of these changes are simple renames.  The compiler should highlight them instantly for anyone updating their use of this library; the fixes to consuming code are trivial to apply; and this should generally be easy to absorb.

Nonetheless, since they're technically "breaking" changes, you'll find that `v0.4` was tagged immediately before the start of this branch; and I'll be tagging a `v0.5` as soon as this branch merges to master.  This should make it easier to step through just these changes (and nothing else) all at once, at the consumer's leisure.


warpfork:
Merged; and v0.5 is now tagged on master :tada:

PR #60, opened 2020-07-08
Codegen of unions, and their keyed representations

warpfork:
Unions!  :tada: 

Unions were one of the last major groups of functionality in the feature table that we had *no* codegen implementations of yet, so this is pretty huge news overall.  Nothing broke; no abstractions shattered.  Whew.  (Enums are the only thing left, but those are pretty easy-peasy.  Unions, being a container/recursive type kind rather than a leaf, were pretty heavy-duty.)

We have two things here, as always: the "type-level" and the "representation-level" implementations of the Node and NodeAssembler interfaces.

- the "type-level" Node for a Union always acts like a map, and the keys are always the names of the member Type.
- the "representation-level" Node behavior may (of course) vary, but for "keyed" representations... it's roughly the same as the type-level semantics; it's just that the keys are the strings defined by the schema author, rather than being the member type's name.

The "keyed" representations are the only representation implemented so far, as those are by far the easiest to ship, but adding the rest should just be a Small Matter of Programming.  (Which is not to say easy or instantaneous to scratch out, but at least it's not _theoretically_ challenging, nor likely to encounter Unknown Unknowns.)

Individual commits have more details in their messages.

---

Two styles of internal memory layout in the generated code are supported!  One is "embedAll"; the other is "interface".

- In "embedAll" memory layout, it... does what it say on the tin: all possible member types are embedded into the union type.  Only one of them is actually initialized with data; which one that is is marked by a 'tag' field which adds another word of memory.  Overall this means the resident memory size can be potentially rather large, but, the allocation count when using these is zero.  (If we had language-level support for "unions" that share memory in Golang, the resident memory size could be much less, as well.  Alas; this is not available (unless one resorts to 'unsafe', which as a general rule I avoid).)
- In "interface" memory layout, the generated golang structure for the union type will contain just one field, which is an interface.  The type info is thus in the golang native interface type info; simple.  These interfaces are smaller in resident memory (and probably the right choice by default), but in contrast to "embedAll", constructing them does necessarily cause heap allocations.

(All this talk about performance differences of these two memory layout strategies may sound theoretical!  And perhaps for applications that don't care about performance, that is true.  However, we do also have examples of both modes elsewhere in the go-ipld-prime codebase already, and they were absolutely the result of profiling which indicated it made a serious, make-or-break difference when those types were involved in "hot" loops.  Check out the `ipld.PathSegment` type, for example, which resembles the "embed" strategy.)

In either case, an interface type is also generated, so that if the user of the generated code wants to do a native type-switch of their own, they can call `AsInterface()` and proceed to do so.

---

There are still substantial "todos" regarding the quality of error handling.  The types used sometimes carry too much information, or too little; or the information that they carry is inconvenient (e.g. may require processing) to actually have on hand in some cases; and the general consistency of the error type ontology... well, isn't.  I've started accumulating more comments and notes about this, but a massive pass over the entire suite of relevant packages will be required at some point, once we've gathered enough experiences to make final strategic decisions about the subject.

(One example thought on the subject of errors is: perhaps we should start attaching NodePrototype values to them, and also introduce a Stringer requirement to that interface.  This would serve up type name info for typed/codegen'd nodes, while also being clearly defineable for untyped nodes; this would help resolve a lot of awkward sticking points currently showing up around error info handling.  But: this is a biggish change; it would live in its own PR.)

---

Looking forward...

In the the near future: one of the reasons having unions be generable now is that a lot of schemas have at least one of them... including, for example, the schema-schema itself (the "self-hosting" document that describes the schema information as a schema).  This means we're getting really close to being able to generate code that we could then use to parse our own schema ASTs!  That, in turn, would let us throw out a great deal of placeholder code which was waiting for this day; and also means we'd get e.g. JSON-unmarshalling of Schema info "for free"... which in turn will mean we're dang near ready to make an end-user CLI tool for all this codegen stuff that you can feed declarative schema documents into and go!

(... but, one step at a time.  First, it turns out we'll need to fix the "placeholder" schema info types one more time: it's not currently possible to construct values with "cycles" in them, and we actually need that to describe the schema-schema.  Heh.)

(... and, the schema-schema currently uses inline unions, as well as keyed unions.  I might actually change the schema-schema.  It would make boostrapping to self-hosting easier, yes.  But there are other reasons as well: I've become much more averse to inline unions through the course of this work, and seeing more and more closely how their performance will (necessarily!) suck.  I also suspect the ease-of-implementation of keyed unions will be a recurring theme for people building IPLD + Schema tooling in other languages, and therefore, making the schema-schema use only keyed unions will be a bootstrap simplifying move for them, too.  All in all, such a change seems likely worth pursuing.)

Tl;dr much more excitement soon to come :3

### DIFF
--- a/nodeBuilder.go
+++ b/nodeBuilder.go
@@ -88,6 +88,11 @@ type MapAssembler interface {
 	// (Design note: a string is sufficient for the parameter here rather than
 	// a full Node, because the only cases where the value types vary are also
 	// cases where the keys may not be complex.)
+	//
+	// FIXME: define a "nopePrototype" -- structs and unions need some way to say that the given 'k' isn't acceptable at all.
+	//  I'm not sure when that thing gets to brass tacks about erroring either though!  The NewBuilder method can't either.
+	//  Maybe we should choose between this panicking (right away, no "nopePrototype"), or adding an error return to it.
+	//   (We don't call this all that much since the NodeAssembler restructuring; maybe an annoying error return is fine.)
 	ValuePrototype(k string) NodePrototype
 }
 
--- a/nodeBuilder.go
+++ b/nodeBuilder.go
@@ -84,15 +84,12 @@ type MapAssembler interface {
 	// For plain maps (that is, not structs or unions masquerading as maps),
 	// the empty string can be used as a parameter, and the returned NodePrototype
 	// can be assumed applicable for all values.
-	// Using an empty string for a struct or union will return a nil NodePrototype.
+	// Using an empty string for a struct or union will return nil,
+	// as will using any string which isn't a field or member of those types.
+	//
 	// (Design note: a string is sufficient for the parameter here rather than
 	// a full Node, because the only cases where the value types vary are also
 	// cases where the keys may not be complex.)
-	//
-	// FIXME: define a "nopePrototype" -- structs and unions need some way to say that the given 'k' isn't acceptable at all.
-	//  I'm not sure when that thing gets to brass tacks about erroring either though!  The NewBuilder method can't either.
-	//  Maybe we should choose between this panicking (right away, no "nopePrototype"), or adding an error return to it.
-	//   (We don't call this all that much since the NodeAssembler restructuring; maybe an annoying error return is fine.)
 	ValuePrototype(k string) NodePrototype
 }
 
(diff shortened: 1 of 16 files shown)


warpfork:
Charging forward: I want to start new branches to work on the cycle issues, and thence onto that self-hosting milestone.  (I know there's going to be more fun speedbumps on the way to the latter, so I want to encounter them as soon as possible, on the figuring they'll also force a lot of other stuff to get straightened out...)

PR #63, opened 2020-07-30
Struct tuple representation codegen

warpfork:
Codegen now supports structs with a tuple representation strategy -- that is, a representation that's a list, where the position alone corresponds to each field in the struct, and thus to the expected type and the meaning of the value.

### DIFF
--- a/schema/gen/go/genStructReprTuple.go
+++ b/schema/gen/go/genStructReprTuple.go
@@ -1,0 +1,443 @@
+package gengo
+
+import (
+	"io"
+	"strconv"
+
+	"github.com/ipld/go-ipld-prime/schema"
+	"github.com/ipld/go-ipld-prime/schema/gen/go/mixins"
+)
+
+var _ TypeGenerator = &structReprTupleGenerator{}
+
+// Optional fields for tuple representation are only allowed at the end, and contiguously.
+// Present fields are matched greedily: if the struct has five fields,
+//  and the last two are optional, and there's four values, then they will be mapped onto the first four fields, period.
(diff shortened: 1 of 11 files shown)


schema/gen/go/genStructReprTuple.go
```diff
@@ -0,0 +1,443 @@
...
+		Type                  *schema.TypeStruct
+		HaveTrailingOptionals bool
+	}{
+		g.Type,
+		haveTrailingOptionals,
+	})
+}
+
+func (g structReprTupleReprGenerator) EmitNodeMethodLength(w io.Writer) {
+	// This is fun: it has to count down for any unset optional fields.
```
rvagg, line 224: don't we have special rules for optional fields in tuple representation? like, only the last one can be optional for it to make much sense?

warpfork: Yep.  There's some comments about that at the top of the file.  (There can be several optionals... as long as they're all consecutive, and the matching is greedy.)

I think I was lazy in this snippet: some of the code is very similar to code used in optional-value-presense-counting in structs representations as maps, and I just didn't bother to rewrite it to use the assumptions we're allowed here.

(Optimistically, I was thinking we might be able to make them share code entirely, but I experienced "mental OOM" trying to get there, so this is the non-DRY'd shrapnel.  There will be a *lot* of stuff like this that deserves revisit in the near future: things will either turn out to be very "obvious" candidates for DRYing, or, very much deserving of further specialization -- but I'm frequently giving up trying to tell the difference as I chase feature-completeness right now.  Mostly, I'm trying to at least mark them in comments, but I'm sure I often fail.)

PR #65, opened 2020-08-20
Add traversal.Get function

warpfork:
Add `traversal.Get` (and `traversal.Progress.Get`), two methods which take a `Node` and a `Path`, do a little walk, and return the reached `Node`.

This is a (long-overdue) helper function.  It's the same functionality as has already been available through `traversal.Focus`, but returns values rather than taking a callback parameter, which is more convenient in some cases.

This new method also doesn't bother to accumulate Path and LastBlock info into an updated Progress object, because it doesn't return one.  If the user wants the Progress info, they should continue to use Focus and a callback -- using a callback encourages clear handling of the scope of the Progress object, which should be desirable anyway.

`traversal.Get` doesn't quite work for call-chaining on its own, because it potentially needs to return an error, but composes nicely with other error-redirection systems like the `must` package: `must.Node(traversal.Get(startNode, ipld.ParsePath("foo/bar"))).Kind()` works, for example (substitute any call you wish to chain in place of `.Kind()`).

### DIFF
--- a/traversal/focus_test.go
+++ b/traversal/focus_test.go
@@ -123,6 +123,26 @@ func TestFocusSingleTree(t *testing.T) {
 	})
 }
 
+// covers Get used on one already-loaded Node; no link-loading exercised.
+// same fixtures as the test for Focus; just has fewer assertions, since Get does no progress tracking.
+func TestGetSingleTree(t *testing.T) {
+	t.Run("empty path on scalar node returns start node", func(t *testing.T) {
+		n, err := traversal.Get(basicnode.NewString("x"), ipld.Path{})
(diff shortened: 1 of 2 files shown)


traversal/focus_test.go
```diff
@@ -123,6 +123,26 @@ func TestFocusSingleTree(t *testing.T) {
 	})
 }
 
+// covers Get used on one already-loaded Node; no link-loading exercised.
```
rvagg, line 126: a test that exercises an attempt to traverse through a link might be good, it would at least show what would be expected to occur in such a situation, I imagine it's some error that bubbles up from the bowels of `Link#Load()`?

PR #66, opened 2020-08-24
New testcase system for exercising typed nodes; Revamp struct tests to use it.

warpfork:
Introduce some new test helpers, focusing on table-driven tests, and leaning heavily upon json as a shorthand for expressing fixtures.

It also makes a great deal more effort to exercise the different features of nodes (and their paired representation nodes) from all directions at once for each test datum, rather than requiring that all be written out manually, which is how I'd been slogging along previously.

The result is that the struct tests we've renovated have a lovelydiffstat shrinkage: 111 insertions, 299 deletions... but the smaller line count results in *more* coverage.  Yay!

(Okay, the linecount increase for the testcase structure and helper methods is much bigger than the savings in fixture size... but, only *so far*.  I assume this will continue to pay off in the future.)

Relatedly: a bug in struct map representations has been fixed. (It was the sibling of 5f58965, embarrassingly.)  Thank goodness we now
get proper coverage of this area.

There's a few TODOs left to further expand the some of the tests performed; not sure if I'll append those to this PR before merging or do them in separate small future PRs.  Same goes for further expansion of usage of this new system.

There's also some future work to do around the lookup error assertions -- the main sore spot there is that the traversal package doesn't emit very well-typed errors.  That's likely best solved in a separate PR since it will mostly involve that package.


schema/gen/go/testcase_test.go
```diff
@@ -0,0 +1,242 @@
...
+			- But you should be able to do it, with some care.
+		- This whole system depends on json parsers and serializers already working.
+			- This is arguably an uncomfortably large and complex dependency for a test system.  However, the json systems are tested by using basicnode; there's no cycle here.
+		- "Unhappy paths" in creation are a bit tricky to test.
+			- It can be done, but for map-like things, only for the AssembleEntry path.
+			- PRs welcome if someone's got a clever idea for a good way to exercise AssembleKey+AssembleValue.  (A variant of unmarshaller implementation?  Would do it; just verbose.)
+		- No support yet for checking properties like Length.
+			- Future: we could add another type-hinted special case to the testcasePoint.expect for this, i suppose.
+*/
+type testcase struct {
```
mvdan, line 65: nit: testCase, since it's two words :) also in the other type names

schema/gen/go/testcase_test.go
```diff
@@ -0,0 +1,242 @@
+package gengo
+
+import (
+	"bytes"
+	"fmt"
+	"testing"
+
+	"github.com/polydawn/refmt/json"
+	"github.com/polydawn/refmt/shared"
+	. "github.com/warpfork/go-wish"
```
mvdan, line 10: I wonder if you've seen https://pkg.go.dev/github.com/frankban/quicktest. Its design is pretty interesting, and chunks of it are being upstreamed in the form of `testing.T.Cleanup` and `testing.T.TempDir`.

I'm not a fan of the dot-import, though. I get that the APIs stutter otherwise, like `wish.Wish(...)`, but that seems like a design issue that should be fixed and not worked around with dot-imports.

schema/gen/go/testStructsContainingMaybe_test.go
```diff
@@ -335,14 +40,119 @@ func TestStructsContainingMaybe(t *testing.T) {
 		}),
 	))
 
+	// There's a lot of cases to cover so a shorthand labels each case:
+	//  - 'v' -- value in that entry
+	//  - 'n' -- null in that entry
+	//  - 'z' -- absent entry
+	specs := []testcase{
+		{
+			name:     "all fields set -- vvvvv",
```
mvdan, line 49: test names, and by extension sub-test names, should generally be camelcase. So you should do `SomeTest/SomeSubtest` instead of `some test/some subtest`, particularly because spaces don't play nice with the input/output and get replaced by other characters.

I think symbols are OK if needed, so you could do something like `AllFieldsSet-vvvvv`.

schema/gen/go/testcase_test.go
warpfork: I hadn't!  Haven't cachebusted my reviews of test libraries for a while.  That one looks nice.  Worth considering.

schema/gen/go/testStructsContainingMaybe_test.go
rvagg: Is this an assumption built into tooling, or just a convention that's good to adhere to just because it's a convention? (Not being disagreeable, simply interested).

schema/gen/go/testStructsContainingMaybe_test.go
```diff
@@ -335,14 +40,119 @@ func TestStructsContainingMaybe(t *testing.T) {
...
+				{"r1", "a"},
+				{"r2", "b"},
+				{"r3", ipld.Null},
+				//{"r4", ipld.ErrNotExists{}}, // TODO: need better error typing from traversal package.
+				//{"f5", ipld.ErrNotExists{}}, // TODO: need better error typing from traversal package.
+			},
+			typeItr: []entry{
+				{"f1", "a"},
+				{"f2", "b"},
+				{"f3", ipld.Null},
```
rvagg, line 141: This is weird, when compared to the previous case, but is it just supposed to be testing something about the guts of serialization (still a `TODO` I see), maybe this could do with a doc to explain why it's a special case—cause I can't really guess why it's this way.

schema/gen/go/testStructsContainingMaybe_test.go
mvdan: Yeah, it's a convention - though driven by the tooling as well, because the test names can't contain whitespace and thus it gets replaced in the output of `go test`. It's also encouraged by the upstream packages; you can search for "naming convention" under https://golang.org/pkg/testing/#hdr-Examples.

schema/gen/go/testcase_test.go
warpfork: camelCase as a rule, of course... but "testcase" is a single word in my mind (if admittedly a neologism).  :thinking:  I find it both ways in the golang codebase itself, although the two-word interpretation is more frequent.  Do you feel strongly about this?

schema/gen/go/testStructsContainingMaybe_test.go
warpfork: IIUC, any funky characters get replaced by underscores.  I typically don't really mind this; the result still reads clearly enough in the terminal.

In this case the resulting "`_--_`" is particularly nasty, though, so yeah, I'll rejigger these.

schema/gen/go/testStructsContainingMaybe_test.go
warpfork: It's [covered in the docs on the testcase struct](https://github.com/ipld/go-ipld-prime/pull/66/files#diff-0e13278986646211b49e3940c88de7f4R45-R47): not all values are marshallable when examined at the type level (even though their `.Representation()` node of course is (and must be) marshallable).  Maps with complex keys, and structs with absent values, are both examples of this.  So for these we still want to test iterators, but can't use marshalling as a convenient way to do so: so we do something explicit instead.

schema/gen/go/testcase_test.go
mvdan: it's a nit, so I don't feel strongly about it

schema/gen/go/testStructsContainingMaybe_test.go
mvdan: it's just easier for everyone if the names are consistent in the output and source code, and it's also more consistent with how most people tend to write Go tests :)

schema/gen/go/testStructsContainingMaybe_test.go
warpfork: Followup: added the code to properly exercise this bit.  And yes, you did spy a feature that was wrong here.  Fixed :)

warpfork:
There's still an open TODO in here about adding a Copy feature and using it to make sure both the AssembleKey+AssembleValue path is exercised on map-like stuff in addition to the 2-for-1 AssembleEntry path (which is what's exercised by unmarshalling, and thus currently exercised by this system).  But I think I'm going to be okay with that for today and merge this; I want to start writing more tests of other stuff with this system, and not pollute this PR's diff view with that stuff.

Rebased and mergin'.

PR #70, opened 2020-08-28
fix(cidlink): check for byte buffer

hannahhoward:
See slack message re: mem allocations + io.ReadAll

### DIFF
--- a/linking/cid/cidLink.go
+++ b/linking/cid/cidLink.go
@@ -30,20 +30,30 @@ func (lnk Link) Load(ctx context.Context, lnkCtx ipld.LinkContext, na ipld.NodeA
 	if !exists {
 		return fmt.Errorf("no decoder registered for multicodec %d", lnk.Prefix().Codec)
 	}
-	var hasher bytes.Buffer // multihash only exports bulk use, which is... really inefficient and should be fixed.
-	decodeErr := mcDecoder(na, io.TeeReader(r, &hasher))
-	// Error checking order here is tricky.
-	//  If decoding errored out, we should still run the reader to the end, to check the hash.
-	//  (We still don't implement this by running the hash to the end first, because that would increase the high-water memory requirement.)
-	//   ((Which we experience right now anyway because multihash's interface is silly, but we're acting as if that's fixed or will be soon.))
-	//  If the hash is rejected, we should return that error (and even if there was a decodeErr, it becomes irrelevant).
-	if decodeErr != nil {
-		_, err := io.Copy(&hasher, r)
-		if err != nil {
-			return err
+	var hasherBytes []byte
+	var decodeErr error
+	byteBuf, ok := r.(*bytes.Buffer)
+	if ok {
+		hasherBytes = byteBuf.Bytes()
+		decodeErr = mcDecoder(na, r)
+	} else {
+		var hasher bytes.Buffer // multihash only exports bulk use, which is... really inefficient and should be fixed.
+		decodeErr = mcDecoder(na, io.TeeReader(r, &hasher))
+		// Error checking order here is tricky.
+		//  If decoding errored out, we should still run the reader to the end, to check the hash.
+		//  (We still don't implement this by running the hash to the end first, because that would increase the high-water memory requirement.)
+		//   ((Which we experience right now anyway because multihash's interface is silly, but we're acting as if that's fixed or will be soon.))
+		//  If the hash is rejected, we should return that error (and even if there was a decodeErr, it becomes irrelevant).
+		if decodeErr != nil {
+			_, err := io.Copy(&hasher, r)
+			if err != nil {
+				return err
+			}
 		}
+		hasherBytes = hasher.Bytes()
 	}
-	cid, err := lnk.Prefix().Sum(hasher.Bytes())
+
+	cid, err := lnk.Prefix().Sum(hasherBytes)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


warpfork:
I'm gonna go ahead and attach some of the observation data that leads to this PR.  (It's always scary to attach profile data, because it's so easy for things to be data-dependent/workload-dependent, and to manage understandings and disclaimers around that -- and this data comes from downstream projects under workloads I don't totally understand, so those disclaimers are in force :shrug: -- but perfect data is impossible, so we might as well record and share what we've got!)

These are some allocation profile reports from an application using [Graphsync](https://github.com/ipfs/go-graphsync/), [Selectors](https://github.com/ipld/go-ipld-prime/tree/master/traversal/selector), and [go-ipld-prime-proto](https://github.com/ipld/go-ipld-prime-proto/), and engaged in the act of transferring around some unixfsv1 files.  There are two reports: one from before these changes, and one from after this change (and similar changes in the other libraries (e.g. https://github.com/ipld/go-ipld-prime-proto/pull/3)) :

Before:

profile001

After:

profile-optimized

So overall these changes to avoid multiple re-allocations and shufflings of data around reduce allocations by something like 50% in this example... a significant result!

mvdan:
I think my comment in https://github.com/ipld/go-ipld-prime-proto/pull/3#pullrequestreview-477579858 applies here too. If we need entire byte chunks anyway, I'm not sure why we're using readers and writers :)

warpfork:
My general rule of thumb has been to prefer `io.Reader` and `io.Writer` interfaces by default.

I figure that it's easier to drill an abstraction like that back out (or periscope through it, like this) later, than it is to try to add it in late to a stack of abstractions have developed where a streaming interface never existed.

And in theory, I'd like our tools to be as close as possible to being ready to work streamingly on very large data.  We regard 1-2MB as a "typical" block size in IPLD, but simultaneously our specs do not state that as an inherent limit to IPLD; and in practice, I'd _like_ to be able to stream gigabytes in one "block" (think: `jq`-like operations) without running into operational difficulties (even if a lot of blockchain/merkle-forest contexts don't lean directly on that size of data).  (All this is easier said than done given that a lot of our code works on node trees, and trees and streaming are not trivially best friends, but I still figure it's best to aim close wherever possible.)

Having an interface as the argument's type also means being able to do feature detection with more interface assertions to add features in the future without changing the function's interface.  Arguments of type `[]byte` do not allow that.

And while we _could_ do a pair of methods, one taking an `io.Reader` parameter and one taking `[]byte` instead, every time this comes up... I think we may want to avoid that.  x2'ing our exported method count once or twice is fine, but... it's better to resist whenever possible: there's enough other things already that are likely to prompt us in the direction of exploding the number of exported methods combinatorically.  (For example, I have a suspicion we're going to sprout another variant of marshal and unmarshal methods at some point in the future that have more parameters to help us support declarative configuration for positional ADL invocations that don't depend on schemas for the signalling.)

So it's certainly possible to consider dropping interfaces entirely and going for bare `[]byte`, but I'm a little wary of the idea.

linking/cid/cidLink.go
```diff
@@ -30,20 +30,30 @@ func (lnk Link) Load(ctx context.Context, lnkCtx ipld.LinkContext, na ipld.NodeA
...
-	//  (We still don't implement this by running the hash to the end first, because that would increase the high-water memory requirement.)
-	//   ((Which we experience right now anyway because multihash's interface is silly, but we're acting as if that's fixed or will be soon.))
-	//  If the hash is rejected, we should return that error (and even if there was a decodeErr, it becomes irrelevant).
-	if decodeErr != nil {
-		_, err := io.Copy(&hasher, r)
-		if err != nil {
-			return err
+	var hasherBytes []byte
+	var decodeErr error
+	byteBuf, ok := r.(*bytes.Buffer)
```
warpfork, line 35: I feel a little dissatisfied making a concrete type assertion like this... but do you suppose we could accomplish the same thing with a small feature detection branch using a tiny interface?

Something like:

```
type Byter interface {
    Bytes() []byte
}
if byter, ok := r.(Byter) { // ...
```

This should match `bytes.Buffer` and do the right thing, I think, and probably costs 10ms more to check but that shouldn't bother :)

mvdan: I agree with this; the extra cost should be negligible, compared to the whole allocate + copy branch.

mvdan:
> So it's certainly possible to consider dropping interfaces entirely and going for bare `[]byte`, but I'm a little wary of the idea.

Gotcha, I generally agree. I raised the question because, in practice, the implementation is very closely tied to `[]byte`. If that will remain true for the foreseeable future, I'd probably replace `io.Reader` with `[]byte`, becuase otherwise we're lying to ourselves. But if we do plan to actually stop using `io.ReadAll` sometime soon, then I agree that keeping `io.Reader` is good.

hannahhoward:
Hey kids I'll make this change for a general interface -- I'm not trying to resolve the larger question about io.Reader right now. Then I'm gonna go head and merge.

hannahhoward:
NM -- you can merge on your own time but I'm going to go ahead and make the change then cherry pick since I'm not updating ipld prime at this time

hannahhoward:
also, I would not use the implementation of `go-ipld-prime-proto` which is a severely undeveloped project, as a reference for "intended behavior" -- I'm not sure whether io.ReadAll is used elsewhere, but just want to clarify it's a bad example :)

That said, in practice, the whole graphsync protocol, which is one of the main clients of go-ipld-prime, is fairly tied to []byte at the protocol level, and I don't see that changing.

warpfork:
Okay, I'm happy to merge this.  It does what it says, it's not overly specific to a concrete type, the name of the feature detection interface is purely internal and doesn't even matter; this is fine.

PR #72, opened 2020-08-31
CI: switch from Travis to Actions

mvdan:
They're actively maintained, faster on Mac and Windows, and generally
a better mechanism long-term.

The only change applied here, in practice, is that we now test on Mac
and Windows as well.


mvdan:
Hm, this got blocked by the actions settings: https://github.com/ipld/go-ipld-prime/actions/runs/232689446

So it will have to wait until I'm given full access to this repo, I think.

mvdan:
@warpfork ok, this should be green now. I had to disable windows for now; I can take a look at that soon-ish, assuming that we do care about windows support.

warpfork:
I'm curious why windows support didn't work, but not especially so.  Whatever it is, it's almost certainly a CI issue rather than something wrong with the library; there's very little (if anything) that's platform specific going on in this entire repo.

(Aside from the tests for codegen, which do some [Interesting things](https://github.com/ipld/go-ipld-prime/blob/09fc5cd8fc2378e9f949e63ee8d6d74a7d289d05/schema/gen/go/HACKME_testing.md#behavioral-tests-are-run-via-plugins).  I wonder if the `skipgenbehavtests` build tag will be needed on windows -- probably.  Some patches to the tags in that area could also be in order.)

So: happy to merge :)

PR #75, opened 2020-09-05
Update tests for unions; several fixes

warpfork:
Update the tests targeting codegen'd unions to use the new test helper system introduced in https://github.com/ipld/go-ipld-prime/pull/66 .

The increased coverage found an infinite loop in iterators, so the fix for this is included.  Similarly, a bug in the assembler's reset mechanism was found, and is fixed.

With these two fixes, I'm also coincidentally able now able to use codegen'd unions in some pretty interesting larger demos without incident.  More news on that shortly ;)

### DIFF
--- a/schema/gen/go/testUnionsKinded_test.go
+++ b/schema/gen/go/testUnionsKinded_test.go
@@ -3,11 +3,7 @@ package gengo
 import (
 	"testing"
 
-	. "github.com/warpfork/go-wish"
-
 	"github.com/ipld/go-ipld-prime"
-	"github.com/ipld/go-ipld-prime/fluent"
-	"github.com/ipld/go-ipld-prime/must"
 	"github.com/ipld/go-ipld-prime/schema"
 )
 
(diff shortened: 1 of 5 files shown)


schema/gen/go/testUnionsKinded_test.go
```diff
@@ -35,36 +31,46 @@ func TestUnionKinded(t *testing.T) {
 		}),
 	))
 
+	// These are the same *type-level* as in TestUnionKeyedComplexChildren,
+	//  but (of course) have very different representations.
```
warpfork, line 35: I've been entertaining the thought of putting some methods on this `testcase` structure so that one can "derive" variations of them with only some fields changed, which could be useful for tests like this that poke at variations in representation.

Could be a future work thing.

schema/gen/go/testUnions_test.go
```diff
@@ -162,3 +178,77 @@ func TestUnionKeyedComplexChildren(t *testing.T) {
...
+			},
+			reprPoints: []testcasePoint{
+				{"0/b/q", "one"},
+				{"1/b/q", "two"},
+				{"2/a", "three"},
+			},
+		},
+	}
+
+	test := func(t *testing.T, getPrototypeByName func(string) ipld.NodePrototype) {
```
warpfork, line 228: These setup blocks (this and a dozen or so lines below) are getting somewhat redundant.  It might soon be time to try to extract something here.

I've been wary to do this prematurely, and it'll be a bit tricky because reasons for doing distinct gen invocation are various (e.g. in these tests, it's always poking `CfgUnionMemlayout`; but that's not true in other tests which don't show up in this diff).  But it's getting to the point it might be worth some looking into soon.

schema/gen/go/testUnions_test.go
```diff
@@ -162,3 +178,77 @@ func TestUnionKeyedComplexChildren(t *testing.T) {
...
+		np := getPrototypeByName("OuterList")
+		nrp := getPrototypeByName("OuterList.Repr")
+		for _, tcase := range specs {
+			tcase.Test(t, np, nrp)
+		}
+	}
+
+	t.Run("union-using-embed", func(t *testing.T) {
```
mvdan, line 236: nitpicking some more, but these should be CamelCase for the sake of consistency. There's are some examples in https://golang.org/pkg/testing/ too.

schema/gen/go/testUnions_test.go
```diff
@@ -162,3 +178,77 @@ func TestUnionKeyedComplexChildren(t *testing.T) {
 		})
 	})
 }
+
+// TestUnionKeyedReset puts a union inside a list, so that we can use the list's reuse of assembler as a test of the assembler's reset feature.
+// The value inside the union is also more complex than a scalar value so that we test resetting gets passed down, too.
```
mvdan, line 183: so that we test *that* resetting...

I'm pretty sure, but not 100% sure :)

schema/gen/go/testUnions_test.go
```diff
@@ -107,40 +116,47 @@ func TestUnionKeyedComplexChildren(t *testing.T) {
...
 
+	specs := []testcase{
+		{
+			name:     "InhabitantA",
+			typeJson: `{"String":"whee"}`,
+			reprJson: `{"a":"whee"}`,
+			typePoints: []testcasePoint{
+				{"", ipld.ReprKind_Map},
+				{"String", "whee"},
+				//{"SmolStruct", ipld.ErrNotExists{}}, // TODO: need better error typing from traversal package.
```
mvdan, line 127: nit: link to the issue in these TODOs? that way when the issue gets fixed, it's trivial to find the spots that can be improved

mvdan (approved):
take my approval with a grain of salt, as I'm still getting used to the codebase. but generally LGTM

PR #76, opened 2020-09-06
schema-schema codegen demo now includes unmarshal exercise

warpfork:
Like it says on the tin: there's now a test which exercises unmarshalling the schema-schema into the codegen'd types we made to represent the schema types.

First there were some minor fixups to the instructions to the codegen (which are still expressed in our placeholder golang constructors):

- it now uses kinded unions (since those are supported now; they weren't yet the last time we touched this demo!).
- a few typos are fixed (field names; in one case, a place that must support type names as well as inline defns) that were uncovered while lining up this test.
- added the 'Schema' type at the very top!  this is what parsing the schema-schema starts with!

Then, rigging up the unmarshal of the schema-schema json is... pretty trivial, really.  (Goal success!)

There are also some alterations to the schema-schema json I've made here.  You can find comments about this in the diff body.  Some of these changes may turn into PRs upstream to the schema-schema document in the specs repo.


schema/gen/go/_demo/omg-schemas/schema/parse_test.go
```diff
@@ -0,0 +1,677 @@
...
+//  - a 'Unit' type is introduced (and might belong in the prelude!).
+//  - enums are specified using the 'Unit' type (which means serially, they have `{}` instead of `null`).
+//
+// There's also a couple errata in the programmatic definitions of the types we used for codegen
+// where further thought generated an idea which may generate schema-schema changes,
+// but these are mostly naming and organizational, and so are minor.
+//
+func TestSchemaSchemaParse(t *testing.T) {
+	nb := Type.Schema__Repr.NewBuilder()
+	if err := dagjson.Decoder(nb, bytes.NewBufferString(`
```
mvdan, line 34: surely this could be just `strings.NewReader`?

schema/gen/go/_demo/omg-schemas/schema/parse_test.go
```diff
@@ -0,0 +1,677 @@
...
+				},
+				"representation": {
+					"map": {}
+				}
+			}
+		}
+	}
+}
+	`)); err != nil {
+		panic(err)
```
mvdan, line 675: t.Fatal

mvdan (approved):

PR #77, opened 2020-09-07
WIP: schema/gen/go: make all top-level tests parallel

mvdan:
They do quite a lot of work, including setting up a type system,
generating Go code, and building it.

This package is by far the slowest to test. On a warm cache, 'go test
-count=1 ./...' shows all packages under 0.2s, while this one sits at
~1.5s.

1.5s is not unreasonable with a warm cache, but we can bring that down
to 0.6s on my quad-core laptop by just making all tests parallel. Note
that sub-tests aren't parallel for now, since there are multiple layers
of them and I'm not sure I even follow what's going on.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,8 +6,9 @@ jobs:
       matrix:
         go-version: [1.14.x, 1.15.x]
         # TODO(mvdan): fix and enable windows
+        # TODO(mvdan): mac seems to timeout too often
         # os: [ubuntu-latest, macos-latest, windows-latest]
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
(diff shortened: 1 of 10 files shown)


mvdan:
Oh, and the test time after deleting the `_test` directory goes down from ~6.7s to ~2.6s, so both are more than halved.

warpfork:
Yayy for fast!

mvdan:
I had removed mac in a later push, but forgot to remove the WIP from the title :) in any case, this was ready, thanks for merging

warpfork:
Oh whoops, I glazed right over that -- figured crossing out the mac builds was intentional since we'd also been talking about that elsewhere.  Ah well :) the "WIP" string didn't land in commit history this way either, haha.

PR #82, opened 2020-09-22
Security fix: patch crashing inputs

hannahhoward:
# Goals

See this report https://github.com/LeastAuthority/go-ipld-prime/issues/2

The Filecoin security audit found some values that can be passed to dagcbor's decode that will produce a panic. (discovered while fuzzing graphsync, which uses IPLD's builders for some of it's work)

This fix was created by the auditors -- I just rebased it on master.

### DIFF
--- a/codec/dagcbor/unmarshal.go
+++ b/codec/dagcbor/unmarshal.go
@@ -41,7 +41,7 @@ func unmarshal(na ipld.NodeAssembler, tokSrc shared.TokenSource, tk *tok.Token) 
 	case tok.TMapOpen:
 		expectLen := tk.Length
 		allocLen := tk.Length
-		if tk.Length == -1 {
+		if tk.Length == -1 || tk.Length > math.MaxInt32 {
 			expectLen = math.MaxInt32
 			allocLen = 0
 		}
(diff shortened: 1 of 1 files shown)


rvagg:
IMO this should result in a "bad block" error rather than softly accepting it as an indefinite length; even indefinite length isn't proper dag-cbor and should be rejected, but we don't do that anywhere else. In this case, the number is far too big to be a valid length, it's not indefinite as per CBOR spec, it's simply invalid and should result in a `fmt.Errorf("malformed data")` of some kind rather than massaging it into an acceptable form. A source generating bad data like this isn't trustworthy.

rvagg:
Here's the minimal crashing case fwiw `0xbb3030303030303030`, this is on the very end of that fuzz input which is mostly just a nested set of maps with this one last.

```
bb 3030303030303030                               # map(3472328296227680304)
```

mikeal:
I agree with @rvagg, we should throw on input that is invalid. The reason it’s invalid doesn’t matter too much, if it’s invalid we shouldn’t accept it. The fact that this is a length that is impossibly large makes me think this *may* be exploitable in a way we haven’t thought of yet that would be more concerning than the current behavior.

warpfork:
Hey @hannahhoward and @bryanchriswhite , thanks for bringing our attention to this.

I think I've addressed the underlying issue in a more complete way in https://github.com/ipld/go-ipld-prime/pull/85 , which adds resource limits.  I believe it should fix the case this diff was guarding against, as well as several more.

I'll close this now but please let me know if we find a case that still isn't handled gracefully!

PR #96, opened 2020-10-19
Make the cidlink.Link type only usable as a pointer.

warpfork:
Totally changed this PR.  Turns out we actually really don't want links to be pointers in many cases.  The behavior of the Link interface when used as a map key is important.

~As the comments in the diff say: it's a fairly sizable footgun for users to need to consider whether they expect the pointer form or the bare form when inspecting what an `ipld.Link` interface contains: so, let's just remove the choice.~

~There's technically no reason for the Link.Load method to need to be attached to the pointer receiver other than removing this footgun.  From the other side, though, there's no reason *not* to make it attached to the pointer receiver, because any time a value is assigned to an interface type, it necessarily heap-escapes and becomes a pointer anyway.  So, making it unconditional and forcing the pointer to be clear in the user's hands seems best.~

### DIFF
--- a/linking/cid/cidLink.go
+++ b/linking/cid/cidLink.go
@@ -15,6 +15,13 @@ var (
 	_ ipld.LinkBuilder = LinkBuilder{}
 )
 
+// Link implements the ipld.Link interface using a CID.
+// See https://github.com/ipfs/go-cid for more information about CIDs.
+//
+// When using this value, typically you'll use it as `Link`, and not `*Link`.
+// This includes when handling the value as an `ipld.Link` interface -- the non-pointer form is typically preferable.
+// This is because the ipld.Link inteface is often desirable to be able to use as a golang map key,
+// and in that context, pointers would not result in the desired behavior.
 type Link struct {
 	cid.Cid
 }
--- a/linking/cid/cidLink.go
+++ b/linking/cid/cidLink.go
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	_ ipld.Link        = &Link{}
+	_ ipld.Link        = Link{}
 	_ ipld.LinkBuilder = LinkBuilder{}
 )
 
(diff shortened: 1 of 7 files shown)


mvdan (approved):

warpfork:
I was going to claim that this should be an easy low-impact change, since few downstream consumers are likely to depend on whether or not this type is typically handled as a pointer... but on further review, I'm not entirely sure.  More examples than I realized have been suggesting the non-pointer form of use.  Hm.

rvagg (approved):

warpfork:
This may have been a bad idea.

It's easy to forget this (and I apparently just did), but we often use link types as map keys.

That means trying to handle them as pointers leads to nonsensical results: pointers are technically valid as a golang map key, but they don't "do the right thing" -- the equality check ends up operating on the the pointer rather than on the data.

I think we might have to abandon this and instead launch some more docs on the Link interface that talk about this.

mvdan:
Then go the opposite route and suggest to *never* use pointers with them? I assume that's what you mean with the doc change.

warpfork:
I've revamped this to be a docs change.

linking/cid/cidLink.go
```diff
@@ -15,6 +15,13 @@ var (
 	_ ipld.LinkBuilder = LinkBuilder{}
 )
 
+// Link implements the ipld.Link interface using a CID.
+// See https://github.com/ipfs/go-cid for more information about CIDs.
+//
+// When using this value, typically you'll use it as `Link`, and not `*Link`.
+// This includes when handling the value as an `ipld.Link` interface -- the non-pointer form is typically preferable.
```
mvdan, line 22: nit: what I mentioned about markdown :)

linking/cid/cidLink.go
```diff
@@ -15,6 +15,13 @@ var (
...
+// This is because the ipld.Link inteface is often desirable to be able to use as a golang map key,
```
mvdan, line 23: I'd remove "golang". this is a godoc, so "map key" is already pretty obvious.

mvdan (approved):

linking/cid/cidLink.go
warpfork: disagree.  Links aren't usable as an IPLD map key, so the potential for ambiguity and confusion here is actually quite high.

PR #104, opened 2020-10-29
first pass at graphql schema generation

willscott:
* Generates a read-only query schema document over an IPLD type system.

Choices made:
* links are collapsed / made implicit in the graphql schema
* graphql has a single opaque `scalar` representation with predefined (bool, int, float, string, id) types that clients will use default serialization and de-serialization on (with caveats like that int may become unsafe outside of int32 sizes)
* there is no map type natively. This codegen generates an interface instead with accessors either to get a specific value by key, or to browse the list of Key-Value tuples. Probably that will eventually extend to support range queries.
* unions cannot include primitive types. We make create a wrapped struct type for primitives that are attempted to be union options


warpfork:
This is really fascinating!

I appreciate the summary of design choices here in the PR.  Some of them are very interesting.  I think they sound reasonable to me, but I also haven't worked on anything GraphQL for long enough that I'm not sure all the implications are immediately clear, so I'll look forward to learning more about this in the future.

Probably would be good to put the design choices info into the diff as a document there too?

Do you have some examples of what some of the output of this looks like, and then what a query of it would look like, or how other graphQL tools would render it?  A "happy path" example would be good, but I'm also really curious what the map and union-with-primitives would look like.

willscott:
Here's the current output from the filecoin schema:

https://gist.github.com/willscott/d901f581075261b1d6a07a7990ae7b9f

that can be copy-pasted into https://apis.guru/graphql-voyager/ to get a graphical view

willscott:
Migrated to the standalone [go-ipld-graphql](https://github.com/ipld/go-ipld-graphql) repo

Issue #93, opened 2020-10-14
schema/gen/go: produce a single Go file


mvdan:
Right now, we generate a separate Go file for each schema type, plus a few extra files like `minima.go` and `typeTable.go`. We could really do with far fewer files, even a single one for the entire generated code.

Here are reasons why a linear amount of files isn't a good idea:

* Removing a schema type and re-running `go generate` leaves orphaned files behind.
* Directories get cluttered, making Go development harder. It even pushes me to generate code in a separate internal package.
* Many of the generated files are tiny; in my current HAMT ADL code, I see two under 50, and a handful of others around 200 lines.

In the future, if the generator learns to parse schema files, we could instead produce Go files alongside them or following their name. For example, `foo.proto` produces `foo.pb.go` with `protoc-gen-go`, and we could similarly use a standard like `ipld_gen_${name}.go`.

/cc @warpfork @aschmahmann

warpfork:
- I definitely agree with getting rid of the linear number of files in the number of types.  That's just silly.  It's a first-draft hangover and indefensible.  :+1: 
- A CLI tool which takes hints for output filenames from the schema file it reads sounds perfectly reasonable. :+1: 
- I'm indifferent, maybe softly negative, to combining the type table and the minima files together with the rest of the gen output. :shrug: 
- There's probably a lot of different ways we could satisfy "output file count should not be linear in count of types" while still having some files split for readability?  I'd like to investigate some. :mag: 

One idea (I've had comments outlining this on my git stash for a while, but not implemented):

- a file for the data structures
    - this is the file a go developer is most likely to actually _look_ at, to see how things got laid out in memory.
    - maybe this file is even worth generating some (brief) comments in (by default; toggleable with configuration).
- a file for all the boilerplate associated with each type
    - the node contract methods
    - the representation type and representation methods
    - any native/"speciated" methods for the type
        - (in _some_ cases it might be worth putting these in the data structure file -- for example structs with stringjoin representations; it's nice to see it when there's such a distinctive behavior that isn't directly implied by the go structure.  maybe comments would be enough, alternatively.)
    - the maybeT for the type
    - the assemblers
    - the representation assemblers
- a file for the type table
    - ... or, this just goes at the top of the file full of datastructures.  (that would be in keeping with the theme of "that's the file to look at".)
- a file for the minima
    - just because it's basically static

(I'm not entirely attached to this; it's just an option.)

To actually do this:

[...]

PR #105, opened 2020-10-30
Codegen output rearrange

warpfork:
Rearrange codegen output into finite number of files.

Also, emit some comments around the type definitions.

The old file layout is still available, but renamed to GenerateSplayed.  It will probably be removed in the future.  (Maybe it should be removed now, for that matter.)

All the files produced now match the pattern "`ipldsch.*.gen.go`".  There's file for the common bits that need to appear in every codegen output package; one file for all the type definitions, including the type/prototype table (I figure this is what a developer will hit most often during quick reference); and one file for everything else (including the MaybeT types, the Node implementations, the NodeAssembler implementations, and all that again for the representation level).

This could be a resolution for https://github.com/ipld/go-ipld-prime/issues/93 .

### DIFF
--- a/schema/gen/go/adjunctCfg.go
+++ b/schema/gen/go/adjunctCfg.go
@@ -65,6 +65,11 @@ func (cfg *AdjunctCfg) FieldSymbolUpper(f schema.StructField) string {
 	return strings.Title(f.Name())
 }
 
+// Comments returns a bool for whether comments should be included in gen output or not.
+func (cfg *AdjunctCfg) Comments() bool {
+	return true // FUTURE: okay, maybe this should be configurable :)
+}
+
 func (cfg *AdjunctCfg) MaybeUsesPtr(t schema.Type) bool {
 	if x, ok := cfg.maybeUsesPtr[t.Name()]; ok {
 		return x
--- a/schema/gen/go/generate.go
+++ b/schema/gen/go/generate.go
@@ -1,6 +1,7 @@
 package gengo
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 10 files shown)


schema/gen/go/generate.go
```diff
@@ -1,14 +1,133 @@
...
+			tg.EmitNativeType(w)
+			fmt.Fprintf(f, "\n")
+		}, f)
+
+	})
+
+	// Emit a file with all the Node/NodeBuilder/NodeAssembler boilerplate.
+	//  Also includes typedefs for representation-level data.
+	//  Also includes the MaybeT boilerplate.
+	withFile(filepath.Join(pth, "ipldsch.satisfaction.gen.go"), func(f io.Writer) {
```
warpfork, line 99: This filename in particular is bit placeholder.  Other suggestions welcome.

schema/gen/go/adjunctCfg.go
```diff
@@ -65,6 +65,11 @@ func (cfg *AdjunctCfg) FieldSymbolUpper(f schema.StructField) string {
 	return strings.Title(f.Name())
 }
 
+// Comments returns a bool for whether comments should be included in gen output or not.
+func (cfg *AdjunctCfg) Comments() bool {
+	return true // FUTURE: okay, maybe this should be configurable :)
+}
```
mvdan, line 71: Do we really need this? It's trivial to remove comments from Go source code by just using `go/parser` with the comment flag disabled, and re-printing the program. Or, if the code is simple enough, a bit of sed.

schema/gen/go/generate.go
```diff
@@ -1,14 +1,133 @@
...
+			fmt.Fprintf(f, "\n")
+		}, f)
+	})
+}
+
+// GenerateSplayed is like Generate, but emits a differnet pattern of files.
+// GenerateSplayed emits many more individual files than Generate.
+//
+// This function should be considered deprecated and may be removed in the future.
+func GenerateSplayed(pth string, pkgName string, ts schema.TypeSystem, adjCfg *AdjunctCfg) {
```
mvdan, line 130: I'd  personally remove it directly. The old code will always be in the git history if we wish to reuse it.

schema/gen/go/generate.go
```diff
@@ -1,14 +1,133 @@
...
+		types := ts.GetTypes()
+		keys := make(sortableTypeNames, 0, len(types))
+		for tn := range types {
+			keys = append(keys, tn)
+		}
+		sort.Sort(keys)
+		for _, tn := range keys {
+			switch t2 := types[tn].(type) {
+			case *schema.TypeBool:
+				fn(NewBoolReprBoolGenerator(pkgName, t2, adjCfg), f)
```
mvdan, line 37: this code feels repetitive to me; it seems like  you could do:

```
var gen func(...)
switch t2 := types[tn].(type) {
case *schema.TypeBool:
	gen = NewBoolReprBoolGenerator
case ...
}
fn(gen(pkgName, t2, adjCfg), f)
```

schema/gen/go/generate.go
```diff
@@ -1,14 +1,133 @@
...
+					panic("unrecognized union representation strategy")
+				}
+			default:
+				panic("add more type switches here :)")
+			}
+		}
+	}
+
+	// Emit a file with the type table, and the golang type defns for each type.
+	withFile(filepath.Join(pth, "ipldsch.types.gen.go"), func(f io.Writer) {
```
mvdan, line 79: not sure I like "ipldsch", it reads like some germanic gibberish to me :) 

how about just ipldschema_types.go? it's self-describing, not too long, and honestly I don't think we need the `gen.go` suffix if we already have an `ipldschema_` prefix. underscores before the extension dot are also generally more common.

mvdan (approved):
Emitting three files feels like a considerable step in the right direction :) Left some thoughts.

schema/gen/go/generate.go
warpfork: I thought (but maybe I imagined) I'd seen a convention ".gen" as a prefix to the file extension suffix as a semi-universal hint that a file is generated.  I wish some convention of that existed, anyway.  But maybe that was wishful thinking?

mvdan: The convention is the "Code generated by..." comment, and nothing else, really. Take stringer, which generates `yourtype_string.go`.

I think a prefix like `ipldschema_` or `ipldsch_` is already extremely clear, to be honest. Adding `.gen` just adds verbosity for a convention that doesn't exist :)

willscott:
😱 it looks like this merged a bit too early. I'm seeing the following error on codegen output now:
in ipldsch_types.go, there are references to `x ipld.Link` within link types, but no `import()` of `ipld` at the top of the file.

Issue #22, opened 2019-07-22
The naming of Node's single step "traversal" methods


warpfork:
The `ipld.Node` interface currently has a method `TraverseField(key string) (ipld.Node, error)`.  This has been sufficient to get going, but there are several issues with this: it's verbose; it always takes a bare string (which is sometimes not going to be ideal); and the name "field" is usually only used in the context of structs, but we're also using it here in describing maps, which is thus awkward.  Should we rename this?  Furthermore, should we have several methods here?

It would seem we want all of these methods at some point, and since Go eschews parametric polymorphism, they all have to have distinct names:

- `TraverseByString(fieldname string)` -- for use with maps
- `TraverseByIndex(idx int)` -- for use with lists
- `TraverseBySegment(segment IntOrString)` -- for use when you've got pathsegments (selectors do this rather than the above two)
- `TraverseByNode(node Node)` -- for generic use that's not segment-powered (retains `typed.Node` reference if applicable: preferable if you've got a struct-represented-as-string as a map key, or anything that has a `Validate()` method on it that would need to be reapplied if we forgot about associated type info).
- `Traverse(x ConcreteGeneratedType)` -- for use when codegen is in flight (similar advantages to TraverseByNode).

If that above fivesome goes through, the first four would on the Node interface now (contrast currently only two) (the fifth is only for codegen).

Questions:

- Is that fivesome correct?
- Can we come up with shorter names?  "Traverse" is a bit wordy when using it so much (and, I tend to think of "traverse" as being a multi-step process (though I figure on renaming any multi-step functions to be "walk", also, to avoid ambiguity)).
- Alternative phrasing for the "ByFoo" suffix would also be welcome.

I've been thinking about this (but pushing it off) for a while now, but if something is to be done here, it should probably be a clean break soon, before we start getting too many additional things depending on this API.

warpfork:
By the same logic, `MapBuilder.Insert(k, v ipld.Node) error` also should have a variant taking primitive args for convenience: `MapBuilder.Insert2(k string, v ipld.Node) error`.  Similar naming questions apply there, then.

rvagg:
Is it really necessary to have `TraverseBySegment(segment IntOrString)`? Or conversely, doesn't that one serve the same purpose as `TraverseByString` and `TraverseByIndex` combined? Seems like redundancy, or at the least, an API documentation smell.

[...]

PR #106, opened 2020-10-30
Codegen various improvements

warpfork:
Bulk PR for a bunch of individually minor codegen improvements.

- More Lookup methods on typed maps which are only available when concrete types are known, but keep you in the realm of concrete types.
- New Iterators on typed maps with the same concreteness.
- More Lookup methods on typed lists with ... you get the idea.  Concreteness.  Less casting in your future when you use these.
- New Iterators on typed lists with the same concreteness.
- Some other misc bits, like: the schema-schema demo had forgotten "stringprefix" in union strategies, and now it has them.

It's basically a laundry list of small quality-of-life things that I encountered today while working on other things that consume codegen'd code.

(These aren't particularly interesting, so I'll probably merge this with or without review, but if anyone wants to give it a quick eyeball, have at it.)

### DIFF
--- a/schema/gen/go/genList.go
+++ b/schema/gen/go/genList.go
@@ -29,7 +29,74 @@ func (g listGenerator) EmitNativeType(w io.Writer) {
 }
 
 func (g listGenerator) EmitNativeAccessors(w io.Writer) {
-	// FUTURE: come back to this -- surely something nice can be done here.
+	// Generate a speciated Lookup as well as LookupMaybe method.
+	// The Lookup method returns nil in case of *either* an out-of-range/absent value or a null value,
+	//  and so should only be used if the list type doesn't allow nullable keys or if the caller doesn't care about the difference.
+	// The LookupMaybe method returns a MaybeT type for the list value,
+	//  and is needed if the list allows nullable values and the caller wishes to distinguish between null and out-of-range/absent.
+	// (The Lookup method should be preferred for lists that have non-nullable keys, because LookupMaybe may incur additional costs;
+	//   boxing something into a maybe when it wasn't already stored that way costs an alloc(!),
+	//    and may additionally incur a memcpy if the maybe for the value type doesn't use pointers internally).
+	doTemplate(`
+		func (n *_{{ .Type | TypeSymbol }}) Lookup(idx int) {{ .Type.ValueType | TypeSymbol }} {
+			if n.Length() <= idx {
+				return nil
+			}
+			v := &n.x[idx]
+			{{- if .Type.ValueIsNullable }}
+			if v.m == schema.Maybe_Null {
+				return nil
+			}
+			return {{ if not (MaybeUsesPtr .Type.ValueType) }}&{{end}}v.v
+			{{- else}}
+			return v
+			{{- end}}
+		}
+		func (n *_{{ .Type | TypeSymbol }}) LookupMaybe(idx int) Maybe{{ .Type.ValueType | TypeSymbol }} {
+			if n.Length() <= idx {
+				return nil
+			}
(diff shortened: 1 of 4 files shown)


schema/gen/go/genList.go
```diff
@@ -29,7 +29,74 @@ func (g listGenerator) EmitNativeType(w io.Writer) {
...
+	// Generate a speciated Lookup as well as LookupMaybe method.
+	// The Lookup method returns nil in case of *either* an out-of-range/absent value or a null value,
+	//  and so should only be used if the list type doesn't allow nullable keys or if the caller doesn't care about the difference.
+	// The LookupMaybe method returns a MaybeT type for the list value,
+	//  and is needed if the list allows nullable values and the caller wishes to distinguish between null and out-of-range/absent.
+	// (The Lookup method should be preferred for lists that have non-nullable keys, because LookupMaybe may incur additional costs;
+	//   boxing something into a maybe when it wasn't already stored that way costs an alloc(!),
+	//    and may additionally incur a memcpy if the maybe for the value type doesn't use pointers internally).
+	doTemplate(`
+		func (n *_{{ .Type | TypeSymbol }}) Lookup(idx int) {{ .Type.ValueType | TypeSymbol }} {
```
willscott, line 41: thought on naming: i might more naturally think of this at `Get` rather than `Lookup`

willscott (approved):

schema/gen/go/genList.go
warpfork: [I thought about this for a surprisingly long time](https://github.com/ipld/go-ipld-prime/issues/22). At this point, I'd say either all the methods on the `Node` interface turn into `Get*` and this goes with them, or, none of them do.  Having the gen'd methods with concrete types have a totally different name prefix than the generic interface methods was the one thing that was definitely a mistake.

mvdan:
A bit of a tangent, but - have you thought about code bloat in general? The current schema generator already outputs quite a lot of code, and we seem to be adding more and more, so I worry that this will result in binaries getting heavier and builds getting slower over time.

One of the main reasons protobuf had a major breaking version last year, I seem to recall, was to remove a lot of non-essential cruft that wasn't really needed and was bloating projects. It cost them a lot of effort to break away from that, so I just want to make sure we're not headed for the same path.

warpfork:
Yes.  The size of output is currently pretty out of control.

There's a massive amount of low-hanging fruit here.  (I hope.)  I've made no attempts to minimize it yet.  It's getting to be time to make some efforts on that.

(This'll be truly into tangent land, but, fwiw: I think the two biggest things to chase first will probably be: all the state machine switches around `ma.state`, which are very redundant right now; and a bunch of `AssignNode` bodies; maybe third is `AssignNull` bodies.  It would also certainly be nice to be able to get rid of all the methods that just have ErrWrongKind stubs; but doing that without dropping type information is a research problem (I think we have an issue about this somewhere; I'm not even sure it's possible).  I haven't counted up the effect of any of these carefully, mind; these are just my guesses at where to start.  Stats tools for size, both GSLOC and compiled binary, would also be a great step.)

These iterators are definitely going to be worth it either way, though.   I tried to write some code without them, and it was pretty horrendous and involved a lot of cast assertions, and the lack of autocomplete made for very poor UX.

warpfork:
Merging; want to launch another PR which builds upon this, momentarily

mvdan:
> It's getting to be time to make some efforts on that.

Could you open this as a separate issue, so I have a place to collect my thoughts alongside yours?

warpfork:
New issue for collecting codegen output size notes: https://github.com/ipld/go-ipld-prime/issues/108

PR #110, opened 2020-11-13
Introduce traversal function that selects links out of a tree.

warpfork:
Like it says on the tin.

This is just a quick, simple, handy function to have on hand.

### DIFF
--- a/traversal/select_links.go
+++ b/traversal/select_links.go
@@ -1,0 +1,54 @@
+package traversal
+
+import (
+	"github.com/ipld/go-ipld-prime"
+)
+
+// SelectLinks walks a Node tree and returns a slice of all Links encountered.
+// SelectLinks will recurse down into any maps and lists,
+// but does not attempt to load any of the links it encounters nor recurse further through them
+// (in other words, it's confined to one "block").
+//
+// SelectLinks only returns the list of links; it does not return any other information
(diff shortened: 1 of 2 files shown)


warpfork:
It may be interesting to compare this to https://github.com/whyrusleeping/cbor-gen/blob/0b9f6c5fb1636544f94f5087817df99699de49ba/utils.go#L47 -- that function also plucks links out of a block.  It's also probably faster and lower overhead if you didn't already otherwise need to parse the block, because it scans over the serial form and *only* bothers to create in-memory objects for the links.  On the other hand, that function also only works with CBOR.

We might be able to create functions a bit closer to those semantics in a codec-agnostic way in go-ipld-prime in the future, too.  Some of the [tokenization reworking](https://github.com/ipld/go-ipld-prime/pull/101) aims to get us closer to able to handle Data Model info as a stream without full tree reification.  We'd need a fair amount of further rework to codec implementations to make that fly, though.

Regardless: this function that does operate on in-memory trees is still reasonable to have in the traversal package -- both immediately, because it's easy; and in the long run even if we did make stream methods with similar outputs, because which is useful is situationally dependent on how much parsing you may have already needed to do for other parts of the situation you're in.

ribasushi (approved):
LGTM! Agreed with motivation and limitations notes.

PR #111, opened 2020-11-14
Validate struct builder sufficiency

willscott:
Check that all struct fields that are required have been set when Build'ing the Node.

### DIFF
--- a/schema/gen/go/genStruct.go
+++ b/schema/gen/go/genStruct.go
@@ -507,7 +507,9 @@ func (g structBuilderGenerator) emitMapAssemblerMethods(w io.Writer) {
 			case maState_finished:
 				panic("invalid state: Finish cannot be called on an assembler that's already finished")
 			}
-			//FIXME check if all required fields are set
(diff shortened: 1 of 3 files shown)


schema/gen/go/genStruct.go
```diff
@@ -507,7 +507,9 @@ func (g structBuilderGenerator) emitMapAssemblerMethods(w io.Writer) {
 			case maState_finished:
 				panic("invalid state: Finish cannot be called on an assembler that's already finished")
 			}
-			//FIXME check if all required fields are set
+			if ma.s & fieldBits__{{ $type | TypeSymbol }}_sufficient != fieldBits__{{ $type | TypeSymbol }}_sufficient {
+				panic("invalid state: Not all required fields set")
```
warpfork, line 511: This should be an error return, not a panic.  Panics are reserved for when the builder has been used in a clearly invalid way (double finish, fed two keys in a row, things like that).  Data that's wanted but not present is just an error.

Bonus points if the error message can say which of the required fields are missing.

warpfork:
Sweet!

PR #115, opened 2020-11-17
add import to ipld in ipldsch_types.go

willscott:
cleanup from #105


mvdan (approved):

mvdan:
sgtm but this seems to suggest we need better tests :)

PR #116, opened 2020-11-18
Support overriden types

willscott:
This change will look at the destination package that codegen is being built into, and will skip generation of types that are already declared by files not prefixed with `ipldsch_`.

This isn't the cleanest escape-hatch, but it's a start.

### DIFF
--- a/schema/gen/go/externUtil.go
+++ b/schema/gen/go/externUtil.go
@@ -1,0 +1,40 @@
+package gengo
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"path"
+	"strings"
+)
+
+// getExternTypes provides a mapping of all types defined in the destination package.
+// It is used by generate to not duplicate defined types to allow overriding of types.
+func getExternTypes(pth string) (map[string]struct{}, error) {
+	set := token.NewFileSet()
+	packs, err := parser.ParseDir(set, pth, nil, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	types := make(map[string]struct{})
+	for _, pack := range packs {
+		for fname, f := range pack.Files {
+			if strings.HasPrefix(path.Base(fname), "ipldsch_") {
+				continue
+			}
+			for _, d := range f.Decls {
+				if t, isType := d.(*ast.GenDecl); isType {
+					if t.Tok == token.TYPE {
+						for _, s := range t.Specs {
+							ts := s.(*ast.TypeSpec)
+							types[ts.Name.Name] = struct{}{}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	return types, nil
+}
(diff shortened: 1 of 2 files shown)


schema/gen/go/externUtil.go
```diff
@@ -0,0 +1,40 @@
...
+	set := token.NewFileSet()
+	packs, err := parser.ParseDir(set, pth, nil, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	types := make(map[string]struct{})
+	for _, pack := range packs {
+		for fname, f := range pack.Files {
+			if strings.HasPrefix(path.Base(fname), "ipldsch_") {
```
mvdan, line 23: I know this is a generator and the performance isn't the most important thing, but parsing all files to then only look at a small minority is a bit wasteful :)

It would be significantly less work to use globbing to find the `ipldsch_*.go` files we're interested in, and then parse and inspect those.

schema/gen/go/externUtil.go
```diff
@@ -0,0 +1,40 @@
...
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"path"
+	"strings"
+)
+
+// getExternTypes provides a mapping of all types defined in the destination package.
+// It is used by generate to not duplicate defined types to allow overriding of types.
+func getExternTypes(pth string) (map[string]struct{}, error) {
```
mvdan, line 13: `overridenTypes`?

mvdan (commented):
Also, what happens if I run the generator twice in succession? Presumably the second time it will think that all the generated types are manually overriden? Perhaps you should only consider overrides when they're contained by a file not matching `^// Code generated .* DO NOT EDIT\.$`.

I think we all agree that this is not a good long-term design, but as far as quick hacks go, I think it's an easy win for now. We can delete this if/when we come up with something nicer.

schema/gen/go/externUtil.go
willscott: This is excluding the three `ipldsch_*.go` files and looking at all others to find types that are defined not in generated code. (this also addresses your other comment of "is it stable over multiple runs")

mvdan: :facepalm: I read the logic backwards. I guess my point here is still valid, though now it's just "you can avoid reading+parsing these 3-4 files".

mvdan:
One idea that came to me while you spoke at the standup: right now, if the code generator changes and you copied an older version of its generated code, you likely won't notice that you're using mismatching versions of generated code until they behave slightly wrong, or break in weird and confusing ways.

If we want to merge this, I'd like to have some sort of sanity check that all the generated types in a single package belong to the same version. Ideally at compile time, via constant expressions :)

schema/gen/go/externUtil.go
warpfork: `manualHijinx`?

mvdan:
> If we want to merge this, I'd like to have some sort of sanity check that all the generated types in a single package belong to the same version. Ideally at compile time, via constant expressions :)

Thinking about this again, it feels overkill given that it's a temporary hacky solultion, anyway. LGTM, and let's merge since my comments were minor too.

PR #118, opened 2020-12-01
schema/gen/go: remove two common subtest levels

mvdan:
Practically every subtest ends up at 7 or so levels of names, like:

	TestMapsContainingMaybe/maybe-using-ptr/generate/compile/bhvtest/non-nullable/typed-create

However, note that the "generate" and "compile" levels are always there,
so their presence just adds verbosity in the output and makes the
developer's life more difficult.

Extremely nested sub-tests are already rare, so at least we can just
keep the components that add useful information in the output.

"bhvtest" is also pretty redundant, but that one actually matters - its
subtest can be skipped depending on build tags.

### DIFF
--- a/schema/gen/go/testEngine_test.go
+++ b/schema/gen/go/testEngine_test.go
@@ -29,58 +29,54 @@ func genAndCompileAndTest(
 	adjCfg *AdjunctCfg,
 	testsFn behavioralTests,
 ) {
-	t.Run("generate", func(t *testing.T) {
-		// Make directories for the package we're about to generate.
-		//  Everything will be prefixed with "./_test".
-		// You can rm-rf the whole "./_test" dir at your leisure.
-		//  We don't by default because it's nicer to let go's builds of things cache.
-		//  If you change the names of types, though, you'll have garbage files leftover,
-		//   and that's currently a manual cleanup problem.  Sorry.
-		os.Mkdir("./_test/", 0755)
-		os.Mkdir("./_test/"+prefix, 0755)
+	// Make directories for the package we're about to generate.
+	//  Everything will be prefixed with "./_test".
+	// You can rm-rf the whole "./_test" dir at your leisure.
+	//  We don't by default because it's nicer to let go's builds of things cache.
+	//  If you change the names of types, though, you'll have garbage files leftover,
+	//   and that's currently a manual cleanup problem.  Sorry.
+	os.Mkdir("./_test/", 0755)
+	os.Mkdir("./_test/"+prefix, 0755)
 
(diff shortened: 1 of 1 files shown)


warpfork:
How would you feel about putting each of these three phases in a subtest, but sequentially, so they render as siblings rather than a deep tree?

I appreciate seeing which of the three major phases (generate, compile, exercise) passed and failed by skimming the headlines and wouldn't want to lose that.

mvdan:
I guess that would be a bit better with losing one level of indentation, but we'd have exactly the same amount of verbosity. I don't know about you, but I don't find `go test -v` printing nearly a thousand result lines useful :)

If you want to know at what stage a test was when it failed, I think the right approach is to just add that context to the error messages. Then you get the best of both worlds; `go test -v` is useful, but you still get enough context when something fails.

warpfork:
It's loud, but I definitely want that output (and `-v` is for verbose!).  It's very nice to be able to use a tool on the complexity level of grep to figure out if a test of interest has been run or not when someone submits a paste of output.  That's not always limited to negative confirmations; the positives are useful too ("hm, why did X fail if Y passed?") and I like to see these quickly and clearly.  This is especially true given there's build tags and the like going on in the vicinity making which tests are run very conditional.

mvdan:
But you can tell if a test ran via a single line per test case, not five of them, no? As far as I know you only need that step-by-step context when something fails.

warpfork (approved):
I've become comfortable with this.

Mechanically, the important thing is that running single tests should still work, and this does still satisfy that criteria.

I don't dislike the longer lines that state every step that happened.  But I do concede it's possible to argue they're redundant.  And after looking at more test failures while doing development recently: I do also concede that in practice, the text emitted by the go compiler when it's the middle of these three phases that's the error is perfectly clearly distinguishable (... by a human, anyway) from the error text emitted by the other phases (... and I don't think we really need it to be mechanically distinguishable).

So, alright :+1:

warpfork:
Hm, one more quick note.  This flies unambiguously because we never do the textual generation step without doing the compiler(-but-not-yet-use-plugins) step.

That's true as written now, but I wonder if we'll ever encounter an environment where we want to run tests but don't want to assume `go build` is on the $PATH and ready to be used.  If so, the 'compile' step would become conditional, and we'd want whether or not it occurred to be reported in test output again.

mvdan:
That's reasonable, though I struggle to imagine what environment would have `go test` but not `go build`. Maybe something like https://github.com/mvdan/dockexec with a Docker image that does not have Go installed? Seems pretty funky though, so I think we should worry about that if/when we need to.

PR #119, opened 2020-12-01
all: fix a lot of "unkeyed literal" vet warnings

mvdan:
Reduces the output of 'go vet ./...' from 374 lines to 96. Many warnings
remain, but I have lost my patience for today.

Most of the changes below were automated, especially the single-line
mixins expressions. Unfortunately, many of the Traits structs required
manual copy-pasting.

### DIFF
--- a/adl/rot13adl/rot13node.go
+++ b/adl/rot13adl/rot13node.go
@@ -41,16 +41,16 @@ func (*_R13String) ReprKind() ipld.ReprKind {
 	return ipld.ReprKind_String
 }
 func (*_R13String) LookupByString(string) (ipld.Node, error) {
-	return mixins.String{"rot13adl.R13String"}.LookupByString("")
+	return mixins.String{TypeName: "rot13adl.R13String"}.LookupByString("")
 }
 func (*_R13String) LookupByNode(ipld.Node) (ipld.Node, error) {
-	return mixins.String{"rot13adl.R13String"}.LookupByNode(nil)
+	return mixins.String{TypeName: "rot13adl.R13String"}.LookupByNode(nil)
 }
 func (*_R13String) LookupByIndex(idx int) (ipld.Node, error) {
-	return mixins.String{"rot13adl.R13String"}.LookupByIndex(0)
+	return mixins.String{TypeName: "rot13adl.R13String"}.LookupByIndex(0)
 }
 func (*_R13String) LookupBySegment(seg ipld.PathSegment) (ipld.Node, error) {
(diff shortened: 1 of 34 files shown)


mvdan:
Also, if we don't like the verbosity of the mixins, we could always do `type Bool string` instead of `type Bool struct { TypeName string }`.

warpfork:
For the record, I sort of hate this.

- Every invocation of the `node/mixins` gets about 20% longer?
- Yeah, we could change the type kind of them to not be structs... but it seems really odd to me that a linter to would give me an incentive to change the kind of type that I use when there is otherwise no incentive based on functional effect.
- The rule is apparently to do this only for structs that are exported?  So if the goal of this vet rule is readability, then, is it implied that readability of package internal structs is less relevant?  I don't really understand the consistency of logic in this.  I don't think whether a symbol is exported really has much bearing at all on how much readability assistance it needs.
- Generally, I don't see what the explicitness buys in any of the cases that show up in this diff.
  - It's not like the `node/mixins` types are going to gain another field; if they got larger it would betray their purpose in a way that's a much bigger problem than the mere syntax of updating.  (And if they did change: still: the syntax is not going to be the limiting factor in dealing with that kind of change.)
  - The `gen/go/mixins` are essentially "friends" of the `gen/go` package (arguably, they could be thrown under an "internal" package tree without ill effect)... I don't hate the added explicitness around their use, but it also doesn't seem very interesting.
- All the uses of the `gen/go/mixins` types previously would break at compile time if you added a new field (which I'd actually _like_, in this case), and with this change they would _not_ break at compile time but instead just contain uninitialized fields.  This seems worse?  At the very least, it's different, and not clearcut better.

Most of these complaints are very specific to the scenario.  The vet rules might be adding more value in other scenarios.  Most of these complaints are also disputable.  ISTM this is kind of a case study in why this particular rule about unkeyed literals would never be put in the go compiler core, though.  It does a lot of things that I, at least, find very controversial.

I'll still merge this, on the assumption that being able to run `go vet` on this repo will provide other value over time.  It's just with misgivings and confusion.

mvdan:
We've already discussed this, so I'll be brief for the sake of context. The whole point of this is being able to add struct fields without breaking downstream users. If you're convinced that you won't ever need more than just one field, then you can either live with the verbosity, or not use a struct at all. Not using a struct also gives you the "changing the type breaks downstream users" semantics.

Also, the rule is "struct type declared in a different package", since the idea is that the package boundary is what defines "downstream users" where compatibility matters.

And indeed, the whole point here is being able to use vet. You can fight its rules if you so wish, but that would have to happen in the upstream Go tracker.

PR #123, opened 2020-12-13
clean up node/gendemo regeneration

warpfork:
Code generation now has standard triggers (`go generate` will do the right thing) based on what @mvdan did in [his hamt work](https://github.com/ipld/go-ipld-adl-hamt/blob/25abfe77c0bc35c0f7ccff285a599afd00f0300e/doc.go#L1-L2) (...instead of my previous weird hack using test packages, which was terrible).


warpfork:
Not sure if want to merge this or not yet: regenerating results in a diff that undoes some of https://github.com/ipld/go-ipld-prime/pull/119 ; apparently we'd have to carry some of that stuff back up into the codegenerator.

node/gendemo/test_test.go
```diff
@@ -1,38 +1,11 @@
 package gendemo
```
mvdan, line 1: nit: call this gendemo_test.go, as test_test.go is a bit silly

node/gendemo/gen_trigger.go
```diff
@@ -0,0 +1,4 @@
+//go:generate go run gen.go
+//go:generate gofmt -w .
+
+package gendemo
```
mvdan, line 4: tip: you can use doc.go for this, which also nudges you towards writing a package godoc, which you should do anyway :) I like having doc.go files with just the package godoc and go:generate lines that don't belong elsewhere.

mvdan (approved):
I did not realise I had changed some of the generated files. A lot of the changes were automated, so I didn't really pay close attention to whether the files were generated.

Sounds good to me anyway. When I inevitably fix the generator, all the generated code will be fixed once again. This is a bit of diff churn, but it's small, and it's a good step anyway.

node/gendemo/gen_trigger.go
warpfork: Is this a common gopher convention?  I'd kind of like it if there was a convention pattern in filename that I can visually scan for to guess if some kind of generation is in order or not.  x and x_trigger would do it for me.  But I'm probably inventing that "convention", which doesn't make it much of a convention either.

I was copying what you did in another repo here with the exact invocations, but I have to admit, when I found the generate commands in doc.go in that other repo, I made a bit of a "Hæ?" sound.

mvdan: well, the general convention is "put the generate line next to the code using the generated files". If that doesn't apply, then you find some package-generic file to put it in, like main.go, pkgname.go, or doc.go. I like that last one because I find it easier to know that all the package-scoped special comments will be there.

I do think that a file like gen.go is a bit unnecessary. If you don't have a doc.go because the godoc is in a pkgname.go, then put the generate lines there too. If you do have a doc.go, then two tiny Go files with just comments feels a bit silly. And if you are in neither case because you don't have a godoc... add one :) 

> I was copying what you did in another repo here with the exact invocations, but I have to admit, when I found the generate commands in doc.go in that other repo, I made a bit of a "Hæ?" sound.

Where did I use a trigger file?

mvdan: I guess a way to summarize my thoughts is - if you can avoid an extra tiny file... avoid it.

warpfork: Alright, fine, I'll roll with it.

I suppose there's also some utility to having a single file per package, because any steps like gofmt shouldn't need to be repeated even if one were to have more than one generator at work that requires other files.

PR #125, opened 2020-12-15
all: rewrite interfaces and APIs to support int64

### LINKED ISSUE #124: Refactor all packages to support 64-bit integers properly
Surfaced here https://github.com/ipld/go-codec-dagpb/pull/3 thanks to @marten-seemann including a `GOARCH=386` variant of `go test` on the new DAG-PB codec.

I have a test case in my cross-language compat suite that's testing `Tsize` at 2^53 to push at the JavaScript maximum safe integer boundary. go-ipld-prime is decoding that and `AsInt()` is returning an `int` which truncates down to `-1`.

https://github.com/ipld/go-codec-dagpb/blob/41b03bfa1f20105c0b852dc81e555f5fd16bce12/compat_test.go#L183-L191

Is it feasable, or desirable to convert all of `AsInt()` to an explicit `int64` so we avoid this silly Go arch variability with `int`?

mvdan:
We only supported representing Int nodes as Go's "int" builtin type.
This is fine on 64-bit, but on 32-bit, it limited those node values to
just 32 bits. This is a problem in practice, because it's reasonable to
want more than 32 bits for integers.

Moreover, this meant that IPLD would change behavior if built for a
32-bit platform; it would not be able to decode large integers, for
example, when in fact that was just a software limitation that 64-bit
builds did not have.

To fix this problem, consistently use int64 for AsInt and AssignInt.

A lot more functions are part of this rewrite as well; mainly, those
revolving around collections and iterating. Some might never need more
than 32 bits in practice, but consistency and portability is preferred.
Moreover, many are interfaces, and we want IPLD interfaces to be
flexible, which will be important for ADLs.

Below are some GNU sed lines which can be used to quickly update
function signatures to use int64:

	sed -ri 's/(func.* AsInt.*)\<int\>/\1int64/g' **/*.go
	sed -ri 's/(func.* AssignInt.*)\<int\>/\1int64/g' **/*.go
	sed -ri 's/(func.* Length.*)\<int\>/\1int64/g' **/*.go
	sed -ri 's/(func.* LookupByIndex.*)\<int\>/\1int64/g' **/*.go
	sed -ri 's/(func.* Next.*)\<int\>/\1int64/g' **/*.go
	sed -ri 's/(func.* ValuePrototype.*)\<int\>/\1int64/g' **/*.go

Note that the function bodies, as well as the code that calls said
functions, may need to be manually updated with the integer type change.
That cannot be automated, because it's possible that an automated fix
would silently introduce potential overflows not being handled.

Some TODOs and FIXMEs for overflow checks are removed, since we remove
some now unnecessary int64->int conversions. On the other hand, the
older codecs based on refmt need to gain some overflow check TODOs,
since refmt uses ints. That is okay for now, since we'll phase out refmt
pretty soon.

While at it, update codectools to use int64 for token Length fields, so
that it properly supports full IPLD integers without machine-dependent
behavior and overflow checks. The budget integer is also updated to be
int64, since the lengths it uses are now int64.

Note that this refactor needed changes to the Go code generator as well
as some of the tests, for the purpose of updating all the code.

Finally, note that the code-generated iterator structs do not use int64
fields internally, even though they must return int64 numbers to
implement the interface. This is because they use the numeric fields to
count up to a small finite amount (such as the number of fields in a Go
struct), or up to the length of a map/slice. Neither of them can ever
outgrow "int".

Fixes #124.


mvdan:
@warpfork here, I assumed that `go generate ./...` covers all generated code. If it does not, please let me know.

mvdan:
... was schema/schema2/kind.go a symlink? I appear to have undone that, but I didn't intend to. A bit weird to use symlinks for Go code, to be honest, because they are not guaranteed to work with the Go tool.

schema/gen/go/_test/.gitignore
```diff
@@ -1 +0,0 @@
-/*
```
warpfork, line 1: Removing this was probably an accident?

schema/gen/go/genList.go
```diff
@@ -186,11 +186,11 @@ func (g listGenerator) EmitNodeMethodListIterator(w io.Writer) {
 			idx  int
 		}
 
-		func (itr *_{{ .Type | TypeSymbol }}__ListItr) Next() (idx int, v ipld.Node, _ error) {
+		func (itr *_{{ .Type | TypeSymbol }}__ListItr) Next() (idx int64, v ipld.Node, _ error) {
 			if itr.idx >= len(itr.n.x) {
 				return -1, nil, ipld.ErrIteratorOverread{}
 			}
-			idx = itr.idx
+			idx = int64(itr.idx)
```
warpfork, line 193: Yeah, these upcasts being needed in iterators all the time are a bit weird.  You explained it well already in the PR description though, and the logic seems sound.  :+1:

warpfork:
- I think `go generate ./...` does indeed now refresh all generated code \o/ :+1: 
- Yes, heh, `schema2/kind.go` was a symlink.  It's fine to flatten it.  I do things like that sometimes in draft packages while they're still in development and not really meant to be used.  It would get flattened when that package comes out of draft anyway.

warpfork (approved):
Looks entirely good to me!  Modulo that one gitignore file that looks like it may have gotten munched by accident.

mvdan:
Ah, yes, fat fingers with `rm` most likely. I'll undo that.

mvdan:
Undone, merging on green.

PR #126, opened 2020-12-16
all: rename AssignNode to ConvertFrom

### LINKED ISSUE #95: rename AssignNode to ConvertFrom
This should be more intuitive to Go programmers, since assignments are generally trivial operations, but conversions imply that extra work might be needed to adapt the value to fit in the recipient.

Should break a lot of users, but has an easy search-and-replace fix.

mvdan:
This should be more intuitive to Go programmers, since assignments are
generally trivial operations, but conversions imply that extra work
might be needed to adapt the value to fit in the recipient.

The entire change is just:

	sed -ri 's/AssignNode/ConvertFrom/g' **/*.go

Downstream users can very likely use the same line to fix their function
declarations and calls.

Fixes #95.

### DIFF
--- a/adl/rot13adl/rot13node.go
+++ b/adl/rot13adl/rot13node.go
@@ -175,7 +175,7 @@ func (_R13String__Assembler) AssignBytes([]byte) error {
 func (_R13String__Assembler) AssignLink(ipld.Link) error {
 	return mixins.StringAssembler{TypeName: "rot13adl.R13String"}.AssignLink(nil)
 }
-func (na *_R13String__Assembler) AssignNode(v ipld.Node) error {
+func (na *_R13String__Assembler) ConvertFrom(v ipld.Node) error {
 	if v.IsNull() {
 		return na.AssignNull()
 	}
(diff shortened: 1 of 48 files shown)


warpfork:
I think I'm going to have to follow this up with cleaning up some comments (and possibly edge-case-filled code) in the basicnode package, where at some point I had speculative ideas about carrying around other nodes without converting them.  But I'm find doing that in a separate small PR right after this one.

warpfork (approved):
:ship:

warpfork:
We might be reverting this again before tagging v0.7.

The original motive for changing this method name is to express that it _might_ involve a restructuring of memory that has non-zero-costs.

The trouble is... this method also _often_ does have very minimal costs, and we want to communicate that it's in fact probably "the fastest thing possible" (it's just that what that is can be contextually dependent).  It often _doesn't_ do any work that would be well-described as "converting" if it's used on two nodes with the same implementations (and that's a case we expect to predominate in typical usage).

This can be seen exemplified in another recent commit: https://github.com/ipld/go-ipld-prime/pull/130/files#diff-ca48982ff44e5ebb0922e408b4af41e521aa1aa048d38902c053a9f64724a56eR239

To say it another way: the contract of this method should be roughly: "carry this information in the smartest way you know how.  If that means restructuring it to your memory layout: fine.  If you already have the same memory layout: you should be doing nothing.  If you don't know if your memory layout is better, and you want to keep whatever you're given, carrying around as a pointer to the original: also fine."  (AssignNode/ConvertFrom on a value assembler while building a basicnode.Map will do this, for example: whatever node implementation type it's given, that's what it's going to store.)
 
Some synonyms to consider might include: TransformNode? TranslateNode? CastNode?  But we haven't selected any of these with confidence, so for now, our action is to revert this rename and thus have one fewer change for people to deal with when moving v0.6->v0.7, and we may reconsider this name again in the future.

PR #127, opened 2020-12-18
all: rename schema.Kind to TypeKind, ipld.ReprKind to Kind

### LINKED ISSUE #94: rename ReprKind to just Kind
This is everywhere, so the shorter name will make writing code significantly nicer. And an ipld-prime kind is still not an ambiguous concept.

This would break practically all users, but the fix is easy via search-and-replace.

/cc @warpfork

mvdan:
As discussed on the issue thread, ipld.Kind and schema.TypeKind are more
intuitive, closer to the spec wording, and just generally better in the
long run.

The changes are almost entirely automated via the commands below. Very
minor changes were needed in some of the generators, and then gofmt.

	sed -ri 's/\<Kind\(\)/TypeKind()/g' **/*.go
	git checkout fluent # since it uses reflect.Value.Kind

	sed -ri 's/\<Kind_/TypeKind_/g' **/*.go
	sed -i 's/\<Kind\>/TypeKind/g' **/*.go
	sed -i 's/ReprKind/Kind/g' **/*.go

Fixes #94.


errors.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 // ErrWrongKind may be returned from functions on the Node interface when
-// a method is invoked which doesn't make sense for the Kind and/or ReprKind
+// a method is invoked which doesn't make sense for the Kind and/or Kind
```
warpfork, line 9: I think we can collapse this bit that became "Kind and/or Kind" to just "Kind" now :laughing: 

(I must've written this quite a while ago, when the relationship between the two was less clear.  Now it seems quite clear that only referring to data model Kind here is sufficient.)

warpfork (changes requested):
I'm so happy about this.  This rename is long due, and the new code reads much better all over the place.  Thank you for grinding this through!

I found a couple spots where things got renamed to TypeKind where they should still be just Kind in the new nomenclature.  Everything is _logically_ correct of course, but the names are off.  I hope it's not too much work to flip those back.  And after that I'll be really excited to merge this! :)

mvdan:
Thanks for the suggestions! All applied, I think.

warpfork:
Big hoorays! :confetti_ball:

PR #130, opened 2020-12-27
Implement traversal.FocusedTransform.

warpfork:
`traversal.FocusedTransform` is now available.  It takes a starting Node, a Path, and a TransformFn (which is essentially just a function from Node to Node), and uses those instructions to build a new Node tree with the transform applied at the given path.

Like all the other `traversal.*` functions, it carries a `traversal.Progress` object around, which you can use in the TransformFn to see where you are, or to do continued traversals or transformations at deeper levels of the data while keeping a coherent track of overall position, etc.

This works even if the Path happens to cross links!  If you've configured enough Loader and Storer functions in the `traversal.Config`, any links crossed will be automatically loaded, the transform will continue inside, and when the transform is complete, the data beyond the link will be reserialized in the same style as the original link (e.g. in the case of CIDs, that means: same multicodec will be used, same multihash will be used, etc etc), and the link will be replaced with a link to the new data, which will also have been stored.

This patch also includes a few new accessors for Path that are helpful and reasonable.

### DIFF
--- a/path.go
+++ b/path.go
@@ -147,11 +147,19 @@ func (p Path) String() string {
 	return sb.String()
 }
 
-// Segements returns a slice of the path segment strings.
+// Segments returns a slice of the path segment strings.
 //
 // It is not lawful to mutate nor append the returned slice.
 func (p Path) Segments() []PathSegment {
 	return p.segments
+}
+
+// Len returns the number of segments in this path.
+//
+// Zero segments means the path refers to "the current node".
+// One segment means it refers to a child of the current node; etc.
+func (p Path) Len() int {
+	return len(p.segments)
 }
 
 // Join creates a new path composed of the concatenation of this and the given path's segments.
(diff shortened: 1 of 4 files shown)


warpfork:
I want to briefly confess that I also tried to work on a bigger "munge" function which would be able to do several of these focused transforms (and some kinds of bigger, non-point mutations) in aggregate.  Ideally, such a function could be much more efficient than doing the same thing with a series of FocusedTransform.

That got... _interesting_.  In order to make such a function as efficient as desired, it's necessary to reorder operations to the optimal order of application.  But not all operations one can imagine wanting (such as, say, list insertions at non-end positions) are commutative.  (And then, worse: deciding whether an operation is commutative would require looking at the data, not just the target path, since path segments for lists vs path segments for maps are not distinguishable.)

Long story short, a multi-transform feature will need design work.

PR #133, opened 2021-01-03
gengo: support for unions with stringprefix representation.

warpfork:
Title says it all, actually.

### DIFF
--- a/node/mixins/delim.go
+++ b/node/mixins/delim.go
@@ -31,3 +31,10 @@ func SplitExact(s string, sep string, count int) ([]string, error) {
 	}
 	return ss, nil
 }
+
+// SplitN is an alias of strings.SplitN, which is only present here to
+// make it usable in codegen packages without requiring conditional imports
+// in the generation process.
+func SplitN(s, sep string, n int) []string {
+	return strings.SplitN(s, sep, n)
+}
--- a/schema/gen/go/genUnionReprStringprefix.go
+++ b/schema/gen/go/genUnionReprStringprefix.go
@@ -12,14 +12,14 @@ var _ TypeGenerator = &unionReprStringprefixGenerator{}
 func NewUnionReprStringprefixGenerator(pkgName string, typ *schema.TypeUnion, adjCfg *AdjunctCfg) TypeGenerator {
 	return unionReprStringprefixGenerator{
 		unionGenerator{
-			adjCfg,
-			mixins.MapTraits{
+			AdjCfg: adjCfg,
+			MapTraits: mixins.MapTraits{
 				PkgName:    pkgName,
 				TypeName:   string(typ.Name()),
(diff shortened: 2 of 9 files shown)


willscott (approved):

node/mixins/delim.go
```diff
@@ -31,3 +31,10 @@ func SplitExact(s string, sep string, count int) ([]string, error) {
 	}
 	return ss, nil
 }
+
+// SplitN is an alias of strings.SplitN, which is only present here to
+// make it usable in codegen packages without requiring conditional imports
+// in the generation process.
+func SplitN(s, sep string, n int) []string {
```
mvdan, line 38: hmm, just run goimports on the generated code? "figure out what imports we need" is a solved problem in go :)

schema/gen/go/genUnionReprStringprefix.go
```diff
@@ -0,0 +1,258 @@
...
+	return unionReprStringprefixGenerator{
+		unionGenerator{
+			adjCfg,
+			mixins.MapTraits{
+				PkgName:    pkgName,
+				TypeName:   string(typ.Name()),
+				TypeSymbol: adjCfg.TypeSymbol(typ),
+			},
+			pkgName,
+			typ,
```
mvdan, line 22: use keyed fields? otherwise it's hard to tell what's going on here. I personally always use keys when the number of fields is more than 2

mvdan (approved):
I have to admit it's hard to actually understand the template code, but it looks fine to me.

schema/gen/go/genUnionReprStringprefix.go
warpfork: Fine, done.

node/mixins/delim.go
warpfork: That seems _significantly_ more complex.  I'm not okay with pushing a need for goimports -- something from the `x/tools` repos -- onto users in order to have *compilable* code come out.  (By contrast, while elsewhere we decided we're okay leaning on post-gen invocation of gofmt, that differs on two fronts: it's effectively always available, and it's also not strictly necessary to apply before compile will succeed.)

A small handful of cherry-picked constants like this seems fine to me.

We may want to consider moving such things to their own small package (or the minima file in the output?) to make them less obtrusive, though.

warpfork: I guess goimports is probably also possible to use as a library, so we could just add it to our own dependencies.  But I'd still say for this small of a need, it's probably not worth it.  I'd favor KISS on this until we really have a desire for conditional imports.

mvdan: I was under the assumption we'd be the only ones running this generator. Why would other people generate their own mixins directly?

warpfork: Maybe the comment is unclear -- SplitN is used in the parsing logic of the code that comes *out*.

PR #135, opened 2021-01-06
schema/gen/go: please vet a bit more

mvdan:
In particular, this removes ~50 out of the 2.7k warnings in 'go vet
./...' in this repository. Mainly, the "unreachable code" ones.

This was caused by edge cases in some of the generated code which caused
an unconditional return or panic statement to be followed by other code.
Fix all of them with a bit more template logic.

Some of the Next methods go a bit further. If they serve no purpose as
the switch has no cases to be matched, just unconditionally return an
error. In the future we can perhaps reuse a single function for that.

Finally, I was having a hard time actually following the logic in
kindedUnionNodeAssemblerMethodTemplateMunge, so I've indented the code a
bit to follow the template logic and scoping.

These changes move us towards pleasing vet, which is nice, but also make
the code waste a bit less space.

### DIFF
--- a/schema/gen/go/genUnion.go
+++ b/schema/gen/go/genUnion.go
@@ -457,6 +457,7 @@ func (g unionBuilderGenerator) emitMapAssemblerMethods(w io.Writer) {
 			if ma.ca != 0 {
 				return nil, schema.ErrNotUnionStructure{TypeName:"{{ .PkgName }}.{{ .Type.Name }}", Detail: "cannot add another entry -- a union can only contain one thing!"}
 			}
+			{{- if .Type.Members }}
 			switch k {
 			{{- range $i, $member := .Type.Members }}
 			case "{{ $member.Name }}":
@@ -478,9 +479,9 @@ func (g unionBuilderGenerator) emitMapAssemblerMethods(w io.Writer) {
 				return ma.ca{{ add $i 1 }}, nil
 				{{- end}}
 			{{- end}}
-			default:
-				return nil, ipld.ErrInvalidKey{TypeName:"{{ .PkgName }}.{{ .Type.Name }}", Key:&_String{k}}
+			{{- end}}
 			}
+			return nil, ipld.ErrInvalidKey{TypeName:"{{ .PkgName }}.{{ .Type.Name }}", Key:&_String{k}}
 		}
 	`, w, g.AdjCfg, g)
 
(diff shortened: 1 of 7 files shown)


schema/gen/go/genUnion.go
```diff
@@ -478,9 +479,9 @@ func (g unionBuilderGenerator) emitMapAssemblerMethods(w io.Writer) {
 				return ma.ca{{ add $i 1 }}, nil
 				{{- end}}
 			{{- end}}
-			default:
-				return nil, ipld.ErrInvalidKey{TypeName:"{{ .PkgName }}.{{ .Type.Name }}", Key:&_String{k}}
+			{{- end}}
 			}
+			return nil, ipld.ErrInvalidKey{TypeName:"{{ .PkgName }}.{{ .Type.Name }}", Key:&_String{k}}
```
warpfork, line 484: Does `go vet` prefer this?

I tend to like writing a default clause with a return statement in it, because then it makes sure I don't forget a return statement in every other branch in every other part of the switch (if there's no return outside the switch, then the compiler would yell at me if I forget one in any of the cases).

mvdan: I thought you might flag this :P The vet rule at play here was just the dead code one, because in the zero-case scenario, we would produce something like:

```
switch {
default:
    return someError
}
return nil // dead!
```

So the simplest fix, from the POV of the generator, was to flip the returns around so that this could never happen. But I can just add some extra template logic to instead emit different code when there are zero cases. Would you prefer that?

Regarding the safety of the code I agree, but I think it matters less here since it's generated code. But I'm happy to undo a change that's not really necessary to appease vet.

mvdan:
I just realise that the title of this commit can be read as "please, use vet more often" :P

schema/gen/go/genUnion.go
warpfork: Except we didn't generate that trailing `return nil`, either, so... the vet dead code detector on the zero-case scenario just doesn't like a lonely default?  (I guess that's not unreasonable, I'm just trying to make sure my understanding is correct.)

I guess this is fine, I'm just almost a little bummed it needs a condition in the template at all rather than just letting range over an empty list quietly DTRT.  Any time zero isn't a special case I'm happier.

warpfork (approved):

schema/gen/go/genUnion.go
mvdan: It did generate dead code, I'm 100% sure. I'm just not going to bother spending half an hour digging up what the original code looked like :P If vet complained about "no-op" switches as dead code, there would be a lot of angry people about it.

You seem to prefer simpler generation templates, as do I, so I'll merge this.

PR #136, opened 2021-01-10
use %q in error strings

mvdan:
%q is superior to the manually quoted "%s", since it will properly
escape the inner string when quoting. For example:

	"here goes a double quote: \" "

Even if this does not matter in practice, using %q is easier too.


warpfork (approved):
Absolutely :raised_hands: 

Jeesh.  What complete goober wrote those?  (Me.  It was me.  Yikes.)

Issue #102, opened 2020-10-26
schema/gen/go: produces code that makes vet unhappy


mvdan:
I see a mix of `composite literal uses unkeyed fields` and `unreachable code`. Both are pretty valid concerns, and would make the generated code easier to read. This is low priority, but also low hanging fruit that would make `go vet ./...` actually useful instead of spitting out dozens of lines.

mvdan:
We're up to thousands of warnings now, since merging the dmt code. But at least the vast majority can be fixed by "just" fixing the code generator.

I want to get this under control and locked to 0 vet warnings before it slips past me again.

mvdan:
This is mostly fixed now. A few edge cases still produce vet warnings, but we can tackle those few remaining ones over time. The important part is that repos like go-codec-dagpb can use vet now, so I can close this.

PR #137, opened 2021-01-10
schema/gen/go: prevent some unkeyed literal vet errors

mvdan:
In particular, use keys for ipld error structs. These have one field, so
the changes are pretty simple.

Reduces 'go vet ./...' from 2647 lines of output to 2365.

Updates #102.


warpfork (approved):
I'd just mark that the change of

```
- ipld.ErrNotExists{ipld.PathSegmentOfString(key)}
+ ipld.ErrNotExists{Segment: ipld.PathSegmentOfString(key)}
```

is very much one of those where I think `go vet` is increasing verbosity while carrying absolutely no semiotic value.

But :shrug: :+1:

PR #139, opened 2021-01-18
fluent/qp: finish writing all data model helpers

mvdan:
I started rewriting the "getting started" Go guide to use fluent/qp
instead of fluent, but I'm missing some of the helpers since the guide
uses links, among others.

This is largely copy-pasted, but there are just a handful of types and
it's barely a dozen lines per type. A generator is not worth it for 100
lines of code that will rarely ever need to change.

### DIFF
--- a/fluent/qp/qp.go
+++ b/fluent/qp/qp.go
@@ -81,16 +81,28 @@ func ListEntry(la ipld.ListAssembler, fn Assemble) {
 	fn(la.AssembleValue())
 }
 
-type stringParam string
+type nullParam struct{}
 
-func (s stringParam) Assemble(na ipld.NodeAssembler) {
-	if err := na.AssignString(string(s)); err != nil {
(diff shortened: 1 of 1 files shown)


warpfork:
> A generator is not worth it for 100 lines of code that will rarely ever need to change.

There have been surprisingly many times throughout the course of this project now where I've thought this.  I'm starting to wonder if a (very) small, (very) KISS little shop-made-jig generator library might actually be a positive sum investment.

But... probably not.  Just as likely that we'd then end up with a dozen "shop made jigs", all different sizes, and that might not be a happy outcome either.

Anyway.  Just taking a moment to think outloud, because that very same thought had passed through my head yet again as recently as sunday.

warpfork (approved):

mvdan:
My personal stance is that this took me maybe twenty minutes to copy-paste and adjust, so it's not worth the time automating. If I have to re-do that work manually two or three more times, then that's probably a sign it needs generation :)

PR #140, opened 2021-01-18
schema/gen/go: cache genned code in os.TempDir

mvdan:
This means we no longer clutter the repository with lots of files, even
if they are git-ignored. It's always a bit of a red flag when you run
"go test ./..." and the result is a bunch of leftover files.

We still want to keep the files around, for the sake of Go's build
cache. And we still want their paths to be static between "go test"
runs. So put them in a static dir under os.TempDir.

This does mean that concurrent runs of these tests will likely not work
well. I don't imagine that's going to be a problem anytime soon, though.
If it really becomes a problem in the future, we could figure something
out like grabbing a file lock for the directory.

The idea behind using os.TempDir is that it will likely remain in place
between a number of "go test" runs within a hacking session, but it will
be eventually cleaned up by the system, such as when rebooting.

Note that we need to use globbing since one can't build "proper
packages" located outside a module. The only exception is building an
ad-hoc set of explicit Go files. While at it, use filepath.Join, to be
nice.

### DIFF
--- a/schema/gen/go/testEngine_test.go
+++ b/schema/gen/go/testEngine_test.go
@@ -3,6 +3,7 @@ package gengo
 import (
 	"io"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/ipld/go-ipld-prime"
@@ -21,6 +22,8 @@ import (
 // their representation prototypes can be obtained by appending ".Repr".
 type behavioralTests func(t *testing.T, getPrototypeByName func(string) ipld.NodePrototype)
 
+var tmpGenBuildDir = filepath.Join(os.TempDir(), "go-ipld-prime-gengo")
+
 func genAndCompileAndTest(
 	t *testing.T,
 	prefix string,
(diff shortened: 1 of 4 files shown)


schema/gen/go/testEngine_test.go
```diff
@@ -30,23 +33,25 @@ func genAndCompileAndTest(
...
-	//  We don't by default because it's nicer to let go's builds of things cache.
-	//  If you change the names of types, though, you'll have garbage files leftover,
-	//   and that's currently a manual cleanup problem.  Sorry.
-	os.Mkdir("./_test/", 0755)
-	os.Mkdir("./_test/"+prefix, 0755)
+	// They will live in a temporary directory, usually
+	// /tmp/go-ipld-prime-gengo on Linux. It can be removed at any time.
+	// We don't by default because it's nicer to let go's builds of things cache.
+	// If you change the names of types, though, you'll have garbage files leftover,
+	// and that's currently a manual cleanup problem.  Sorry.
```
warpfork, line 40: Oh, heh.  Not exactly related to your change, but since it shows up in the diff: we could remove these two lines now.  We no longer have this problem :D

warpfork (approved):

PR #142, opened 2021-01-25
add an ADL interface type

mvdan:
As a method superset of Node, with the Substrate method added on top.
This way, we can generally say that any ADL implementation should
include this method to allow for easy encoding over the wire.

Reify, the opposite of Substrate, is notably missing here. This is
because it's generally a top-level function, not a method. We might want
to rethink this disparity in the future.

### DIFF
--- a/node.go
+++ b/node.go
@@ -162,6 +162,19 @@ type Node interface {
 	Prototype() NodePrototype
 }
 
+// ADL represents an Advanced Data Layout, a special kind of Node which
+// implements custom logic while still behaving like an IPLD node.
+//
+// For more details, see the docs at
+// https://github.com/ipld/specs/blob/master/schemas/authoring-guide.md.
+type ADL interface {
+	Node
+
+	// Substrate returns the underlying Data Model node, which can be used
+	// to encode an ADL's raw layout.
+	Substrate() Node
+}
+
 // NodePrototype describes a node implementation (all Node have a NodePrototype),
 // and a NodePrototype can always be used to get a NodeBuilder.
 //


mvdan:
@warpfork PTAL, also the part about Reify.

I also wonder if we should have some way to mark which nodes are plain old data model. Because it would be a lot clearer to do `Substrate() ipld.DataModelNode`, for example. Similarly, `Reify(ipld.DataModelNode) ipld.ADL` :)

warpfork (approved):
This seems good to me.

For a while, I had been wondering if there are any more explicit features that might demarcate an ADL -- for example, some kind of standardize access to a "config" concept, or similar? -- but that doesn't seem to have emerged.

Getting a view to the substrate data -- *that* has seemed to be a feature that's present 100% of the time.  So defining the interface around this seems correct.  :+1:

mvdan:
> for example, some kind of standardize access to a "config" concept, or similar? -- but that doesn't seem to have emerged.

At least, that pattern has emerged in the form of Prototype, I think. But that doesn't really fit an interface.

willscott (approved):

warpfork:
Yeah, I don't know if there's anything that can be done about standardizing Reify.  It's a constructor function, more or less; those do tend to end up being fairly freestanding.

We could make it a convention to put that on the prototype, and also make an `ADLPrototype` interface that defines a standard shape for that feature.  Might be reasonable.  Might also be a bit odd, since prototypes generally carry the config, and in some cases, we'll expect the config to come (in part?  conditionally?  this is unclear and I suspect may vary by ADL and per implementation) from the substrate data itself when reifying.  Perhaps we should try it and see?  But maybe not propose it as an interface until we try it out for a while in an implementation and see if it feels right.

>  At least, that pattern has emerged in the form of Prototype, I think. But that doesn't really fit an interface.

Right, and also we already have access to it in a standard way stated on the Node interface.

PR #146, opened 2021-03-03
Demonstration of removing cidlink.Link by removing prototype method.

hannahhoward:
# Goals

Make cid.Cid satisfy ipld.Link by moving Prototype method to LinkSystem

### DIFF
--- a/link.go
+++ b/link.go
@@ -30,9 +30,6 @@ import (
 // This is because Link is often desirable to be able to use as a golang map key,
 // and in that context, pointers would not result in the desired behavior.
 type Link interface {
-	// Prototype should return a LinkPrototype which carries the information
-	// to make more Link values similar to this one (but with different hashes).
-	Prototype() LinkPrototype
 
 	// String should return a reasonably human-readable debug-friendly representation the Link.
 	// There is no contract that requires that the string be able to be parsed back into a Link value,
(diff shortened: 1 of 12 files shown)


link.go
```diff
@@ -30,9 +30,6 @@ import (
 // This is because Link is often desirable to be able to use as a golang map key,
 // and in that context, pointers would not result in the desired behavior.
 type Link interface {
-	// Prototype should return a LinkPrototype which carries the information
-	// to make more Link values similar to this one (but with different hashes).
-	Prototype() LinkPrototype
```
warpfork, line 35: The main thing I'd be uncomfortable with here is now this interface says effectively nothing.

I could accidentally use anything that has a `String() string` method as a Link.

There's already a very uncomfortable practical gap in what these interfaces can actually enforce vs what has to be feature-detected and casted in the functions that compose a LinkSystem, but it's still nice imo to have an interface that has enough entropy in it to defend you from most typos.  Without this function, it wouldn't defend from many typos at all.

willscott: i guess you either end up with a prototype() of nil when the system doesn't understand the link as it's own, or you could have a `(LinkPrototype, error)` response?

warpfork:
I think I will close this.  It's agreed that something is uncomfortable about the indirections here, but this is probably not the way forward to fix it.

PR #147, opened 2021-03-03
remove cidlink.Link and cidlink.LinkPrototype entirely.

hannahhoward:
# Goals

Just an expansion of last PR that also removes BuildLink from LinkPrototype.


warpfork:
As with the parent, https://github.com/ipld/go-ipld-prime/pull/146, I think I will close this.  It's agreed that something is uncomfortable about the indirections here, but this is probably not the way forward to fix it.

PR #150, opened 2021-03-05
codec/raw: implement the raw codec

mvdan:
(see commit message)

### DIFF
--- a/codec/raw/codec.go
+++ b/codec/raw/codec.go
@@ -1,0 +1,63 @@
+// Package raw implements IPLD's raw codec, which simply writes and reads a Node
+// which can be represented as bytes.
+//
+// The codec can be used with any node which supports AsBytes and AssignBytes.
+// In general, it only makes sense to use this codec on a plain "bytes" node
+// such as github.com/ipld/go-ipld-prime/node/basic.Prototype.Bytes.
+package raw
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	ipld "github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+)
+
+// TODO(mvdan): make go-ipld-prime use go-multicodec soon
+const rawMulticodec = 0x55
+
+func init() {
+	cidlink.RegisterMulticodecDecoder(rawMulticodec, Decode)
+	cidlink.RegisterMulticodecEncoder(rawMulticodec, Encode)
+}
+
+// Decode implements decoding of a node with the raw codec.
+//
+// Note that if r has a Bytes method, such as is the case with *bytes.Buffer, we
+// will use those bytes directly to save having to allocate and copy them. The
+// Node interface is defined as immutable, so it is assumed that its bytes won't
+// be modified in-place. Similarly, we assume that the incoming buffer's bytes
+// won't get modified in-place later.
+//
+// To disable the shortcut above, hide the Bytes method by wrapping the buffer
+// with an io.Reader:
+//
+//     Decode([...], struct{io.Reader}{buf})
+func Decode(am ipld.NodeAssembler, r io.Reader) error {
+	var data []byte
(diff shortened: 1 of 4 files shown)


mvdan:
I also could not find a spec doc for it, otherwise I would have linked it.

codec/raw/codec.go
```diff
@@ -0,0 +1,63 @@
...
+
+// TODO(mvdan): make go-ipld-prime use go-multicodec soon
+const rawMulticodec = 0x55
+
+func init() {
+	cidlink.RegisterMulticodecDecoder(rawMulticodec, Decoder)
+	cidlink.RegisterMulticodecEncoder(rawMulticodec, Encoder)
+}
+
+type byteAccesor interface {
```
warpfork, line 31: Okay, I know this is copying a typo elsewhere, but `s/Accesor/Accessor/g` please

codec/raw/codec.go
```diff
@@ -0,0 +1,63 @@
...
+	Bytes() []byte
+}
+
+func Decoder(am ipld.NodeAssembler, r io.Reader) error {
```
warpfork, line 35: Would prefer to call this `Decode` (minus the 'r') now.  (I've changed most of the other functions to this grammar in #143.)

codec/raw/codec.go
```diff
@@ -0,0 +1,63 @@
...
+	// We need r's bytes, and r is often just a buffer.
+	// So if we can get them directly without ReadAll's copying,
+	// just do that and also save an allocation.
+	// Nodes are immutable, so we won't modify the bytes.
+	// We assume that the incoming buffer won't get modified either.
```
warpfork, line 40: I think we should move this to the actual docs block on the function.  (I'm probably guilty of putting too many tidbits like this in comments too, but I should also quit it.)

I'm trying to get in the habit of using phrasing about "it is not lawful to xyz" ([example](https://github.com/ipld/go-ipld-prime/blob/7d918468f40d06d8a64a2fb6c162134da497bebb/path.go#L152)) for things like this mutability contract that we can't really enforce with the compiler.  Just to have _some_ convention for talking about this semi-formally.

warpfork (approved):
I had some nits to pick, but this is holistically pretty good.

codec/raw/codec.go
mvdan: actually, there's no need for this to be a named type. I'll just use an anonymous interface.

codec/raw/codec.go
mvdan: Ah, right, thanks for reminding me.

codec/raw/codec.go
mvdan: Fair enough. I thought about it initially, and thought that some of these details are too internal for the godoc. But I think you're right, it's still better than no godoc.

mvdan:
@warpfork ready, I think.

codec/raw/codec.go
```diff
@@ -0,0 +1,63 @@
...
+// Note that if r has a Bytes method, such as is the case with *bytes.Buffer, we
+// will use those bytes directly to save having to allocate and copy them. The
+// Node interface is defined as immutable, so it is assumed that its bytes won't
+// be modified in-place. Similarly, we assume that the incoming buffer's bytes
+// won't get modified in-place later.
+//
+// To disable the shortcut above, hide the Bytes method by wrapping the buffer
+// with an io.Reader:
+//
+//     Decode([...], struct{io.Reader}{buf})
```
warpfork, line 37: oh wow, TIL.  That's a neat idiom.

warpfork (approved):
:ship:

codec/raw/codec.go
mvdan: one can accomplish it the explicit way, like I did in the test, but this shortcut is much shorter. It's not easy to realise it's possible unless you've seen it before, so... doc :)

PR #152, opened 2021-03-22
add non-dag json codec

willscott:
Would this acceptable as a first-pass at the 'non-dag' json?

### DIFF
--- a/codec/dagjson/multicodec.go
+++ b/codec/dagjson/multicodec.go
@@ -17,13 +17,23 @@ var (
 
 func init() {
 	multicodec.RegisterEncoder(0x0129, Encode)
+	multicodec.RegisterEncoder(0x0200, Encode)
 	multicodec.RegisterDecoder(0x0129, Decode)
+	multicodec.RegisterDecoder(0x0200, decodeNonDagJSON)
 }
 
 func Decode(na ipld.NodeAssembler, r io.Reader) error {
+	return decode(na, r, true)
+}
+
+func decodeNonDagJSON(na ipld.NodeAssembler, r io.Reader) error {
+	return decode(na, r, false)
+}
+
+func decode(na ipld.NodeAssembler, r io.Reader, parseLinks bool) error {
 	// Shell out directly to generic builder path.
 	//  (There's not really any fastpaths of note for json.)
-	err := Unmarshal(na, json.NewDecoder(r))
+	err := unmarshal(na, json.NewDecoder(r), parseLinks)
 	if err != nil {
 		return err
 	}
@@ -49,7 +59,6 @@ func Decode(na ipld.NodeAssembler, r io.Reader) error {
 			return err
 		}
 	}
-	return err
 }
 
 func Encode(n ipld.Node, w io.Writer) error {
--- a/codec/dagjson/multicodec.go
+++ b/codec/dagjson/multicodec.go
@@ -56,5 +56,5 @@ func Encode(n ipld.Node, w io.Writer) error {
 	return Marshal(n, json.NewEncoder(w, json.EncodeOptions{
 		Line:   []byte{'\n'},
 		Indent: []byte{'\t'},
-	}))
+	}), true)
 }
--- a/adl/rot13adl/example_test.go
+++ b/adl/rot13adl/example_test.go
@@ -59,7 +59,7 @@ func ExampleCreatingViaADL() {
 
 	// To marshal the ADL, just use marshal methods on its substrate as normal:
 	var marshalBuffer bytes.Buffer
-	err := dagjson.Marshal(substrateNode, json.NewEncoder(&marshalBuffer, json.EncodeOptions{}))
+	err := dagjson.Marshal(substrateNode, json.NewEncoder(&marshalBuffer, json.EncodeOptions{}), true)
 	fmt.Printf("marshalled: %v\n", marshalBuffer.String())
 	fmt.Printf("marshal error: %v\n", err)
 
--- a/adl/rot13adl/example_test.go
+++ b/adl/rot13adl/example_test.go
@@ -20,7 +20,7 @@ func ExampleUnmarshallingToADL() {
 	nb := rot13adl.Prototype.SubstrateRoot.NewBuilder()
 
 	// Unmarshal -- using the substrate's nodebuilder just like you'd unmarshal with any other nodebuilder.
-	err := dagjson.Unmarshal(nb, json.NewDecoder(strings.NewReader(`"n pbby fgevat"`)))
+	err := dagjson.Unmarshal(nb, json.NewDecoder(strings.NewReader(`"n pbby fgevat"`)), true)
 	fmt.Printf("unmarshal error: %v\n", err)
 
 	// Use `Reify` to get the synthetic high-level view of the ADL data.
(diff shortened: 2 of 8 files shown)


willscott:
and follow-up question is if you have any concerns with doing the same thing (a parallel variant that skips the link logic) for dagcbor, @warpfork

codec/dagjson/multicodec.go
```diff
@@ -17,13 +17,23 @@ var (
 
 func init() {
 	multicodec.RegisterEncoder(0x0129, Encode)
+	multicodec.RegisterEncoder(0x0200, Encode)
```
warpfork, line 20: I'm inclined to think that for the json codec, if we encounter a Link in the middle of a structure, we should have the codec error explicitly.  It's not going to roundtrip via that same codec, and explicit errors are better than unstated incorrect behavior.

warpfork:
Yeah, this seems mostly fine.  In a perfect world there's a lot of DRY work I want to do between both this and the cbor codecs, but it would be silly to hold this up on that.

warpfork:
Thoughts about adding a (very small) `go-ipld-prime/codec/json` package, which does the multicodec registrations, and has its own exported `Encode` and `Decode` function... which are just thin wrappers to the workhorse functions here?

The code is largely shared, but I think having a separate package per codec is feels clean and consistent and generally reasonable.

It would also make it possible to implement the `codec/json` and `codec/cbor` packages without a transitive dependency to `linking/cid` in the future, which could be nice (even though not presently enough of a priority to be worth the larger effort required).

codec/dagjson/multicodec.go
willscott: it produces a reasonable serialization into json of a dag. making the caller traverse the dag to understand there's no links in it, rather than just flatting them isn't clear to me as a win

warpfork: I don't think callers generally have to _do_ anything.  They would just... get an error if they try to encode data that's unencodable with this format.

In general, things that are unencodable should result in errors.  That's a rule we'd follow anywhere[‡] that isn't JSON, and I think we should follow it for JSON too.

Failure to have bijections for data that we emit tends to cause unhappiness and confusion.  The problems are magnified by the fact they often sweep under the rug _for a while_ and then often reappear at points in time distant from when the data was first written.

I'd say the direction of burden-of-proof should be overwhelming weighted such that doing something looseygoosey that's non-bijective is the case that needs support.  Maintaining bijectiveness, even if it comes at the cost of more errors, should be the default design bias.

[‡] - except where floating point numbers are involved.  But that's the exception that really proves the reason for the rule, isn't it -- float consistency is a horrific minefield that we would deeply prefer didn't exist, and there are horrific amounts of time wasted by this seemingly small problem.

aschmahmann: @warpfork my understanding was that people should be able to take DAG-JSON data and by changing the codec to JSON they'd just have an object where there are no links. This is similar to how we can take an existing CID and slap a Raw codec on it and get the data in that block back as raw bytes.

This case is mostly covered by the decode case, it seems a bit weird that I wouldn't be able to roundtrip data since Encode(Decode(dagJSONDataAsJSON)) would error. Am I missing something?

willscott: when you decode as json you don't  get a link. when you re-encode the map with a string in it you'd get the same data.

willscott: maybe we're flipping around `decode` and `encode`?  decoding will realize some ipld-prime nodes. you'll end up with a map containing a string.
When you go to re-encode it to serialized json it'll work fine. it's when you encode an IPLD prime link node that it would get unhappy with you.

willscott:
@warpfork is it okay to change the public `Unmarshal` method in dagjson? (see the newer commit and offer comments)

warpfork:
I _almost_ objected to that, but caught myself in time (there's a convention that `Encode` and `Decode` methods should take no additional parameters / be the same as what you'd get via multicodec table; there's _not a convention for `Marshal` and `Unmarshal`, and in fact I think those functions even disappear in dagjson2).

warpfork (approved):
I'm good with this.

There are some specks here and there I'd clean up in a bigger refactor (e.g. separate decoder _state_ vs _config_; both are now present in one struct), but there's plenty of other things in that same heading, so I'm fine leaving these things on the wishlist until they can all be handled.

warpfork (changes requested):
Wait, hang on, where'd the conditional branch for encode end up?

willscott:
(i didn't do that yet 😁 )

codec/dagjson/multicodec.go
warpfork: As a general rule, there are _no_ guarantees about what happens when you take some data and try to wangle it with a _different_ codec.  Maybe it works out fine.  Maybe it doesn't.  It probably doesn't.  In the cases it does, that was very "lucky" and probably based mostly on coincidences.

I'm concerned with bijection *within the same codec*.

warpfork: `{"/":"bafyquux"}` is a map with one key in JSON.  When you deserialize that block with a JSON codec, you get a map with one key; when you serialize it with a JSON codec, you get the same block.

(@aschmahmann , the reason I thumbs-down'd your comment is that there's no error in this case above.  One just emits this as a map.  That's normal.)

`{"/":"bafyquux"}` is a link in DAG-JSON.  When you deserialize that block with a DAG-JSON codec, you get a link; when you serialize it with a DAG-JSON codec, you get the same block.

Now, I have a link, and I attempt serialize it with a JSON codec, I want that to yield an error, because that's not really encodable in JSON.  As a user, if I attempt to do this, _I have made a serious mistake_, and I _want to be informed immediately_.

If we didn't do this, and instead smash the link into a map with one entry during the encode process, we've changed the data.  When we deserialize it later with the *same* codec we serialized it with, we would get different data.  This would not be good.

I don't really care what happens when we serialize something with a JSON codec and attempt to later deserialize it with a DAG-JSON codec.  If that gives different data... well, yes.  A different codec was used.  That's gonna happen.

warpfork: It may help to be reminded that neither DAG-JSON nor JSON are total supersets of each other.  This sucks, and we often wish it wasn't so, but wishes... well.   Wishes.  They have limited power.

DAG-JSON can't encode maps with single entries that happen to have the key "/".  JSON can.  This makes JSON larger.

In the other direction, DAG-JSON can encode links.  JSON can't.  This makes DAG-JSON larger.

This is generally a headache.  But I'm fairly certain we make the headache _bigger_ rather than smaller if we were to add more silent coercions to the picture.

warpfork (approved):
:shipit:

codec/dagjson/multicodec.go
aschmahmann: np, just wanted to clarify what you were actually intending/trying to do. I have no problem with a codec's `Encode` function erroring when trying to convert an IPLD data model object to bytes when that codec doesn't support part of the data model used in that object (e.g. JSON doesn't support links).

aschmahmann: > DAG-JSON can't encode maps with single entries that happen to have the key "/".

@warpfork maybe I'm missing it, but it doesn't look like DAG-JSON's `Encode` function errors with these map entries. Should that be fixed in another PR?

warpfork: ... :weary: yes, you're quite correct

PR #160, opened 2021-04-06
schema/gen/go: avoid Maybe pointers for small types

mvdan:
(see commit message)

### DIFF
--- a/schema/gen/go/adjunctCfg.go
+++ b/schema/gen/go/adjunctCfg.go
@@ -2,8 +2,11 @@ package gengo
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
+	ipld "github.com/ipld/go-ipld-prime"
+	basicnode "github.com/ipld/go-ipld-prime/node/basic"
 	"github.com/ipld/go-ipld-prime/schema"
 )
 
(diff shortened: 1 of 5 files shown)


schema/gen/go/adjunctCfg.go
```diff
@@ -74,13 +74,58 @@ func (cfg *AdjunctCfg) MaybeUsesPtr(t schema.Type) bool {
...
+func pointerSizeOfSchemaType(t schema.Type) int {
+	switch t.(type) {
+	// Go basic types which aren't bigger than a pointer.
+	// We assume that Enum is represented like Int.
+	case *schema.TypeBool, *schema.TypeInt, *schema.TypeFloat, *schema.TypeEnum:
+		return 1
+
+	// Go types which are just a pointer.
+	case *schema.TypeMap:
+		return 1
```
warpfork, line 109: I think this might technically be closer to 4, because our maps internally use both a golang map and a list to preserve order.

I'm not sure if that argues for bumping the threshhold to 4 or not.

mvdan: Oh, I wasn't aware. If it's indeed Go map + Go slice, then we should change it to 4.

Are these declared as Go types anywhere? Because if so, we could at least grab the sizes as globals, which would then be kept up to date as the types change.

mvdan:
I've re-done this with reflection and basicnode, which means it's no longer an estimation, and will stay up to date as the internals change.

If there is a better place to grab Go values/types than basicnode for these measurements, I'm all ears. Perhaps some code already generated by schema/gen/go?

Anyway, there's zero behavioral change compared to before. Another `go generate ./...` run produced zero changes in both repos, so the benchmark numbers are still up to date.

warpfork:
Ahh, right.  We can't readily sizeof on the types before they're generated.  This does still seem like a reasonable heuristic though, and certainly does do the right thing right now for practical values.  :+1:

warpfork (approved):
Merge freely :)

PR #161, opened 2021-04-09
schema/gen/go: batch file writes via a bytes.Buffer

mvdan:
(see commit message)

### DIFF
--- a/schema/gen/go/generate.go
+++ b/schema/gen/go/generate.go
@@ -1,9 +1,10 @@
 package gengo
 
 import (
+	"bytes"
 	"fmt"
 	"io"
-	"os"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


mvdan:
There are better libraries that perform better at atomic file replaces, like google/renameio, but I feel like that's a bit overkill for a code generator right now - especially given the added dependency.

mvdan:
Uh, weird that the tests broke. Looking.

mvdan:
Ah, the tests run in parallel, so no buffer reuse. Should be green now.

warpfork:
Okay, phew, that was the one thing I was going to comment made me uncomfortable anyway :sweat_smile: (admittedly, unexported, so maybe could've been fine, but, still.)

PR #163, opened 2021-04-09
schema/gen/go: apply gofmt automatically

mvdan:
(see commit message)


mvdan:
This also means no longer having to be extra careful to write the templates to try to produce gofmt-valid output directly. So I imagine we could simplify some of the generation code to stop trying to do that.

warpfork:
Oh.  Not even an external command invocation.  Nice!

warpfork (approved):
:ship:

PR #164, opened 2021-04-26
fluent/qp: don't panic on string panics

mvdan:
(see commit message)


warpfork (approved):
Grumpy that this is relevant, but, ayy, golang.

mvdan:
panics use the empty interface :) you can't assume they are errors.

PR #166, opened 2021-04-26
Allow emitting & parsing of bytes per dagjson codec spec

### LINKED ISSUE #165: dag-json should support bytes types
the dag json spec defines a canonical encoding for bytes: https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-json.md#bytes-kind

it appears the implementation does not support this piece of the spec.

willscott:
fix #165

The 'allowLinks' flag to `marshal`/`unmarshal` seems at this point to have extended past it's original name. Would happily take suggestions for a better a name.

### DIFF
--- a/codec/dagjson/marshal.go
+++ b/codec/dagjson/marshal.go
@@ -1,6 +1,7 @@
 package dagjson
 
 import (
+	"encoding/base64"
 	"fmt"
 
 	"github.com/polydawn/refmt/shared"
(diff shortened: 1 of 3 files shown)


warpfork:
Do we have test vectors for this yet?

I'd be especially interested in what happens for pernicious data like: `{"/":{"bytes":{"/":{"bytes":"abcdef", "oh no": "combo breaker"}}}}`

willscott:
You can see the round-trip test I put in the PR. that's the test vector i included.
If there are others, I don't know where they are. Do you?

rvagg:
here was my start at that: https://github.com/ipld/js-dag-json/commit/ef139b7d20bd7cb2086c3e6b0b7246822fd00c1e#diff-ab28b86b8997575fa7518cf8b4a311e14b64521cfe7bd1c229a9b9621bf00e78 - branch is `rvagg/cborg` not in a PR yet but that branch supports https://github.com/ipld/specs/pull/356 which we've all agreed on

rvagg:
```js
      assert.throws(() => decode(encode({ '/': link.toString(), x: 'bip' })))
      assert.throws(() => decode(encode({ '/': { bytes: 'mS7ldeA', x: 'bip' } })))
      assert.throws(() => decode(encode({ '/': { bytes: 'mS7ldeA' }, x: 'bip' })))
      assert.throws(() => decode(encode({ '/': { bytes: 'mS7ldeA', x: 'bip' }, bop: 'bip' })))
```

there's your combo-breaker

willscott:
this is more strict than the current implementation.
Since these do not match our strict matching requirements, they are loaded as maps, and are not parsed in all cases as either bytes or links.

I would argue that adding that strictness is a separate PR from this one. I signed up to be able to interpret bytes through a round trip of IPFS / ipld-prime, not fully-fix dag-json

warpfork:
(d'oh, I was blind on my first review and didn't see some of the tests, sorry for that.)

So the one case I proposed as being specifically interested in is trying to hammer: encountering the maximally confusing things several times in a row, and only getting to the combo breaker at the very end.  The reason I want to poke that case in particular is to make sure we don't have exceptionally weird bugs that emerge from the limited look-behind window.  Given the state of the specs otherwise, that's the only thing I really feel compelled to push on.

willscott:
To confirm: in the case you propose, it will be parsed simply as a map, and will not gain bytes because it does not match our strict-match criteria. In the same way it would not be matched or rejected by the link detection, it will likewise not be matched or rejected by the new byte detection.

warpfork (approved):
Thanks for the additional tests.  I think I'm not scared of this now!  :+1:

PR #168, opened 2021-05-11
Add multibase prefix to DAG-JSON bytes

rvagg:
To add to the discussion @ https://github.com/ipld/specs/pull/356
An option .. demonstrating that it's not so painful (well .. there's a bit more but I'll explain there)


rvagg:
overruled @ https://github.com/ipld/specs/pull/356

also just noticing now that it pretty-prints the JSON output, which is another spec difference I need to note

PR #170, opened 2021-05-14
node/bindnode: start of a reflect-based Node implementation

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -1,0 +1,1112 @@
+package bindnode
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	ipld "github.com/ipld/go-ipld-prime"
+	basicnode "github.com/ipld/go-ipld-prime/node/basic"
+	"github.com/ipld/go-ipld-prime/schema"
+)
+
+// WrapNoSchema implements an ipld.Node given a pointer to a Go value.
+//
+// Same rules as PrototypeNoSchema apply.
+func WrapNoSchema(ptr interface{}) ipld.Node {
+	panic("TODO")
+	// ptrVal := reflect.ValueOf(ptr)
+	// if ptrVal.Kind() != reflect.Ptr {
+	// 	panic("must be a pointer")
+	// }
+	// return &_node{val: ptrVal.Elem()}
+}
+
+// Unwrap takes an ipld.Node implemented by one of the Wrap* or Prototype* APIs,
+// and returns a pointer to the inner Go value.
+//
+// Unwrap returns the input node if the node isn't implemented by this package.
+func Unwrap(node ipld.Node) (ptr interface{}) {
+	var val reflect.Value
+	switch node := node.(type) {
+	case *_node:
+		val = node.val
+	case *_nodeRepr:
+		val = node.val
+	default:
+		return node
+	}
+	if val.Kind() == reflect.Ptr {
+		panic("didn't expect val to be a pointer")
+	}
+	if !val.CanAddr() {
+		// Not addressable? Just return the interface as-is.
+		// TODO: This happens in some tests, figure out why.
+		return val.Interface()
+	}
+	return val.Addr().Interface()
+}
+
+// PrototypeNoSchema implements an ipld.NodePrototype given a Go pointer type.
+//
+// In this form, no IPLD schema is used; it is entirely inferred from the Go
+// type.
+//
+// Go types map to schema types in simple ways: Go string to schema String, Go
+// []byte to schema Bytes, Go struct to schema Map, and so on.
+//
+// A Go struct field is optional when its type is a pointer. Nullable fields are
+// not supported in this mode.
+func PrototypeNoSchema(ptrType interface{}) ipld.NodePrototype {
+	panic("TODO")
+	// typ := reflect.TypeOf(ptrType)
+	// if typ.Kind() != reflect.Ptr {
+	// 	panic("must be a pointer")
+	// }
+	// return &_prototype{goType: typ.Elem()}
+}
+
+// PrototypeOnlySchema implements an ipld.NodePrototype given an IPLD schema type.
+//
+// In this form, Go values are constructed with types inferred from the IPLD
+// schema, like a reverse of PrototypeNoSchema.
+func PrototypeOnlySchema(schemaType schema.Type) ipld.NodePrototype {
+	goType := inferGoType(schemaType)
+	return prototype(goType, schemaType)
+}
+
+// from IPLD Schema field names like "foo" to Go field names like "Foo".
+func fieldNameFromSchema(name string) string {
+	return strings.Title(name)
+}
+
+func inferGoType(typ schema.Type) reflect.Type {
+	switch typ := typ.(type) {
+	case *schema.TypeInt:
+		return goTypeInt
+	case *schema.TypeString:
+		return goTypeString
+	case *schema.TypeStruct:
+		fields := typ.Fields()
+		goFields := make([]reflect.StructField, len(fields))
+		for i, field := range fields {
+			ftyp := inferGoType(field.Type())
+			if field.IsNullable() {
+				ftyp = reflect.PtrTo(ftyp)
+			}
+			if field.IsOptional() {
+				ftyp = reflect.PtrTo(ftyp)
+			}
+			goFields[i] = reflect.StructField{
+				Name: fieldNameFromSchema(field.Name()),
+				Type: ftyp,
+			}
+		}
+		return reflect.StructOf(goFields)
+	case *schema.TypeMap:
+		ktyp := inferGoType(typ.KeyType())
+		vtyp := inferGoType(typ.ValueType())
+		if typ.ValueIsNullable() {
+			vtyp = reflect.PtrTo(vtyp)
+		}
+		return reflect.MapOf(ktyp, vtyp)
+	case *schema.TypeList:
+		etyp := inferGoType(typ.ValueType())
+		if typ.ValueIsNullable() {
+			etyp = reflect.PtrTo(etyp)
+		}
+		return reflect.SliceOf(etyp)
+	case *schema.TypeUnion:
+		type goUnion struct {
+			Index int // 0..len(typ.Members)-1
+			Value interface{}
+		}
+		return reflect.TypeOf(goUnion{})
+	}
+	panic(fmt.Sprintf("%T\n", typ))
+}
+
+// Prototype implements an ipld.NodePrototype given a Go pointer type and an
+// IPLD schema type.
+//
+// In this form, it is assumed that the Go type and IPLD schema type are
+// compatible. TODO: check upfront and panic otherwise
+func Prototype(ptrType interface{}, schemaType schema.Type) ipld.NodePrototype {
+	goPtrType := reflect.TypeOf(ptrType)
+	if goPtrType.Kind() != reflect.Ptr {
+		panic("ptrType must be a pointer")
+	}
+	return prototype(goPtrType.Elem(), schemaType)
+}
+
+func prototype(goType reflect.Type, schemaType schema.Type) ipld.NodePrototype {
+	if goType.Kind() == reflect.Invalid {
+		panic("goType must be valid")
+	}
+	if schemaType == nil {
+		panic("schemaType must not be nil")
+	}
+	return &_prototype{schemaType: schemaType, goType: goType}
+}
+
+var (
+	_ ipld.NodePrototype = (*_prototype)(nil)
+
+	_ ipld.Node        = (*_node)(nil)
+	_ schema.TypedNode = (*_node)(nil)
+
+	_ ipld.NodeBuilder   = (*_builder)(nil)
+	_ ipld.NodeAssembler = (*_assembler)(nil)
+
+	_ ipld.MapAssembler = (*_structAssembler)(nil)
+	_ ipld.MapIterator  = (*_structIterator)(nil)
+
+	_ ipld.ListAssembler = (*_listAssembler)(nil)
+	_ ipld.ListIterator  = (*_listIterator)(nil)
+)
+
+type _prototype struct {
+	schemaType schema.Type
+	goType     reflect.Type // non-pointer
+}
+
+func (w *_prototype) NewBuilder() ipld.NodeBuilder {
+	return &_builder{_assembler{
+		schemaType: w.schemaType,
+		val:        reflect.New(w.goType).Elem(),
+	}}
+}
+
+// TODO: consider these Typed interfaces for the schema package
+
+type TypedPrototype interface {
+	ipld.NodePrototype
+
+	Representation() ipld.NodePrototype
+}
+
+type TypedAssembler interface {
+	ipld.NodeAssembler
+
+	Representation() ipld.NodeAssembler
+}
+
+func (w *_prototype) Representation() ipld.NodePrototype {
+	return (*_prototypeRepr)(w)
+}
+
+var (
+	goTypeString = reflect.TypeOf("")
+	goTypeInt    = reflect.TypeOf(int(0))
+	goTypeBytes  = reflect.TypeOf([]byte{})
+	goTypeLink   = reflect.TypeOf((*ipld.Link)(nil)).Elem()
+
+	schemaTypeFieldName = schema.SpawnString("fieldNameString")
+)
+
+type _node struct {
+	schemaType schema.Type
+
+	val reflect.Value // non-pointer
+}
+
+// TODO: only expose TypedNode methods if the schema was explicit.
+// type _typedNode struct {
+// 	_node
+// }
+
+func (w *_node) Type() schema.Type {
+	return w.schemaType
+}
+
+func (w *_node) Representation() ipld.Node {
+	return (*_nodeRepr)(w)
+}
+
+func (w *_node) Kind() ipld.Kind {
+	return w.schemaType.TypeKind().ActsLike()
+}
+
+func (w *_node) LookupByString(key string) (ipld.Node, error) {
+	switch typ := w.schemaType.(type) {
+	case *schema.TypeStruct:
+		field := typ.Field(key)
+		if field == nil {
+			return nil, ipld.ErrInvalidKey{
+				TypeName: typ.Name().String(),
+				Key:      basicnode.NewString(key),
+			}
+		}
+		fval := w.val.FieldByName(fieldNameFromSchema(key))
+		if !fval.IsValid() {
+			panic("TODO: go-schema mismatch")
+		}
+		if field.IsOptional() {
+			if fval.IsNil() {
+				return ipld.Absent, nil
+			}
+			fval = fval.Elem()
+		}
+		if field.IsNullable() {
+			if fval.IsNil() {
+				return ipld.Null, nil
+			}
+			fval = fval.Elem()
+		}
+		node := &_node{
+			schemaType: field.Type(),
+			val:        fval,
+		}
+		return node, nil
+	case *schema.TypeMap:
+		var kval reflect.Value
+		switch ktyp := typ.KeyType().(type) {
+		case *schema.TypeString:
+			kval = reflect.ValueOf(key)
+		default:
+			asm := &_assembler{
+				schemaType: ktyp,
+				val:        reflect.New(w.val.Type().Key()).Elem(),
+			}
+			if err := (*_assemblerRepr)(asm).AssignString(key); err != nil {
+				return nil, err
+			}
+			kval = asm.val
+		}
+		fval := w.val.MapIndex(kval)
+		if !fval.IsValid() { // not found
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		// TODO: Error/panic if fval.IsNil() && !typ.ValueIsNullable()?
+		// Otherwise we could have two non-equal Go values (nil map,
+		// non-nil-but-empty map) which represent the exact same IPLD
+		// node when the field is not nullable.
+		if typ.ValueIsNullable() {
+			if fval.IsNil() {
+				return ipld.Null, nil
+			}
+			fval = fval.Elem()
+		}
+		node := &_node{
+			schemaType: typ.ValueType(),
+			val:        fval,
+		}
+		return node, nil
+	case *schema.TypeUnion:
+		var idx int
+		var mtyp schema.Type
+		for i, member := range typ.Members() {
+			if member.Name().String() == key {
+				idx = i
+				mtyp = member
+				break
+			}
+		}
+		if mtyp == nil { // not found
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		haveIdx := int(w.val.FieldByName("Index").Int())
+		if haveIdx != idx { // mismatching type
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		mval := w.val.FieldByName("Value").Elem()
+		node := &_node{
+			schemaType: mtyp,
+			val:        mval,
+		}
+		return node, nil
+	}
+	return nil, ipld.ErrWrongKind{
+		TypeName:   w.schemaType.Name().String(),
+		MethodName: "LookupByString",
+		// TODO
+	}
+}
+
+func (w *_node) LookupByNode(key ipld.Node) (ipld.Node, error) {
+	panic("TODO: LookupByNode")
+}
+
+func (w *_node) LookupByIndex(idx int64) (ipld.Node, error) {
+	switch typ := w.schemaType.(type) {
+	case *schema.TypeList:
+		if idx < 0 || int(idx) >= w.val.Len() {
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		// TODO: bounds check
+		val := w.val.Index(int(idx))
+		if typ.ValueIsNullable() {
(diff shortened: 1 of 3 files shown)


mvdan:
To verify that this works:

```
$ cd schema/gen/go
$ go test -tags fluentmirror
PASS
ok      github.com/ipld/go-ipld-prime/schema/gen/go     0.010s
```

schema/gen/go/testMaps_test.go
```diff
@@ -58,7 +59,7 @@ func TestMapsContainingMaybe(t *testing.T) {
 					ma.AssembleEntry("one").AssignString("1")
 					ma.AssembleEntry("two").AssignString("2")
 				})
-				Wish(t, n, ShouldEqual, nr)
+				Wish(t, bind.Unwrap(n), ShouldEqual, bind.Unwrap(nr))
```
warpfork, line 62: Okay, so I realize this was probably a "fasted way to adapt these tests for reuse" thing, but it's a pretty goodsized wat.

I'd like to have these tests not needing to import the bind package.  To do so is just a really weird shape for the import graph.

warpfork: In fact, it would be pretty nice to remove these from the `schema/gen` prefix, since we're now seeing they're definitely more broadly applicable than that (rather than just hopefully so :sweat_smile: ).

I don't know how much work that is / if we want to couple it to this diff.  If we _could_ get these extracted in one PR, and then use it for the new bind features in an immediately subsequent PR, though, that'd be super dreamy.

fluent/bind/mirror.go
```diff
@@ -0,0 +1,1099 @@
+package bind
```
warpfork, line 1: I'd move this to `node/bind` instead of `fluent/bind`.

(And possibly declare it as `package bindnode`, to match theme with `package basicnode`... although if you want to argue that `basicnode` shouldn't have been named like that, yeah, okay, maybe so.)

It's a node implementation moreso than a way of working with other node implementations.  So definitely goes under `node/*` rather than `fluent/*`.

mvdan: Fair enough. I'll put it under node/bindnode though, because `import ".../bindnode"` is always going to be shorter and simpler than `import bindnode ".../bind"`.

schema/gen/go/testMaps_test.go
mvdan: As for the DeepEqual itself, I don't think it's okay for the tests to assume that will work on any ipld.Node implementation. Maybe the "engine" passed to the nodetest package should be an interface, and another method could be "get me the underlying comparable value of this node", or "tell me if these two nodes are equal".

Another option is to declare a semi-standard "Unwrap" optional method on Node types. I have a TODO for that on my code, but I felt like it was too early to bring that up.

fluent/bind/mirror.go
mvdan: Done.

schema/gen/go/testMaps_test.go
mvdan: Done; removed all the test changes from this PR, they'll come in a follow-up PR.

mvdan:
I still don't intend to continue writing code/tests/examples until Friday, but if we could review and merge this today, that would help - then tomorrow I can send the tests PR, and I can continue on Friday without being blocked by the stack of changes that are still up for review and might need changes.

schema/gen/go/testMaps_test.go
warpfork: Yes... I think we could also create a generic `Equals` function for `ipld.Node` (traversals of them are well defined after all) and that would be a generally reasonable and useful thing to export even beyond tests.  Or, even more powerfully (and helpful for debugging), a `Diff` feature, though that's more work.

We could also go further and make [`Checker`](https://pkg.go.dev/github.com/frankban/quicktest#Checker) implementations (or, [`Checker`](https://pkg.go.dev/github.com/warpfork/go-wish#Checker), depending on how quickly we migrate testing libraries) which use that feature.

mvdan: Oh, right, we could have our own `ipld.DeepEqual` which would do the same kind of nested comparison that `reflect.DeepEqual` does, but at the IPLD data model layer instead of Go reflection layer.

warpfork: Exactly :D

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+	goType := inferGoType(schemaType)
+	return prototype(goType, schemaType)
+}
+
+// from IPLD Schema field names like "foo" to Go field names like "Foo".
+func fieldNameFromSchema(name string) string {
+	return strings.Title(name)
+}
+
+func inferGoType(typ schema.Type) reflect.Type {
```
warpfork, line 83: Short notes on what this is for would be helpful.  I see what it does, but harder to see _why_ it's necessary to do the things it does.

I think I see it mostly being used by... code that's related to unions?

Okay, and `PrototypeOnlySchema`.  Which I guess is pretty significant.

When I started writing this comment, I was wondering if the behaviour for unions is being used in a special enough way it should be separated from the rest of this, but I think I've talked myself back out of that already, so this is just a thinking-out-loud note now :)

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+	// 	panic("must be a pointer")
+	// }
+	// return &_prototype{goType: typ.Elem()}
+}
+
+// PrototypeOnlySchema implements an ipld.NodePrototype given an IPLD schema type.
+//
+// In this form, Go values are constructed with types inferred from the IPLD
+// schema, like a reverse of PrototypeNoSchema.
+func PrototypeOnlySchema(schemaType schema.Type) ipld.NodePrototype {
```
warpfork, line 73: This is *wild*.  :D  (You already blew my mind a couple days ago when telling me `reflect.StructOf` exists, you probably remember, heh.)

I have some rambling questions as I try to fit my head around when one would use this:

Do we know if this is going to be able to work for all schema typekinds and all their representations?  If there are any typekinds or representation strategies it's *not* going to be able to support, that could be a serious limitation of this entire feature (... possibly enough to make it questionable whether the feature is worth supporting?).

There are going to be some pretty interesting caveats on the usability of this too, right?  Like if it returns a struct... it's going to be some kind of slightly odd anonymous type, which can really only be usefully interacted with via more reflection or our Node interfaces?  ... Okay, I guess that may still sometimes be pretty useful, because it lets one get nice compact native memory layouts without having to use codegen.  That's cool.  But this feature wouldn't really be able to provide a "developer experience" win, right?

warpfork (commented):
I'm thinking a hunk of documentation somewhere talking about the pros and cons of different ways to start using this could be important.

There's a huge comparison to be made vs basicnode (e.g. basicnode is general, but not efficient, and has zero DX bonuses), and a huge comparison to be made vs codegen (e.g. codegen is superfast, and has some DX bonuses (but other drawbacks), but has high friction to start using, and increases binary sizes)... and there's also _several_ different ways of using this reflection stuff, if I'm reading right (some of which has certain DX bonsues, and some which doesn't).

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+
+type _prototype struct {
+	schemaType schema.Type
+	goType     reflect.Type // non-pointer
+}
+
+func (w *_prototype) NewBuilder() ipld.NodeBuilder {
+	return &_builder{_assembler{
+		schemaType: w.schemaType,
+		val:        reflect.New(w.goType).Elem(),
```
warpfork, line 174: So there is basically never going to be a way to avoid this alloc in here, right?  Short of, maybe, some very fancy use of `unsafe`?

Maybe an acceptable cost for this system, just thinking outloud about how docs should talk about performance of this system.

mvdan:
> I'm thinking a hunk of documentation somewhere talking about the pros and cons of different ways to start using this could be important.

For sure. I plan to work on runnable examples after this, then finishing the important TODOs, and lastly high-level docs - once we're happy with the APIs.

node/bindnode/node.go
mvdan: > Okay, and `PrototypeOnlySchema`. Which I guess is pretty significant.

Right, that's the only use case. When supplying both a schema and a Go type, inferGoType should never be used - that API entrypoint is just not properly done yet.

And yeah, unions are special because they need the extra "tag", but otherwise they are handled like any other type.

node/bindnode/node.go
mvdan: > Do we know if this is going to be able to work for all schema typekinds and all their representations?

I'm pretty sure it should support all schema types, and by extension all representations too - representations are just some logic atop the same inferred Go types.

The big caveat is "what if I want my Go type to be slightly different", such as having your own Go type with methods, or using different types like `int` instead of `int64` for a field if you're absolutely sure the values will be small. That's where the answer will be "bring your own Go type and use the other API".

mvdan: Regarding anonymous types: it certainly has some disadvantages, like what I mentioned above. But the advantage is... it just works, without extra effort. No codegen, no having to declare the types yourself. If you have schema X and you want to convert data encoded in dagjson to dagcbor following the schema, you can do all of that with just the schema and PrototypeOnlySchema. The Go value you will hold in between will use anonymous types built at run-time, but for many simple use-cases, that's fine.

mvdan: But yes, for most real use cases like go-codec-dagpb where the type should be easy to use (methods, type switches, better error messages, etc), I do expect the code to provide both the schema and a Go type maintained by a human.

node/bindnode/node.go
mvdan: I mean, the Builder/Assembler themselves already alloc, as they are interfaces. And this is one alloc per built Node, so I don't think it's particularly worrysome.

I am not worrying about performance just yet. My high-level intuition is that I'll be able to allocate about the same, or perhaps a little less, than the codegen. Another question will be the actual performance, where allocs are not the only factor.

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+	switch typ := w.schemaType.(type) {
+	case *schema.TypeStruct:
+		field := typ.Field(key)
+		if field == nil {
+			return nil, ipld.ErrInvalidKey{
+				TypeName: typ.Name().String(),
+				Key:      basicnode.NewString(key),
+			}
+		}
+		fval := w.val.FieldByName(fieldNameFromSchema(key))
```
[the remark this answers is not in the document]
mvdan: Maybe. For the time being I think the only way to customize this will be... to bring your own Go type, and change the names to whatever you want, as long as the "shape" of the Go type still matches the schema.

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+func (w *_prototype) Representation() ipld.NodePrototype {
+	return (*_prototypeRepr)(w)
+}
+
+var (
+	goTypeString = reflect.TypeOf("")
+	goTypeBytes  = reflect.TypeOf([]byte{})
+	goTypeLink   = reflect.TypeOf((*ipld.Link)(nil)).Elem()
+
+	schemaTypeFieldName = schema.SpawnString("fieldNameString")
```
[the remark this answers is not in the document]
mvdan: I'm not allowed to spawn an anonymous schema string type, so I made up one :)

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+			val:        fval,
+		}
+		return node, nil
+	case *schema.TypeMap:
+		var kval reflect.Value
+		switch ktyp := typ.KeyType().(type) {
+		case *schema.TypeString:
+			kval = reflect.ValueOf(key)
+		default:
+			asm := &_assembler{
```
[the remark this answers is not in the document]
mvdan: I'm not worrying about alloc trickery just yet :) When I do, it will be driven by benchmarks anyway, of which I have none so far.

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		if typ.ValueIsNullable() {
+			if fval.IsNil() {
+				return ipld.Null, nil
+			}
+			fval = fval.Elem()
+		}
```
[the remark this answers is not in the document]
mvdan: I'll add a TODO about this, thanks. There are just a lot of edge cases to worry about, so I want to make sure this is covered by a test first.

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+	}
+	return nil, ipld.ErrWrongKind{
+		TypeName:   w.schemaType.Name().String(),
+		MethodName: "LookupByIndex",
+		// TODO
+	}
+}
+
+func (w *_node) LookupBySegment(seg ipld.PathSegment) (ipld.Node, error) {
+	panic("TODO: LookupBySegment")
```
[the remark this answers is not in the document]
mvdan: I imagined so, but I thought this PR was large enough as is :P 

Also, I wonder what would be the best way to test this, because the schema/gen/go tests don't use it. Maybe some of the selector test cases could also be made "generic" to run over both schema/gen/go and node/nodebind.

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+
+	val reflect.Value // non-pointer
+}
+
+// TODO: only expose TypedNode methods if the schema was explicit.
+// type _typedNode struct {
+// 	_node
+// }
+
+func (w *_node) Type() schema.Type {
```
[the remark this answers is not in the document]
mvdan: Because I initially called all these nodeWrapper, assemblerWrapper, etc. The names just got veeeery long and I didn't like them, so I went for _node, _assembler, etc.

I'm not too fond of `n _node` because `n` is usually `n int`, and `p _prototype` is usually `p []byte`, and `s _structAssembler` is usually `s string`... so I was a bit lazy and left that out for now, defaulting to `w` for all of them, which is almost never used in a standard way, and still gives the hint that these are "wrappers".

node/bindnode/node.go
```diff
@@ -0,0 +1,1099 @@
...
+				mtyp = member
+				break
+			}
+		}
+		if mtyp == nil { // not found
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		haveIdx := int(w.val.FieldByName("Index").Int())
```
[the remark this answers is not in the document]
mvdan: This would work if we assume we always infer the type, but longer term I want to allow the user more control. You can imagine allowing the user to provide a custom type to back a union by providing methods like `Get() (tag int, value interface{})` and `Put(tag int, value interface{})`.

node/bindnode/node.go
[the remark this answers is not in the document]
warpfork: Yep, those test batteries should still grow.  IIRC I did some work to make sure both iterators and lookup methods got exercised for the same data (with the newest form of tests, anyway), but maybe there's still a bit more to do there to make sure _all_ the lookup method variants are exercised too.

Getting selector tests to run over arbitrary nodes would also be super good.  I think there's a small barrier there right now, in that those tests only take one NodePrototype at the moment, and assume they can use it for both selector data and the test data, and that needs fixing... tractable but definitely enough of a diff to make it its own little quest.

node/bindnode/node.go
```diff
@@ -0,0 +1,1103 @@
...
+	// TODO: does this ever trigger?
+	// newVal := reflect.ValueOf(node)
+	// if newVal.Type().AssignableTo(w.val.Type()) {
+	// 	w.val.Set(newVal)
+	// 	return nil
+	// }
+	switch node.Kind() {
+	case ipld.Kind_Map:
+		itr := node.MapIterator()
+		am, err := w.BeginMap(-1) // TODO: length?
```
warpfork, line 672: I think I would use node.Length here, yeah.

Also: I've been considering extracting this whole hunk below to a generic function like, say, `ipld.BlindCopyMap`, because it does come up in a couple places now.  (The other big example is similarly, in the AssignNode methods emitted by codegen, this logic exists as a fallback path right after trying to do direct copy for known types.)  So if that looks reasonable to extract, go for it.

node/bindnode/node.go
```diff
@@ -0,0 +1,1103 @@
...
+
+type _structAssembler struct {
+	// TODO: embed _assembler?
+
+	schemaType *schema.TypeStruct
+	val        reflect.Value // non-pointer
+	finish     func() error
+
+	// TODO: more state checks
+	doneFields []bool // TODO: replace with just reflect.Value field pointer checks
```
warpfork, line 749: I think this TODO is probably overoptimistic.  If a field has either the nullable or optional properties in the schema, checking the go field for nil doesn't provide any useful information about if it was assigned or not.

Keeping a bitfield explicitly is also just going to be a lot simpler code, I suspect.

node/bindnode/node.go
mvdan: I sorta left these TODOs as reminders for future optimizations, and as said before, I have no benchmarks. From what you say, we might dedup this anyway :)

I'll add this refactor as another TODO.

node/bindnode/node.go
mvdan: Ah, you're right, I keep forgetting about AssignNull which does assign into a node, but explicitly keeps it null.

I still think there's room for improvement here potentially, but I also don't think it's going to be a bottleneck soon. It would always be possible to avoid an extra allocation for most structs by backing this slice with a `[16]bool` in the struct., for example.

I'll leave this TODO about reminding myself to look at this at some point.

mvdan:
I added a runnable example, but I don't get why it's failing...

```
--- FAIL: ExamplePrototypeOnlySchema (0.00s)
got:
{
        "Name": "Michael",
        "Friends": [
                "Sarah",
                "Alex"
        ]
}
want:
{
        "Name": "Michael",
        "Friends": [
                "Sarah",
                "Alex"
        ]
}
FAIL
exit status 1
FAIL    github.com/ipld/go-ipld-prime/node/bindnode     0.002s
```

node/bindnode/repr.go
```diff
@@ -0,0 +1,814 @@
...
+		return reprNode(v), nil
+	default:
+		panic(fmt.Sprintf("TODO: %T", stg))
+	}
+}
+
+func (w *_nodeRepr) LookupBySegment(seg ipld.PathSegment) (ipld.Node, error) {
+	return nil, ipld.ErrWrongKind{
+		TypeName:   w.schemaType.Name().String(),
+		MethodName: "LookupBySegment", AppropriateKind: ipld.KindSet_JustList, ActualKind: ipld.Kind_Map,
```
warpfork, line 184: Should be `AppropriateKind: ipld.KindSet_Recursive` among other things.  Probably this whole method is actually a todo right now?

node/bindnode/node.go
```diff
@@ -0,0 +1,1112 @@
...
+			}
+		}
+		if mtyp == nil { // not found
+			return nil, ipld.ErrNotExists{
+				// TODO
+			}
+		}
+		haveIdx := int(w.val.FieldByName("Index").Int())
+		if haveIdx != idx { // mismatching type
+			return nil, ipld.ErrNotExists{
```
willscott, line 314: more specificity on attempted vs reality would indeed be useful here :)

node/bindnode/node.go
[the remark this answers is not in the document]
willscott: I'd be in favor to see this done in a subsequent PR once there are tests for that path.

node/bindnode/node.go
```diff
@@ -0,0 +1,1112 @@
...
+			val:        w.val,
+		}
+	case *schema.TypeUnion:
+		return &_unionIterator{
+			schemaType: typ,
+			members:    typ.Members(),
+			val:        w.val,
+		}
+	case *schema.TypeMap:
+		panic("TODO: ")
```
willscott, line 380: this one probably does need to be in this PR

node/bindnode/node.go
```diff
@@ -0,0 +1,1112 @@
...
+
+// TODO: consider making our own Node interface, like:
+//
+// type WrappedNode interface {
+//     ipld.Node
+//     Unwrap() (ptr interface)
+// }
+//
+// Pros: API is easier to understand, harder to mix up with other ipld.Nodes.
+// Cons: One usually only has an ipld.Node, and type assertions can be weird.
```
willscott, line 1112: the current Unwrap seems like a reasonable interface to me.

node/bindnode/node.go
```diff
@@ -0,0 +1,1112 @@
...
+	}
+	key = basicnode.NewString(mtyp.Name().String())
+	return key, node, nil
+}
+
+func (w *_unionIterator) Done() bool {
+	return w.done
+}
+
+// TODO: consider making our own Node interface, like:
```
willscott, line 1104: given the length of this file, it might be worth splitting it into per-type files. e.g `node_union`, `node_map`, `node_list` for the type-specific assemblers / iterators.

node/bindnode/repr.go
```diff
@@ -0,0 +1,814 @@
...
+	case schema.UnionRepresentation_Stringprefix:
+		return ipld.Kind_String
+	case nil:
+		return (*_node)(w).Kind()
+	default:
+		panic(fmt.Sprintf("TODO Kind: %T", stg))
+	}
+}
+
+func outboundMappedKey(stg schema.StructRepresentation_Map, key string) string {
```
willscott, line 67: a comment explaining the need for inbound vs outbound here would be useful.

node/bindnode/repr.go
```diff
@@ -0,0 +1,814 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return reprNode(v), nil
+	default:
+		panic(fmt.Sprintf("TODO: %T", stg))
+	}
+}
+
+func (w *_nodeRepr) LookupByNode(key ipld.Node) (ipld.Node, error) {
```
willscott, line 153: this should work when the key is a string node, right?

node/bindnode/repr.go
mvdan: Yeah, a lot of these are copy pasted and TODOs. I'll make it explicit.

node/bindnode/node.go
```diff
@@ -0,0 +1,1112 @@
...
+func PrototypeNoSchema(ptrType interface{}) ipld.NodePrototype {
+	panic("TODO")
+	// typ := reflect.TypeOf(ptrType)
+	// if typ.Kind() != reflect.Ptr {
+	// 	panic("must be a pointer")
+	// }
+	// return &_prototype{goType: typ.Elem()}
+}
+
+// PrototypeOnlySchema implements an ipld.NodePrototype given an IPLD schema type.
```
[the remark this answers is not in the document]
mvdan: The intent is to make things explicit. Prototype vs PrototypeSchema could also be relatively ambiguous, if one doesn't look at the signatures.

I don't feel strongly about these yet and most of the exposed API is still TODO, so I'm happy to take ideas. I'll probably focus on more tests and finishing the code first before worrying about names and docs, though :)

node/bindnode/node.go
```diff
@@ -0,0 +1,1112 @@
...
+func WrapNoSchema(ptr interface{}) ipld.Node {
+	panic("TODO")
+	// ptrVal := reflect.ValueOf(ptr)
+	// if ptrVal.Kind() != reflect.Ptr {
+	// 	panic("must be a pointer")
+	// }
+	// return &_node{val: ptrVal.Elem()}
+}
+
+// Unwrap takes an ipld.Node implemented by one of the Wrap* APIs, and returns
```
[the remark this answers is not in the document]
mvdan: Ah, good catch. The original API was like `fluent.WrapNoSchema` etc.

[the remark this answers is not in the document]
mvdan: Ah, wait, we do still have WrapNoSchema. I just haven't done anything with it yet, because the codegen tests only require prototypes.

But this doc still needs updating.

node/bindnode/node.go
mvdan: It's hard to draw the line. This PR is huge as is, so I drew the line at "what is necessary to make the tests pass". I plan to continue fixing TODOs as I add more tests in future PRs. If we decide that this many TODOs are not acceptable in master, then I'd pretty easily slip into 5k-line PR territory :)

node/bindnode/node.go
mvdan: Splitting is a good point. I just kept going and didn't really think about file sizes that much.

Maybe node_scalar and node_recursive would be better for now, and later on I can split recursive into map/list/union if necessary. It feels a bit silly to have e.g. node_bool or node_int.

node/bindnode/repr.go
mvdan: These are pretty much just hacks, since the schema API gives the mappings one way but not the other. I'll add a TODO about working this out with Eric - I didn't add docs because I don't intend to maintain nor expose this code here.

node/bindnode/repr.go
mvdan: Indeed, but the same I mentioned elsewhere about tests and TODOs applies :) This PR is by no means attempting to be complete

mvdan:
@warpfork are you up for merging today? The example is passing, and I've replied to all the review comments. I know I replied to many with "yep, in the next PRs", but I really do want to work incrementally rather than keep expanding this PR.

warpfork (approved):
Yep, I'm good with this merging.  There's work still to do, but it's all encapsulated in one package, so we should not have difficulty making it reasonably clear in the interim.

PR #169, opened 2021-05-14
Add enumerate methods for the multicodec registries

willscott:
Adds `ListDecoders` and `ListEncoders` methods to learn which multicodecs have been registered with encoders/decoders.

PR #172, opened 2021-05-19
Make a multicodec.Registry type available.

warpfork:
The hope is that this might be helpful if you want to build a
multicodec registry other than the global one, but are still buying
into the idea of the registry being keyed by indicator numbers.

(I'm... not actually sure how useful this is, because I'd think if
you're building something other than the default, there's also a good
chance you'd want *more* features than a primitive numerical mapping,
which are probably going to be related to whatever your application
logic in the area is, and therefore not possible for this library
to usefully anticipate.  But, I dunno.  I'm doing this because
people are proposing attaching more features to the global, and
I'm not comfortable with it, and I'm hoping this will provide a
pressure release valve.)

The current interfaces are functionally unchanged.

The multicodec.Registry type can now be used when constructing
a cidlink.LinkSystem.  (This saves you from having to write the
glue functions to unbox cidlink and then do the LookupEncoder
and LookupDecode calls.)


willscott:
this doesn't help with the need in #169 though? I still can't enumerate what's in a given registry, right?

multicodec/registry.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// go-ipld-prime also has a default registry, which has the same methods as this structure, but are at package scope.
+// Some systems, like cidlink.DefaultLinkSystem, will use this default registry.
+// This Registry type is for helping if you wish to make your own registry which does not share that global state.
+//
+// Multicodec indicator numbers are specified in
+// https://github.com/multiformats/multicodec/blob/master/table.csv .
+// You should not use indicator numbers which are not specified in that table
+// (however, there is nothing in this implementation that will attempt to stop you, either; please behave).
+type Registry struct {
```
aschmahmann, line 23: It seems like we should specify the rules around when you can register encoders/decoders either here or in the LinkSystem constructor.

Similarly, should we be adding a mutex lock around the registry operations?

aschmahmann (changes requested):

aschmahmann:
@willscott correct, we'll still need #169, although the functions will end up on the struct instead of (or in addition to) globally.

multicodec/registry.go
warpfork: I've added a glut of additional documentation about the synchronization (or rather, lack of it, and caller's responsibility).

Issue #173, opened 2021-05-21
provide a standard way to compare Nodes for equality


mvdan:
Right now we only have one ipld.Node implementation: it's code-generated via `schema/gen/go`, and it's assumed by default that we can use `reflect.DeepEqual` on it. Some of the tests do this, such as when doing a codec encode-decode roundtrip and checking that the node looks the same.

However, this breaks down with other Node implementations. Here are two examples:

1) ADLs, where the Go value could be... anything. Moreover, if two ADLs behave like nodes with different kinds, but their Go values happen to be the same, should they really be considered equal nodes? I imagine not.

2) Nodes built via reflection, as per https://github.com/ipld/go-ipld-prime/pull/170. Since those "wrapper" Go values also include a reference to the IPLD schema, and schemas can be cyclic... DeepEqual panics.

I think the only reasonable solution to this is to define a high-level `ipld.DeepEqual` function, which is similar to `reflect.DeepEqual`, but works on IPLD kinds and values instead. For example, a basic IPLD string "foo" and a fancy custom-built ADL whose kind is string and whose AsString returns "foo", would be equal as far as this API is concerned.

I don't think it would be necessary to support options for this API. If we really think we'd need options, I think we should either rely on the traversal APIs for this, or reuse the existing semi-standard [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp#Options).

I discarded an optional `Equal(ipld.Node) bool` method on nodes because this would be repetitive, likely inconsistent, and could result in weirdness like `a.Equal(b) == true && b.Equal(a) == false`. Some Go APIs like crypto/... went this route, but note that they are a single implementation forever, so they are always dealing with just one non-interface type.

cc @warpfork @willscott @hannahhoward

mvdan:
I'm open to better names, by the way. The behavior would essentially be:

* kinds match
* if the kind is scalar, the values must be equal as per Go's `==` operator
* if the kind is recursive (maps, lists, ...), the values must be recursively equal as per the rules above

So maybe `ipld.KindDeepEqual` if we want to make it very clear that this is all around how a node represents itself via the IPLD interface, and not its underlying Go value or type.

willscott:
`DeepEqual` seems like a reasonable addition to me

warpfork:
sounds perfect.

mvdan:
Cool, I'll probably do this as part of my IPLD schema+node test refactor next week, because otherwise I can't use the tests with the reflection implementation.

PR #174, opened 2021-05-24
add DeepEqual and start using it in tests

mvdan:
(see commit message)

### DIFF
--- a/equal.go
+++ b/equal.go
@@ -1,0 +1,153 @@
+package ipld
+
+// DeepEqual reports whether x and y are "deeply equal" as IPLD nodes.
+// This is similar to reflect.DeepEqual, but based around the Node interface.
+//
+// Two nodes must have the same kind to be deeply equal.
+// If either node has the invalid kind, the nodes are not deeply equal.
+//
+// Two nodes of scalar kinds (null, bool, int, float, string, bytes, link)
+// are deeply equal if their Go values, as returned by AsKind methods, are equal as
+// per Go's == comparison operator.
+//
+// Note that Links are compared in a shallow way, without being followed.
+// This will generally be enough, as it's rare to have two different links to the
+// same IPLD data by using a different codec or multihash type.
+//
+// Two nodes of recursive kinds (map, list)
+// must have the same length to be deeply equal.
+// Their elements, as reported by iterators, must be deeply equal.
+// The elements are compared in the iterator's order,
+// meaning two maps sorting the same keys differently might not be equal.
+//
+// Note that this function panics if either Node returns an error.
+// We only call valid methods for each Kind,
+// so an error should only happen if a Node implementation breaks that contract.
+// It is generally not recommended to call DeepEqual on ADL nodes.
+func DeepEqual(x, y Node) bool {
+	xk, yk := x.Kind(), y.Kind()
+	if xk != yk {
+		return false
+	}
+
+	switch xk {
+
+	// Scalar kinds.
+	case Kind_Null:
+		return x.IsNull() == y.IsNull()
+	case Kind_Bool:
+		xv, err := x.AsBool()
+		if err != nil {
+			panic(err)
+		}
+		yv, err := y.AsBool()
+		if err != nil {
+			panic(err)
+		}
+		return xv == yv
+	case Kind_Int:
+		xv, err := x.AsInt()
+		if err != nil {
+			panic(err)
+		}
+		yv, err := y.AsInt()
+		if err != nil {
+			panic(err)
+		}
+		return xv == yv
+	case Kind_Float:
+		xv, err := x.AsFloat()
+		if err != nil {
+			panic(err)
+		}
+		yv, err := y.AsFloat()
+		if err != nil {
+			panic(err)
+		}
+		return xv == yv
+	case Kind_String:
+		xv, err := x.AsString()
+		if err != nil {
+			panic(err)
+		}
+		yv, err := y.AsString()
+		if err != nil {
+			panic(err)
+		}
+		return xv == yv
+	case Kind_Bytes:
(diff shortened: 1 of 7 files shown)


mvdan:
@warpfork this is my take on #173, needed for the schema/gen/go test refactor.

Once you review, I'll add tests and write a good commit message, and then we can merge.

Worth noting that the quality of the test checks went down when nodes aren't equal:

```
        --- FAIL: TestStructNesting/bhvtest/repr-create (0.00s)
            testStruct_test.go:136: ShouldEqual check rejected:
                          bool(
                        -       false,
                        +       true,
                          )
```

It no longer diffs, so it just tells us yes/no. However, we could easily improve this with https://pkg.go.dev/github.com/google/go-cmp/cmp#Comparer, since our DeepEqual does match the shape `func(T, T) bool`.

However, I could not get this to work with go-wish via `Wish(t, n, ShouldEqual, nr, cmp.Comparer(ipld.DeepEqual))`:

```
schema/gen/go/testLists_test.go:62:45: cannot use cmp.Comparer(ipld.DeepEqual) (type cmp.Option) as type wish.options in argument to wish.Wish:
        cmp.Option does not implement wish.options (missing wish._options method)
schema/gen/go/testLists_test.go:97:45: cannot use cmp.Comparer(ipld.DeepEqual) (type cmp.Option) as type wish.options in argument to wish.Wish:
        cmp.Option does not implement wish.options (missing wish._options method)
```

So I think we could do that if/when we move to quicktest, which definitely supports cmp options: https://pkg.go.dev/github.com/frankban/quicktest#CmpEquals

equal.go
```diff
@@ -0,0 +1,157 @@
...
+// Two nodes must have the same kind to be deeply equal.
+// If either node has the invalid kind, the nodes are not deeply equal.
+//
+// Two nodes of scalar kinds (null, bool, int, float, string, bytes)
+// are deeply equal as per the rules of reflect.DeepEqual.
+// In other words, their Go values are compared, as returned by methods such as
+// AsString and AsInt.
+//
+// Links are an exception, since this function does not load links.
+// As such, if either node has the link kind, the nodes are not deeply equal.
```
warpfork, line 15: I think this was discussed elsewhere already, but I'd advocate for checking link equality, just in a very literal way -- treating it like a scalar.

I think you can actually use `Link.String()` for this (which will avoid any need to look for pointers vs not, etc).  For CIDs, that should have the effect of ignoring multibase, but regarding the full digest, multicodec indicator, and everything else -- which should be about right.

equal.go
```diff
@@ -0,0 +1,157 @@
...
+// must have the same length to be deeply equal.
+// Their elements, as reported by iterators, must be deeply equal.
+// The elements are compared in the iterator's order,
+// meaning two maps sorting the same keys differently might not be equal.
+//
+// This function panics if any Node methods return a non-nil error.
+// It only calls valid methods for each node's kind,
+// so those errors should truly be rare and unexpected.
+//
+// Note that this function is mainly intended for testing purposes.
```
warpfork, line 27: I think we might skip this mention.  It's probably true, but there's nothing particularly regrettable about this function that means we need to discourage its use, as far as I can figure.

warpfork (commented):
Mostly solid except I think we'll need to do something about links.  If links always return false, that'll make a lot of blocks turn false, and it won't be very useful.

equal.go
mvdan: As discussed, I'll use == on these.

equal.go
mvdan: The panics :)

warpfork: Still.

We can document that if a node breaks contract badly (e.g. fails to answer AsString after saying Kind()==String) it panics... but that requirement is sufficiently baseline that I think it's acceptable.

warpfork (approved):
I'm good with this being merged as soon as the links get handled :)

equal.go
rvagg: Since there are lots of places where you're having to `panic()` would it be worth just returning `(error, bool)`? or is it really important to follow the `reflect.DeepEqual` pattern here? If you did that then it'd at least not have to come with caveats and could be used for legitimate runtime purposes (I could imagine up a bunch of ways this would be really useful outside of debugging).

mvdan: The thing with returning an error is that it can only happen if a Node implementation breaks the interface contract. For example, if calling AsInt on a Kind==Int node errors.

So those errors should be truly rare and unexpected, unless I'm missing something. Panic seems fine in that way.

Arguably, the interface could have been designed in a way that errors are not part of this equation, and one simply cannot call `AsString` on an integer kind node. One could do this with optional interfaces, for example, so that AsString is only implemented by nodes which support it, and it does not return an error.

Maybe ADLs would still need an error? I'm still not super clear on that. If we think DeepEqual might be used for ADLs, and it's good for them to return errors like these, then maybe we should make this return an error after all. Though, for the average use case, the error would always be nil.

mvdan:
This now has a proper commit message and its own tests. I've also implemented links as discussed. Needs a final review :)

equal.go
```diff
@@ -0,0 +1,157 @@
...
+
+// DeepEqual reports whether x and y are "deeply equal" as IPLD nodes.
+// This is similar to reflect.DeepEqual, but based around the Node interface.
+//
+// Two nodes must have the same kind to be deeply equal.
+// If either node has the invalid kind, the nodes are not deeply equal.
+//
+// Two nodes of scalar kinds (null, bool, int, float, string, bytes)
+// are deeply equal as per the rules of reflect.DeepEqual.
+// In other words, their Go values are compared, as returned by methods such as
```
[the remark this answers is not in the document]
mvdan: done

PR #179, opened 2021-06-01
node/tests: put most of the schema test cases here

mvdan:
(see commit message)


willscott (commented):
This looks primarily like a refactor. Are there new tests / functionality you want particularly close eyes on?

mvdan:
Yeah, it's purely a test refactor to unblock my reflection work. I added you mainly as a CC, in case you want to be in the loop or fancy doing a skim review of the "engine" interface :)

schema/gen/go/testEngine_disabled_test.go
```diff
@@ -33,14 +37,15 @@ func buildGennedCode(t *testing.T, prefix string, pkgName string) {
 	cmd := exec.Command("go", args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	err := cmd.Run()
+	err = cmd.Run()
 	if err != nil {
 		t.Fatalf("genned code failed to compile: %s", err)
 	}
+
+	t.Skip("behavioral tests for generated code skipped: you used the 'skipgenbehavtests' build tag.")
```
warpfork, line 45: I think this probably should still be wrapped in some t.Run clause?  Because _some_ things are still appropriate to flag as passing if we get this far.

schema/gen/go/testEngine_nocgo_test.go
```diff
@@ -48,14 +52,14 @@ func buildGennedCode(t *testing.T, prefix string, pkgName string) {
 	cmd := exec.Command("go", args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	err := cmd.Run()
+	err = cmd.Run()
 	if err != nil {
 		t.Fatalf("genned code failed to compile: %s", err)
 	}
+
+	t.Skip("behavioral tests for generated code skipped: cgo is required for these tests")
```
warpfork, line 60: Ditto to https://github.com/ipld/go-ipld-prime/pull/179/files#r643567693

node/tests/testEngine.go
```diff
@@ -0,0 +1,22 @@
...
+// Engine describes the interface that can be supplied to run tests on schemas.
+//
+// The PrototypeByName function can get its job done using only interface types
+// that we already know from outside any generated code, so you can write tests
+// that have no _compile time_ dependency on the generated code.  This makes it
+// easier for IDEs and suchlike to help you write and check the test functions.
+//
+// Ask for prototypes using the type name alone (no package prefix);
+// their representation prototypes can be obtained by appending ".Repr".
+type Engine interface {
```
warpfork, line 19: Just for discussion:

The other way I could imagine writing this would be more functionally: instead of this interface, which has to be held in a certain order, there could be a function like `Engine(t *testing.T, ts schema.TypeSystem) func(name string) ipld.NodePrototype`.

Usage would then involve holding onto that function returned, and using it the same places `engine.PrototypeByName` is used with this approach.

Already discussed this with @mvdan in another medium and he points out it's harder to add more methods in that approach.  Also true.

warpfork (approved):
I think I'm good with this.  I left some comments that are thinking out loud about changes that would get us even further towards ideal, but they're forward looking and not blocking.

The comments about the skips might still need some action, but other than that, :+1:

node/tests/testEngine.go
mvdan: Yep, and that would be akin to a constructor. I don't have a strong opinion on this, but I do find that "func that returns func with some state" can be a bit confusing, so I tend to avoid it whenever possible. Plus I think it's likely we'll extend this interface with time, perhaps to solve the problem involving repetition of test names etc.

schema/gen/go/testEngine_disabled_test.go
mvdan: Ah, will do, thanks.

mvdan: Actually, I can't do this, because there's no longer a `runBehavioralTests` layer with a sub-test. There's just engine Init + PrototypeByName.

I could go to every `func SchemaTest...` and wrap all the "test the prototypes" logic with a t.Run and asking the engine if it just wants to skip, but that's too much repeated boilerplate.

I could perhaps go back to a double-callback system where the engine is in charge of running all the behavioral tests, so it can run them in a sub-test and however it wants, but then the interface gets kinda complex and confusing.

I get that marking "up to this point" as passing can be useful, but I don't think it's to the point that we should turn the interface around that way. I'd rather have us move towards just always running the proper tests no matter the build tags, which would actually be possible now that the "behavioral test" logic is in libraries. For example:

* Run the code generator outputting to /tmp/somedir
* Generate an extra file with an Engine implementation, using the same PrototypeByName we generate today
* Have it call the right node/tests.SchemaTest func with the Engine (probably via a replace directive in a temporary module)
* Build a binary and run it, verifying it succeeds

So that would be akin to a plugin, but with a normal binary that either succeeds or errors in some way. About the same speed, and should work everywhere no matter cgo.

mvdan: Pretty sure this would let us get rid of the boilerplate in schema/gen/go, too. Happy to briefly chat about it later, and we could take a crack at it.

mvdan:
I'll merge without addressing the skip stuff right now, left you a detailed comment. Hope that's okay - I'd like to unblock myself :)

PR #187, opened 2021-06-07
node/tests: cover yet more interface methods

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -175,20 +175,35 @@ func prototype(goType reflect.Type, schemaType schema.Type) ipld.NodePrototype {
 	return &_prototype{schemaType: schemaType, goType: goType}
 }
 
+// Assert that we implement all the interfaces as expected.
+// Grouped by the interfaces to implement, roughly.
 var (
 	_ ipld.NodePrototype = (*_prototype)(nil)
+	_ TypedPrototype     = (*_prototype)(nil)
+	_ ipld.NodePrototype = (*_prototypeRepr)(nil)
 
 	_ ipld.Node        = (*_node)(nil)
 	_ schema.TypedNode = (*_node)(nil)
+	_ ipld.Node        = (*_nodeRepr)(nil)
 
 	_ ipld.NodeBuilder   = (*_builder)(nil)
 	_ ipld.NodeAssembler = (*_assembler)(nil)
+	_ ipld.NodeBuilder   = (*_builderRepr)(nil)
+	_ ipld.NodeAssembler = (*_assemblerRepr)(nil)
 
 	_ ipld.MapAssembler = (*_structAssembler)(nil)
 	_ ipld.MapIterator  = (*_structIterator)(nil)
+	_ ipld.MapAssembler = (*_structAssemblerRepr)(nil)
+	_ ipld.MapIterator  = (*_structIteratorRepr)(nil)
 
 	_ ipld.ListAssembler = (*_listAssembler)(nil)
 	_ ipld.ListIterator  = (*_listIterator)(nil)
+	_ ipld.ListAssembler = (*_listAssemblerRepr)(nil)
+
+	_ ipld.MapAssembler = (*_unionAssembler)(nil)
+	_ ipld.MapIterator  = (*_unionIterator)(nil)
+	_ ipld.MapAssembler = (*_unionAssemblerRepr)(nil)
+	_ ipld.MapIterator  = (*_unionIteratorRepr)(nil)
 )
 
 type _prototype struct {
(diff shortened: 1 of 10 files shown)


node/tests/string.go
```diff
@@ -0,0 +1,71 @@
+package tests
+
+import (
+	ipld "github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/node/mixins"
+)
+
+// A trimmed down copy from node/basic,
+// for the sake of test fixtures without import cycles.
+// TODO(mvdan): consider, instead, making node/basic tests be external
```
mvdan, line 10: I think this would work, but I also want to avoid going down rabbit holes before I can "publish" a v0.1 of basicnode.

node/bindnode/repr.go
```diff
@@ -473,16 +500,31 @@ func (w *_assemblerRepr) AssignNull() error {
 	}
 }
 
-func (w *_assemblerRepr) AssignBool(bool) error {
-	panic("TODO: AssignBool")
+func (w *_assemblerRepr) AssignBool(b bool) error {
+	switch stg := reprStrategy(w.schemaType).(type) {
+	case nil:
```
warpfork, line 505: I'm not sure I understand what `nil` is meaning in these blocks?

node/tests/string.go
warpfork: If this is something that can be fixed by painting a `_test` suffix onto some package declarations somewhere, yeah, go for it.

mvdan: it's a bit more than that, as I also have to rewrite a bunch of expressions to use the package qualifier. I'll give it a quick go.

node/bindnode/repr.go
mvdan: it means "no representation strategy". essentially, I want to panic if someone uses a method with a repr strategy I do not yet support.

one open question is what AssignBool should do on a known repr strategy where it doesn't make sense, like a struct with stringjoin. I guess we should return a "wrong kind" error then.

node/tests/string.go
mvdan: done

masih (approved):
Questions, otherwise LGTM 🍻

node/bindnode/node.go
```diff
@@ -379,7 +384,43 @@ func (w *_node) LookupByIndex(idx int64) (ipld.Node, error) {
...
+		idx, err := seg.Index()
+		if err != nil {
+			return nil, err
+		}
+		return w.LookupByIndex(idx)
+	}
+	return nil, ipld.ErrWrongKind{
+		TypeName:   w.schemaType.Name().String(),
+		MethodName: "LookupBySegment",
+		// TODO
```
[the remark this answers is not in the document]
mvdan: Just unimplemented stuff :) All the TODOs which don't have an explanation are too small for docs to be worth it, and I intend to get rid of them in the coming days. The TODOs which are trickier or will take more time to fix do have comments.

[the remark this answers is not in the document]
mvdan: (in this particular case, the errors lack some fields, because I wanted to get the PoC working as fast as possible)

node/bindnode/repr.go
warpfork: Yep, "wrong kind" error would be more correct.

node/tests/schemaStruct.go
```diff
@@ -101,16 +102,32 @@ func SchemaTestStructNesting(t *testing.T, engine Engine) {
...
 			Wish(t, n.Length(), ShouldEqual, int64(1))
+
 			n2 := must.Node(n.LookupByString("x"))
 			Require(t, n2.Kind(), ShouldEqual, ipld.Kind_Map)
+
+			n2Seg := must.Node(n.LookupBySegment(ipld.PathSegmentOfString("x")))
+			Wish(t, ipld.DeepEqual(n2, n2Seg), ShouldEqual, true)
+
+			n2Node := must.Node(n.LookupByNode(basicnode.NewString("x")))
+			Wish(t, ipld.DeepEqual(n2, n2Node), ShouldEqual, true)
```
warpfork, line 113: I'm not sure if this is a hair worth splitting or not, but... I think you could actually use pointer equality here for these last two equality checks (or native equality in general).  Thoughts?

warpfork (approved):

node/bindnode/repr.go
mvdan: ACK, I'll handle this in a future PR

node/tests/schemaStruct.go
mvdan: I would say not, because that still assumes one can use == on Nodes, which might not be true. It's an interface, after all.

PR #193, opened 2021-06-09
node/bindnode: rethink and better document APIs

mvdan:
(see commit message)


mvdan:
Tagging more people than usual for IPLD reflection work, since this is how I'm thinking the final API should look like. There are two runnable examples now, too. See the commit message as well.

The only major thing that's a TODO at this point is inferring a schema type from a Go type, but it won't mean any API changes.

node/bindnode/example_test.go
```diff
@@ -27,9 +27,44 @@ func ExamplePrototypeOnlySchema() {
...
+	type Person struct {
+		Name    string
+		Age     *int64 // optional
+		Friends []string
+	}
+	person := &Person{
+		Name:    "Michael",
+		Friends: []string{"Sarah", "Alex"},
+	}
+	node := bindnode.Wrap(person, schemaType)
```
mvdan, line 40: @willscott this is what I'm thinking in terms of what dealbot could have used.

mvdan: (or, of course, further slimmed down once inferring schemas is implemented)

willscott: that would have been very nice :)

node/bindnode/example_test.go
```diff
@@ -27,9 +27,44 @@ func ExamplePrototypeOnlySchema() {
...
+
+	nodeRepr := node.(schema.TypedNode).Representation()
```
willscott, line 42: why is the interface cast needed here?

mvdan: Ah, very good point, the bindnode APIs always work with a schema so they can return TypedNode/TypedPrototype directly.

mvdan: Done - type assertion here removed. I can't remove the type assert in the other example because going through ipld.NodeBuilder gets us an ipld.Node.

warpfork:
I like it.  The reduction in entry points, as described in the commit message, sounds very reasonable to me.

warpfork (approved):
I don't know if I want to rush to say we've reached API finalization without letting more folks weigh in (and maybe just plain some more burn in time with some alpha users)... but first impressions?  Looks (very) good to me.  And I continue to be :+1: to merging this stuff to master; it's just getting better every cycle.

node/bindnode/api.go
```diff
@@ -0,0 +1,95 @@
+// Package bindnode provides an ipld.Node implementation via Go reflection.
```
warpfork, line 1: Maybe minor wishlist: add more of the pointer to the two entrypoints (`Prototype` and `Wrap`) to the package scope docs.  Some of the prose in the commit message is already quite good for this.

warpfork: ... although on second thought, now that I look at how many exported symbols this package has (not many!), maybe this isn't very necessary.

mvdan:
> I don't know if I want to rush to say we've reached API finalization without letting more folks weigh in

I agree. Do you want me to mark the API as experimental? Then maybe around v0.11 or v0.12, if it has been successfully used by a few people, we can stabilize it.

node/bindnode/api.go
mvdan: Good call, will do.

warpfork:
Mostly saying that to disclaim how much I'm trying to grease the wheels without stepping on any other opinions :)

I think a quick note in the changelog file saying "this is new!  try it!  we may iterate, minorly" is probably appropriate.  I don't think we have any more rigourous flagging standards to follow.

node/bindnode/api.go
mvdan: Yeah, I was aiming at keeping the package godoc short, and hopefully make the (few) APIs pretty quick to navigate. I'll leave this out for now, and we can expand later if users find it confusing.

Issue #192, opened 2021-06-09
selectors vs selector dmt vs selector builders: confusing constructor proliferation


warpfork:
There are several different types in circulation when working with selectors in go-ipld-prime.  Some of this is essential complexity; some of it is not.

Of these types, they do very slightly different things.  Some can be converted to others, but the conversations are not bidirectional.  It's not clear which are meant to compose (and how) and which aren't.  The ease of use is lower than it should be as a result of all this.

First, let's enumerate these different types and talk about them a bit.  Afterwards, let's think about which of these we should emphasize or deemphasize when planning new APIs and writing documentation.

## Enumerating

- 1: the "executable" selector (`selector.Selector`)
- 2: the "dmt" (has no type, it's just `ipld.Node` that has to have a certain data convention inside)
- 3: the `builder.SelectorSpec`, which is a wrapper for # 2 and also has a builder method for # 1.

You'll instantly notice that # 2 isn't much of a type.  It's pretty much our IPLD equivalent of golang's `interface{}` -- it's practically a wildcard.  Nonetheless, this should count here, because [selector.Parse()](https://pkg.go.dev/github.com/ipld/go-ipld-prime/traversal/selector#ParseSelector)'s input parameter is... `ipld.Node`.

### Why did we end up with all these?

- `# 1` (the "executable" selector) exposes the functions needed while actually executing.
  - Some preprocessing happens while constructing these, which saves time when executing.
  - The act of constructing these and getting a handle to this type in golang means it's been validated!
  - These are relatively lightweight to allocate, compared to making a general-purpose `ipld.Node` tree out of `basicnode`, which is what the "dmt" form (# 2) defacto is.  This is important because Selectors sometimes return other Selectors during execution (e.g. recursion does this a lot).
    - _This might be an outdated reason._  Selectors were written before we had codegen, and *way* before we had the `bindnode` option available.  We might be able to revisit this now.
    
- `# 2` (the "dmt", the declarative Data Model Tree defining the selector) exists... because it has to.
  - This is our "thin waist".  This is how Selectors are actually defined, specified, serialized, and communicated.
    - The first step of getting a selector ready to go is you write it in JSON (or any other format there's an IPLD Codec for!), then parse that into Data Mode (aka `ipld.Node`), and then... you're here; now you call `selector.Parse(node)` with it. 
    - ... or, you use # 3, below, but that does indeed still bounce through here, if briefly.
[...]

PR #194, opened 2021-06-09
Proposal: a better selector builder based on QP

hannahhoward:
# Goals

This is a draft a more ipld native selector builder based on QP. I don't have time to finish it. However, I think it would work quite well and solve a number of the challenges and allow us to deprecate the builders.

# Implementation

as an example, `foo/bar/**` could be built with:

```golang
BuildExploreFields(basicnode.Prototype.Map, func(ma ipld.MapAssembler) {
   qp.MapEntry(ma, "foo", ExploreFields(func(ma ipld.MapAssembler) {
      qp.MapEntry(ma, "bar", ExploreRecursive(selector.RecursionLimitNode, ExploreAll(ExploreRecursiveEdge())))
   })
})
```

I have already spent my max amount of time on this so someone will have to pick it up if we want to implement it.

For @ribasushi 's example:
```golang
func SelectorSpecFromPath(path Expression, optionalSubselectorAtTarget ipld.Node) (ipld.Node, error) {

	if path == "/" {
		return nil, fmt.Errorf("a standalone '/' is not a valid path")
	} else if m := invalidChar.FindStringIndex(string(path)); m != nil {
		return nil, fmt.Errorf("path string contains invalid character at offset %d", m[0])
	}

        var ss qp.Assemble
	if optionalSubselectorAtTarget == nil {
	    ss = Matcher()
	} else {
	    ss = qp.Node(optionalSubselectorAtTarget)
	}

	segments := strings.Split(string(path), "/")

	// walk backwards wrapping the original selector recursively
	for i := len(segments) - 1; i >= 0; i-- {
		if segments[i] == "" {
			// allow one leading and one trailing '/' at most
			if i == 0 || i == len(segments)-1 {
				continue
			}
			return nil, fmt.Errorf("invalid empty segment at position %d", i)
		}

		if segments[i] == "." || segments[i] == ".." {
			return nil, fmt.Errorf("unsupported path segment '%s' at position %d", segments[i], i)
		}

		if onlyDigits.MatchString(segments[i]) {
			if segments[i][0] == '0' && len(segments[i]) > 1 {
				return nil, fmt.Errorf("invalid segment '%s' at position %d", segments[i], i)
			}

			idx, err := strconv.ParseInt(segments[i], 10, 31)
			if err != nil {
				return nil, fmt.Errorf("invalid index '%s' at position %d: %s", segments[i], i, err)
			}

			ss = ExploreIndex(
				idx,
				ss,
			)
		} else {
			ss = ExploreFields(func(ma ipld.MapAssembler) {
				qp.MapEntry(ma, segments[i], ss)
			})
		}
	}

       nb := basicnode.Prototype.Any.NewBuilder()
       ss(nb)
       return nb.Build()
}
```

### DIFF
--- a/traversal/selector/qp/qp.go
+++ b/traversal/selector/qp/qp.go
@@ -1,0 +1,142 @@
+package qp
+
+import (
+	ipld "github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/fluent/qp"
+	selector "github.com/ipld/go-ipld-prime/traversal/selector"
+)
+
+func exploreRecursiveEdge(ma ipld.MapAssembler) {
+	qp.MapEntry(ma, selector.SelectorKey_ExploreRecursiveEdge, qp.Map(0, func(ipld.MapAssembler) {}))
+}
+
+func ExploreRecusiveEdge(pt ipld.NodePrototype) (ipld.Node, error) {
+	return qp.BuildMap(pt, 1, exploreRecursiveEdge)
+}
+
+func ExploreRecursiveEdge() qp.Assemble {
+	return qp.Map(1, exploreRecursiveEdge)
+}
+
+func exploreRecursive(limit selector.RecursionLimit, sequence qp.Assemble) func(ipld.MapAssembler) {
+	return func(ma ipld.MapAssembler) {
+		qp.MapEntry(ma, selector.SelectorKey_ExploreRecursive, qp.Map(2, func(ma ipld.MapAssembler) {
+			qp.MapEntry(ma, selector.SelectorKey_Limit, qp.Map(1, func(ma ipld.MapAssembler) {
+				switch limit.Mode() {
+				case selector.RecursionLimit_Depth:
+					qp.MapEntry(ma, selector.SelectorKey_LimitDepth, qp.Int(limit.Depth()))
+				case selector.RecursionLimit_None:
+					qp.MapEntry(ma, selector.SelectorKey_LimitNone, qp.Map(0, func(na ipld.MapAssembler) {}))
+				default:
+					panic("Unsupported recursion limit type")
+				}
+			}))
+			qp.MapEntry(ma, selector.SelectorKey_Sequence, sequence)
+		}))
+	}
+}
+
+func BuildExploreRecursive(np ipld.NodePrototype, limit selector.RecursionLimit, sequence qp.Assemble) (ipld.Node, error) {
+	return qp.BuildMap(np, 1, exploreRecursive(limit, sequence))
+}
+
+func ExploreRecursive(limit selector.RecursionLimit, sequence qp.Assemble) qp.Assemble {
+	return qp.Map(1, exploreRecursive(limit, sequence))
+}
+
+func exploreAll(next qp.Assemble) func(ipld.MapAssembler) {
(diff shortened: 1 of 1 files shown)


mvdan:
Just one request before this is considered for merging: I wouldn't call the package qp, because it would be pretty confusing to have two very related packages in the same module with the same name. Maybe that means we try merging the packages, or coming up with some other name.

traversal/selector/qp/qp.go
```diff
@@ -0,0 +1,142 @@
...
+	ipld "github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/fluent/qp"
+	selector "github.com/ipld/go-ipld-prime/traversal/selector"
+)
+
+func exploreRecursiveEdge(ma ipld.MapAssembler) {
+	qp.MapEntry(ma, selector.SelectorKey_ExploreRecursiveEdge, qp.Map(0, func(ipld.MapAssembler) {}))
+}
+
+func ExploreRecusiveEdge(pt ipld.NodePrototype) (ipld.Node, error) {
```
warpfork, line 13: Typo in the function name.  And has collision when typo fixed.

warpfork:
This is really cool.

The example usage reads and composes really nicely.

(It's also still a golang DSL, so, consistent with concerns I expressed in https://github.com/ipld/go-ipld-prime/issues/192#issuecomment-857917384 about how useful it is to spend time on _any_ language's DSLs, I'm ambivalent about how much time to spend on this.  But let's roll with it for a moment anyway and see what can be learned.  Especially, if we can detect some patterns here for doing this generally in golang, that seems provocative.)

I'm getting increasingly fond of this kind of pattern where some constructor-like functions are returning functions that are applied on NodeAssembler.  The concept is a mouthful but it composes really well.

Something that's especially interesting about this pattern is how it adds guardrails and syntactic nicety... without attaching it to the Node implementation type.  Which means it can also be applied to `basicnode`, etc.  Now, maybe that seems unremarkable, because that was kinda table-stakes here, because we never did codegen'd selector nodes.  But it's also really interesting to consider in a broader view: one could intentionally use this pattern as part of doing well-structured operations on semi-structured data (ex: some data that _mostly_ conforms to a schema, but also has extra fields tagging along, which means we don't want to shoehorn it into generated structures that would discard it).  Conventions for this could be really useful.

Seems like they'll be some naming conventions needed for keeping things from getting ad-hoc as more variations of constructors are added.  (A standard golang problem, I guess.)  Something like `{Outcome}Of{ConstructionStyle}` maybe?

I wonder if it would be possible to pursue more specific return types for a lot of these functions.  `qp.Assemble` is what most of them return.  That means it's fairly easy to nest some constructions that aren't actually sensible, and get no remark from the compiler at compile-time.  (This might be mild for this scenario with selectors, because most of the selector clauses _do_ nest freely.  But it's interesting if trying to generalize this pattern.)  Thinking outloud on this at this point.  There'd be tradeoffs here: it _is_ totally logically fine to want to drop out of this particular constructor stacking, and apply a general `qp.Assemble` deep in a tree; making that illegal gives more guardrails against what is often going to be errors, but is also outlawing something that's not necessarily wrong.

I'm not sure what direction we'll want to carry this, or how to prioritize it right now, but this is definitely interesting.

Issue #191, opened 2021-06-08
absent-vs-null caltrops


warpfork:
We have some errata about how `Absent` vs `Null` are handled.  This is a distinction that is sometimes useful, but also very subtle.

(`Absent` is somewhat similar to javascript's `undefined`.  I hope we pull it off better than javascript did, frankly -- but javascript's history with this concept is certainly a valid cautionary tale about the degree of subtlety, too.)

At least one piece of code that touches on these corner cases resulted in pain today: a codec coerced an `Absent` to a `Null` and serialized data that way.  The resulting serial data could then not be parsed back: because the schema declared "optional" rather than "nullable" for the affected fields, a null token was (correctly!) rejected in that position as data that was not matching the schema.

It would be better to not be able to end up in this position (or, at minimum, make it much harder to end up in this position accidentally).

I have three sets of ideas, in increasing levels of invasiveness, which could be relevant:

---

The first change is the most direct.

- Codecs should absolutely *not* silently coerce an `ipld.Absent` node to be encoded as a `null` token.
	- In the past: I thought this might be "helpful", because it would mean you could use e.g. the dag-json codec as a "debug" view on the typed data (even if you have a low-readability representation strategy somewhere in the schema).
	- In the present: It is now clear that this footgun is *way* bigger and more painful than the hypothetical upside.  So let's cut this out.
	- What it should do: is **error** immediately upon detecting an `Absent`.  Erroring is much better than silent coersion of values.

We can make a norm of having codec implementations do this check, and flunk that data.

(Unfortunately, we can only make it a strongly recommended feature for codec implementations to do this check.  I can't think of a way to make the type system help us enforce this rule for any third party codecs.  But maybe that's fine.)

(It's also still concerningly easy for a codec to forget to check this, because absent nodes report their Kind as `ipld.Kind_Null`, and you have to check `IsAbsent` or `== ipld.Absent` to disambiguate.  But again, maybe that's fine (for now), and we just try to make a norm of it anyway.)

This will detect this particular situation of accidental coersion of absent values into nulls.  It should stop people from reaching a situation where they have serial data which they believe is correct but was actually serialized from a level they didn't intend.  That's the most important goal.  (However, also remember that this only kicks in if there's `Absent` values in the data -- it's not a *general* detector for accidentally serializing type-level data when one meant to serialize representation-level data... so, depending on where you're putting your goalposts, this may be seen as either a full solution to the specific problem, or a "partial" solution to a bigger problem.)

[...]

PR #196, opened 2021-06-15
codecs: error upon encountering an 'absent' value during encodes.

warpfork:
See the changelog entry in the diff for some more explanation.

This is one of the low-hanging-fruit/low-risk steps to address the worst of the issues outlined in https://github.com/ipld/go-ipld-prime/issues/191 .  It should fix the most dire of scenarios where accidental misuse of an API could cause lossy data coersions silently.

### DIFF
--- a/codec/dagjson/marshal.go
+++ b/codec/dagjson/marshal.go
@@ -19,8 +19,11 @@ func Marshal(n ipld.Node, sink shared.TokenSink, allowLinks bool) error {
 	var tk tok.Token
 	switch n.Kind() {
 	case ipld.Kind_Invalid:
-		return fmt.Errorf("cannot traverse a node that is absent")
+		return fmt.Errorf("invalid node encountered")
 	case ipld.Kind_Null:
+		if n.IsAbsent() {
+			return fmt.Errorf("cannot encode a node that is absent")
+		}
 		tk.Type = tok.TNull
 		_, err := sink.Step(&tk)
 		return err
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,10 @@ Unreleased on master
 Changes here are on the master branch, but not in any tagged release yet.
 When a release tag is made, this block of bullet points will just slide down to the [Released Changes](#released-changes) section.
 
-- _nothing yet :)_
+- Changed: encoding functions for codecs in this repo (dag-cbor, dag-json, cbor, json) will now **error** if they encounter an `ipld.Absent` node (or more specifically, if `IsAbsent() == true`).
+  [[#196](https://github.com/ipld/go-ipld-prime/pull/196)]
+	- Previously, these codecs would quietly emit a serial "null" token in this scenario -- which was effectively implicitly coercing a value into null.  Implicit coercion creates bugs and is bad, so we will not be doing it anymore.
+	- This situation is only reachable when working with schema-typed data (since otherwise `Absent` values aren't used).  In that situation, handing that typed value to an encode function without having called `.Representation()` first is also typically a logical error, so, returning an explicit error now is also probably the right thing to do.
 
 
 
--- a/codec/dagcbor/marshal.go
+++ b/codec/dagcbor/marshal.go
@@ -22,8 +22,11 @@ func Marshal(n ipld.Node, sink shared.TokenSink, allowLinks bool) error {
 func marshal(n ipld.Node, tk *tok.Token, sink shared.TokenSink, allowLinks bool) error {
 	switch n.Kind() {
 	case ipld.Kind_Invalid:
-		return fmt.Errorf("cannot traverse a node that is absent")
+		return fmt.Errorf("invalid node encountered")
 	case ipld.Kind_Null:
+		if n.IsAbsent() {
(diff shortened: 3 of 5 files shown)


mvdan (approved):
I like this. It would have definitely caught our misuse in dealbot, even though it won't catch all misuses of non-repr nodes.

Out of curiosity, you mentioned that one might want to encode a non-repr node for the sake of debugging the non-repr node structure. Is that now kinda abandoned, given how we'd just error on absent fields etc?

mvdan:
Also, I just noticed after approving, but... Add a test?

warpfork:
> I like this. It would have definitely caught our misuse in dealbot, even though it won't catch all misuses of non-repr nodes.

Correct on both counts.

> you mentioned that one might want to encode a non-repr node for the sake of debugging the non-repr node structure. Is that now kinda abandoned, given how we'd just error on absent fields etc?

Yes.  I give up on / walk back on the idea of that "debug" story as a high priority goal.  It was a bad idea.  This coersion behavior has proven to be infinitely more dangerous and footgunny than the hypothetical utility of the debug behavior.

I'm still a bit sad to give up the idea of that debug story, but ditching it is looking like by far the lesser of two evils.

One can still _kind of_, even with this change, feed non-repr nodes to encoding, and get that sort of "debug" story... as long as they don't contain optional fields that are absent.  But that's sufficiently likely to be a show-stopper for that story to be useful in practice that it's not really worth much; it's more just a coincidence, now.

codec/dagjson/marshal.go
```diff
@@ -19,8 +19,11 @@ func Marshal(n ipld.Node, sink shared.TokenSink, allowLinks bool) error {
 	var tk tok.Token
 	switch n.Kind() {
 	case ipld.Kind_Invalid:
-		return fmt.Errorf("cannot traverse a node that is absent")
+		return fmt.Errorf("invalid node encountered")
 	case ipld.Kind_Null:
+		if n.IsAbsent() {
+			return fmt.Errorf("cannot encode a node that is absent")
```
willscott, line 25: i worry that this error message, which will be at runtime, will be fairly opaque. Consider adding a hint like "perhaps you are trying to encode the non-representation version of a node?"

willscott (commented):
I'm unconvinced that the answer is to return an error in the common case when we generally know what the end user wants to do. If i am marshaling an instance of a node, having it fail in a data-dependent way because an optional field happens to be absent seems like it's still huge foot gun. My expectation as someone asking to marshal that node is that i get the marshaling of what we call the representation version.
I've become pretty strongly opinionated that we should make that behavior the default because it's what a user is going to expect.

mvdan:
> If i am marshaling an instance of a node, having it fail in a data-dependent way because an optional field happens to be absent seems like it's still huge foot gun. My expectation as someone asking to marshal that node is that i get the marshaling of what we call the representation version.

I actually came to the same conclusion this morning, after I was thinking some more about my reply. If encoding a non-repr node is almost never what you want, and it can fail depending on the presence of an absent, and we're giving up on using codecs as debug helpers... Why not just make the codecs consistently grab a node's representation by default?

Here's how I propose to do it: We strongly recommend that any codec, when encoding a node or decoding into a prototype, should check if it's typed. If it is, it should use its representation.

In the very rare case that someone *does* want to encode or decode the non-repr version, we can offer Prototype and Node wrappers that simply hide the top-level `Representation` method. This should be as simple as:

```
type NodeHidingRepr struct {
    ipld.Node
}

node := someTypedNode(...)

dagjson.Encode(node) // uses the representation, as per the default

dagjson.Encode(NodeHidingRepr{node}) // doesn't see a TypedNode, uses the non-repr view
```

rvagg:
putting this on my list to review and possibly close or merge (tbh I don't even recall this discussion! rough 2021)

rvagg:
TODO: does https://github.com/ipld/go-ipld-prime/issues/246 get covered by this?

rvagg:
Sooooo we've had #232 since this was opened, which kind of addresses this, although we haven't been consistent in using that API (at least I haven't! tbh I missed #232 and keep on defaulting to direct codec usage).

Looking back over #191 that this comes from (as option 1) and considering #232 got added, I don't really buy the case for this, it's exchanging hostility for a notion of correctness; at least it's saying "no, because I'm seeing this thing, you're _probably_ using the API wrong, so have an error", with no room for bargaining.

I'm going to take executive action on this one and close it, for now, and open an issue to adjust documentation throughout go-ipld-prime to recommend usage of the top-level `Encode()`, `Decode()` and their `*Streaming()` partners. Maybe also hunting down some usage across our repos and replace them there to help point people in the right direction instead of copy/pasting our bad usage. I'm pretty guilty of this and can think of many instances of this in the last couple of months.

Dan's [suggestion](https://github.com/ipld/go-ipld-prime/pull/196#issuecomment-862383955) of how to encode the representation form even works for those new APIs so we could potentially document that too.. although as we've said in here and other places, the case for encoding the typed (or ADL) form is pretty minimal.

Of course, if Eric or someone else wants to come back and re-litigate the case for this then you're welcome to.

PR #197, opened 2021-06-18
fluent/quip: remove in favor of qp

mvdan:
(see commit message)


warpfork:
Yep.  I give up on this package.

Some more comments on lessons learned, for any future review on this, or for future attempts on these fluent APIs:

- The strategy of error handling `quip` attempted was intended to score high on explicitness.  It's an interesting idea.  But:
    - It failed to see that explicitness translate into reliably better error handling or better DX.
    - The odds of making mistakes in the error gathering, which is left largely up to you as the user, is just way way too great.
        - My own attempts to use it have run into this so frequently that there's just... no.  No way.
- The outcomes when you do have an error gathering mistake are just too terrible: you'll get "misuse" panics from the Node implementation's guts, but at an arbitrary distance from the problem code.  The result is utterly horrible to debug.
- The composition strategy when starting to use your own functions for repeated pieces of data wrangling is just too odd.
    - The least-weird way to go about this probably involved repeating the pattern of an error pointer parameter in your own functions.  Which is very unusual in golang, to put it mildly.  I didn't expect this to be contagious.  That it turned out to be is problematic.

warpfork:
Some other tangentially related lessons learned during my day trying to work with this package (not strictly related, but I want to note before I forget): it would be nice to have a clone method on assemblers, or other way to inspect partially built state, so you can stop and take stock of the situation.  This is currently kind of hard to do, because that API was driven by efficiency concerns and by trying to shield incomplete/not-yet-validated data from being read, but lacking this during debugging is painful.

warpfork (approved):

PR #199, opened 2021-07-03
selectors: docs enhancements, new construction helpers.

warpfork:
(This PR is probably a discussion started first and foremost, but might actually also just be minimally useful.)

Tried to enhance the Selector type docs a bit more.  This is feeling easier to talk about having looked at other systems we've built in IPLD and coming back to it now: there's a couple recurring themes: we've got a "compiled" type and a data model tree ("dmt") type again.  (In the case of selectors, the latter still doesn't really have a type in the golang signatures here -- it's just `Node`, which is fairly unspecific -- but someday we could make this more strongly typed, and probably would like to.)

I added a `CompileSelector` function, which is exactly the same as the old `ParseSelector`.  I think I'd like to push this naming convention, though.  "parse" (at least in my mind) connotes something that has much more to do with codecs than this function's behavior actually does -- it's no codec, it's just dmt->compiled.  I didn't get rid of `ParseSelector`, because I think slow change rollout is becoming valuable, but I commented it as deprecated, and consider this a "move" operation.

I added a `CompileJSONSelector` function as well.  This function does more work in one step: you give it a json string, it parses json->dmt, and then it compiles dmt->compiled selector.  Should be a nice usability enhancement, but there's a couple things about it that are bikeshedable, and a couple things that are debatable -- I'll start comment threads on the diff for those.

Finally, started adding some constants of selectors that are "common", as @willscott has been advocating for for a while now :)  These are using JSON string literals as their constructors, which I think will make them useful for educational purposes -- folks looking at them can see this clearly serializable thing that they can copy, edit, and paste, even paste into APIs (rather than being limited to use in golang code)... I think that should be worth about a thousand words of other docs.


traversal/selector/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/ipld/go-ipld-prime/codec/json"
+	"github.com/ipld/go-ipld-prime/node/basic"
+)
+
+// REVIEW: I feel sketchy about CompileJSONSelector because it brings in json as a transitive dependency.
+// This has consequences as far as dragging go-cid in because our json shares code with dagjson...
+// and then go-cid ends up in the transitives of the traversal package, since traversal depends on selector.
+// That's... not great.
+// On the other hand: trying to keep up that paper wall is creating a lot of strife.  Maybe it's time to give up.
```
warpfork, line 14: Big question here about what we let this do to our dependency tree.

I think the usability argument probably needs to win the match here, but it's a pretty major boundary relaxation, so flagging it for some discussion would make me more comfortable with it.

traversal/selector/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+
+func CompileJSONSelector(jsonStr string) (Selector, error) {
```
warpfork, line 16: I think there'd be a strong argument for calling this `ParseJSONAndCompileSelector`, but of course that's a mouthful.

traversal/selector/helpers.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func must(s Selector, e error) Selector {
+	if e != nil {
+		panic(e)
+	}
+	return s
+}
+
+var CommonSelector_MatchPoint = must(CompileJSONSelector(`{".":{}}`))
```
warpfork, line 35: We could of course add many more of these :)  I think maybe 3 or 4 is probably a good number to aim for -- there's no need to put a kitchen sink here, and the more we do, the more we're gonna get in danger of having to bikeshed their variable names, but several distinctive examples would be good.

traversal/selector/helpers.go
willscott: if you want to put the helper functions in a separate folder/module to avoid it, that seems okay?

traversal/selector/helpers.go
willscott: comments explaining them would also be good :)

willscott (approved):

traversal/selector/helpers.go
willscott: the other ones i'd like to see are: "here and direct children" and "full dag"

traversal/selector/helpers.go
rvagg: They're pretty entangled, not much room for extracting a shared helper. An alternative way might be to switch the dependency so dagjson is an extension of json that adds bytes and cid.

.. or we address the go-cid dependency issues.

warpfork: @willscott has just brought up elsewhere that if we add features like "list of CIDs to stop at" into Selectors, then any codec that can parse that is going to have a go-cid dependency anyway.

So, a separate package here is still an option, and address go-cid's dependency sprawl is still a long term wish, but "fix json to not depend on go-cid" now seems irrelevant because it looks like selectors may end up using dag-json anyway.

warpfork:
I've pushed this around a bit to a state I think I'm happy with.  Contains much more docs now.

I did move things to a sub-package so we can also merge this while remaining fairly noncommittal to it, and feel relatively free to iterate for a bit.

I think I'd like to merge this now?  Even if there's nothing earthshattering here, incremental improvement to the docs would be nice to have on master now.  (I almost accidentally started writing docs improvements again before realizing there are some on this branch... :facepalm: )

warpfork:
"no complaints" is good enough for me on this one.

PR #202, opened 2021-07-14
Remove pretty-printing

rvagg:
but .. this is only part of the story, because we still have map key ordering "problems"

### DIFF
--- a/codec/dagjson/multicodec.go
+++ b/codec/dagjson/multicodec.go
@@ -54,7 +54,7 @@ func Encode(n ipld.Node, w io.Writer) error {
 	//  (There's not really any fastpaths of note for json.)
 	// Write another function if you need to tune encoding options about whitespace.
 	return Marshal(n, json.NewEncoder(w, json.EncodeOptions{
-		Line:   []byte{'\n'},
-		Indent: []byte{'\t'},
+		Line:   nil,
+		Indent: nil,
 	}), true)
 }
--- a/codec/dagjson/multicodec.go
+++ b/codec/dagjson/multicodec.go
@@ -53,8 +53,5 @@ func Encode(n ipld.Node, w io.Writer) error {
 	// Shell out directly to generic inspection path.
 	//  (There's not really any fastpaths of note for json.)
 	// Write another function if you need to tune encoding options about whitespace.
-	return Marshal(n, json.NewEncoder(w, json.EncodeOptions{
-		Line:   nil,
-		Indent: nil,
-	}), true)
+	return Marshal(n, json.NewEncoder(w, json.EncodeOptions{}), true)
 }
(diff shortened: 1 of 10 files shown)


traversal/focus_test.go
```diff
@@ -318,7 +318,7 @@ func TestFocusedTransformWithLinks(t *testing.T) {
 			Wish(t, progress.Path.String(), ShouldEqual, "linkedMap/nested/nonlink")
 			Wish(t, must.String(prev), ShouldEqual, "zoo")
 			Wish(t, progress.LastBlock.Path.String(), ShouldEqual, "linkedMap")
-			Wish(t, progress.LastBlock.Link.String(), ShouldEqual, "baguqeeyevmbz3ga")
+			Wish(t, progress.LastBlock.Link.String(), ShouldEqual, "baguqeeye2opztzy")
```
rvagg, line 321: I'm pretty uncomfortable about changing this without verifying the block encodes to this CID, but the fixture is complicated enough that I'm not keen to reconstruct it for a one-off verification unless this arises out of some previously constructed fixture.

codec/dagjson/multicodec.go
```diff
@@ -54,7 +54,7 @@ func Encode(n ipld.Node, w io.Writer) error {
 	//  (There's not really any fastpaths of note for json.)
 	// Write another function if you need to tune encoding options about whitespace.
 	return Marshal(n, json.NewEncoder(w, json.EncodeOptions{
-		Line:   []byte{'\n'},
-		Indent: []byte{'\t'},
+		Line:   nil,
+		Indent: nil,
```
mvdan, line 58: FYI you can simply remove these, as the zero value is the default.

mvdan (approved):
Nice! I've wanted this for some time.

You could also simplify the bindnode examples, which explicitly set refmt parameters to disable pretty-printing: https://pkg.go.dev/github.com/ipld/go-ipld-prime@master/node/bindnode#example-Prototype-OnlySchema

I think those could now use the default behavior.

warpfork (approved):

rvagg:
@mvdan can you have a look at 89258dd please as I've simplified a few places where refmt was previously used. It's not clear why it was used in a couple of those places and maybe I've overstepped by undoing those too.

mvdan:
@rvagg LGTM, looks amazing :love_letter:

mvdan:
> It's not clear why it was used in a couple of those places and maybe I've overstepped by undoing those too.

As far as I can tell, using refmt explicitly is only to supply non-default options, which only matters for the encoding pretty-printing. So I'm pretty sure it's a no-op to remove empty refmt options from decodes. And for the encodes you're changing the default to not pretty-print, so that also looks like a no-op change to me.

warpfork:
:+1: to seeing less of refmt -- I'm wholeheartedly in favor of change in that direction.  Most (all?) of the places it appears were punts where I didn't have a better plan at the time.

If we do get into working on configurability patterns for ipld codecs, I'd like to not have any refmt symbols appear in them.  There's just no reason to have codec configurability cause the user's attention to get stretched out across repos if we can help it.

PR #204, opened 2021-07-16
Sort map entries marshalling dag-cbor

rvagg:


mvdan:
Nothing looks wrong here, though I think we should talk to Eric about codec map key sorting first, using the dag-json PR as a specific example. If and when we agree to merge that, then merging this should be easy :)

warpfork:
I think we probably discussed things pretty sufficiently in https://github.com/ipld/go-ipld-prime/pull/203 now --

Same one major change request before merge here: the `cbor` codec package actually (for better or for worse; worse, probably, for dependency sprawl, but nonetheless, for code DRYness) depends on the `dagcbor` package, so we should put a bool through here, so the `cbor` codec can continue to not sort.

rvagg:
If UnmarshalOptions and MarshalOptions are acceptable in #203 then I'll copy the approach here. (It'd also be possible to overload this even further to do more interesting things like having custom sorters, or "don't sort typednodes". But I'll restrain on feature creep.

rvagg:
k, I've added a `MarshalOptions` and `UnmarshalOptions` here with both control for links (encode and decode) and a `MapSortMode` for encode that lets you switch between sorting modes. Added a simple test for codec/cbor to exercise it like the others and it demonstrates lack of sorting _(mind you ... there's zero definition of what a "cbor" IPLD codec should be encoding, it has some up before with users wanting to throw all sorts of tags into plain "cbor" blocks, we have no rules at all other than maybe a best-attempt at decoding what you're given)._

PR #205, opened 2021-07-21
Make `go vet` happy

rvagg:
@mvdan @warpfork I know you've had discussions about linting and whatnot and there's strongly held opinions about such things, and I'm wading into this without recollection of your discussions about the path forward (if any) here. It seemed to me that this was low-hanging fruit on the path toward a possible inclusion in the https://github.com/protocol/.github list. There's one minor codegen change in here which is going to add a few more lines to generated code.
Feel free to dismiss this! Not a big deal, just had some muscle memory doing this in a couple of other repos so I thought'd I'd try my hand while I was here.

### DIFF
--- a/adl/rot13adl/example_test.go
+++ b/adl/rot13adl/example_test.go
@@ -16,7 +16,7 @@ import (
 	"github.com/ipld/go-ipld-prime/storage"
 )
 
-func ExampleUnmarshallingToADL() {
+func ExampleReify_unmarshallingToADL() {
 	// Create a NodeBuilder for the ADL's substrate.
 	//  Unmarshalling into this memory structure is optimal,
 	//   because it immediately puts data into the right memory layout for the ADL code to work on,
@@ -42,7 +42,7 @@ func ExampleUnmarshallingToADL() {
 	// adl node kind: string
 	// adl view value: "a cool string"
 }
-func ExampleLoadingToADL() {
+func ExampleReify_loadingToADL() {
 
 	// Create a NodeBuilder for the ADL's substrate.
 	//  Unmarshalling into this memory structure is optimal,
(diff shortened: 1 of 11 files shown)


mvdan:
I'm certainly in favor of making the standard checks happy :) I did put in most of the legwork, especially when it came to removing vet warnings in generated code. I didn't get it over the finish line as I just lost interest with the pushback: https://github.com/ipld/ipld/pull/92#discussion_r561835634

I do think it would be unfortunate to have the same standard checks across PL, but not for one of the "flagship core" Go repos, so I think there's value in finishing this. I even helped file [a proposal](https://github.com/golang/go/issues/43864) to make the check more lenient, in an attempt to reach some middle ground.

rvagg:
I've done more than my share of style-rage in JavaScript through the past so understand how rusted-on opinions can get when it comes to personal style.

There are a few fixes in here that I assume would unambiguously good, such as. the dead code. Having codegen be cleaner for users who are going to `go vet` their own code is probably a worthy goal too but also bumps up against the imposition of personal style.

It turns out that building a `fmt` into a language still doesn't solve these personal style things! There has to be just a _single way_ to avoid that I guess!

mvdan:
> Having codegen be cleaner for users who are going to `go vet` their own code is probably a worthy goal too but also bumps up against the imposition of personal style.

FYI Eric and I already agreed on making the generator vet-happy, to unblock the "unified CI" standard checks for other repos. I simply forgot to fix one or two remaining vet errors, because they only triggered in some schema edge cases.

warpfork (approved):
LGTM.

(I still don't care for the struct-initializers-always-explicit-field-names thing, but I've ranted about that before elsewhere, and am giving up on it as not really worth the debate.)

warpfork:
I don't reckon this will conflict with anything else open or has been referenced by commit hash anywhere, so I'll just go ahead and rebase+merge this.

PR #207, opened 2021-07-22
node/bindnode: first pass at inferring IPLD schemas

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/infer.go
+++ b/node/bindnode/infer.go
@@ -5,7 +5,24 @@ import (
 	"reflect"
 	"strings"
 
+	ipld "github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/schema"
+)
+
+var (
+	goTypeBool   = reflect.TypeOf(false)
+	goTypeInt    = reflect.TypeOf(int(0))
+	goTypeFloat  = reflect.TypeOf(0.0)
+	goTypeString = reflect.TypeOf("")
+	goTypeBytes  = reflect.TypeOf([]byte{})
+	goTypeLink   = reflect.TypeOf((*ipld.Link)(nil)).Elem()
+
+	schemaTypeBool   = schema.SpawnBool("Bool")
+	schemaTypeInt    = schema.SpawnInt("Int")
+	schemaTypeFloat  = schema.SpawnFloat("Float")
+	schemaTypeString = schema.SpawnString("String")
+	schemaTypeBytes  = schema.SpawnBytes("Bytes")
+	schemaTypeLink   = schema.SpawnLink("Link")
 )
 
 // Consider exposing these APIs later, if they might be useful.
@@ -24,21 +41,21 @@ func inferGoType(typ schema.Type) reflect.Type {
 		return goTypeBytes
 	case *schema.TypeStruct:
 		fields := typ.Fields()
-		goFields := make([]reflect.StructField, len(fields))
+		fieldsGo := make([]reflect.StructField, len(fields))
 		for i, field := range fields {
-			ftyp := inferGoType(field.Type())
+			ftypGo := inferGoType(field.Type())
 			if field.IsNullable() {
-				ftyp = reflect.PtrTo(ftyp)
+				ftypGo = reflect.PtrTo(ftypGo)
 			}
 			if field.IsOptional() {
-				ftyp = reflect.PtrTo(ftyp)
+				ftypGo = reflect.PtrTo(ftypGo)
 			}
-			goFields[i] = reflect.StructField{
+			fieldsGo[i] = reflect.StructField{
 				Name: fieldNameFromSchema(field.Name()),
-				Type: ftyp,
+				Type: ftypGo,
 			}
 		}
-		return reflect.StructOf(goFields)
+		return reflect.StructOf(fieldsGo)
 	case *schema.TypeMap:
 		ktyp := inferGoType(typ.KeyType())
 		vtyp := inferGoType(typ.ValueType())
@@ -54,7 +71,7 @@ func inferGoType(typ schema.Type) reflect.Type {
 		//		Keys   []K
 		//		Values map[K]V
 		//	}
-		goFields := []reflect.StructField{
+		fieldsGo := []reflect.StructField{
 			{
 				Name: "Keys",
 				Type: reflect.SliceOf(ktyp),
@@ -64,7 +81,7 @@ func inferGoType(typ schema.Type) reflect.Type {
 				Type: reflect.MapOf(ktyp, vtyp),
 			},
 		}
-		return reflect.StructOf(goFields)
+		return reflect.StructOf(fieldsGo)
 	case *schema.TypeList:
 		etyp := inferGoType(typ.ValueType())
 		if typ.ValueIsNullable() {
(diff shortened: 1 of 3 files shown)


mvdan:
@warpfork please pay extra attention to the TODOs :)

node/bindnode/infer.go
```diff
@@ -87,6 +104,78 @@ func fieldNameFromSchema(name string) string {
...
+		etyp := typ.Elem()
+		nullable := false
+		if etyp.Kind() == reflect.Ptr {
+			etyp = etyp.Elem()
+			nullable = true
+		}
+		etypSchema := inferSchema(typ.Elem())
+		name := schema.TypeName(typ.Name())
+		if name == "" {
+			name = "List_" + etypSchema.Name()
```
mvdan, line 174: It's going to be common to want to do `SomeField []SomeType`, so I think falling back to `List_SomeType` is reasonable. The alternative is to force the user to write `type List_SomeType []SomeType` on the Go side, but that's going to get pretty annoying.

node/bindnode/infer.go
```diff
@@ -87,6 +104,78 @@ func fieldNameFromSchema(name string) string {
...
+	defaultTypeSystem.Accumulate(schemaTypeInt)
+	defaultTypeSystem.Accumulate(schemaTypeFloat)
+	defaultTypeSystem.Accumulate(schemaTypeString)
+	defaultTypeSystem.Accumulate(schemaTypeBytes)
+	defaultTypeSystem.Accumulate(schemaTypeLink)
+}
+
+// TODO: support IPLD maps and unions in inferSchema
+
+// TODO: support bringing your own TypeSystem?
```
willscott, line 122: this can probably happen in a subsequent iteration

node/bindnode/infer.go
```diff
@@ -87,6 +104,78 @@ func fieldNameFromSchema(name string) string {
...
+
+// TODO: we should probably avoid re-spawning the same types if the TypeSystem
+// has them, and test that that works as expected
```
willscott, line 125: this seems like an optimization, right? keeping multiple primitive types reflecting each distinct aliased name in go seems like a reasonable place to end up.

node/bindnode/infer.go
```diff
@@ -87,6 +104,78 @@ func fieldNameFromSchema(name string) string {
...
+
+				// TODO: support nullable/optional with tags
+				false,
+				false,
+			)
+		}
+		name := schema.TypeName(typ.Name())
+		if name == "" {
+			panic("TODO: anonymous composite types")
+		}
```
willscott, line 155: do you have a strategy for these anonymous types? data-dependent runtime panics are something it seems very preferable to avoid

mvdan: All these TODO panics will need to be fleshed out before we can call bindnode "ready for production use" for sure.

I left a comment further up with what I'm doing for anonymous slices and why. Structs are trickier, because I don't want to do `Struct_Foo_Bar_Baz` for `struct{ Foo; Bar; Baz }`. So I might end up just erroring in this scenario, or using incrementing struct names like `Struct_1`, `Struct_2`, etc.

node/bindnode/infer.go
mvdan: Yep indeed. Just trying to keep PRs small :)

node/bindnode/infer.go
mvdan: On one hand, yes, it's about deduplication. If I call Wrap on the same Go type ten thousand times, I don't want to add ten thousand schema types to a typesystem. They should all use the same schema type pointer.

On the other hand, it's also about correctness. If you call Wrap on two values of the same Go type, which gives you a TypedNode, and you grab their schema types... I think one would expect the two to have exactly the same schema type pointer.

node/bindnode/infer.go
mvdan: Here's one reason why having a global default can be a bad idea: if package A wraps its named type T1, that gets registered in the global typesystem. If package B wraps its own different named type T1, then the two names conflict.

I don't want to use one typesystem per Wrap call, because that's just wasteful.

I don't want to force the user to supply a TypeSystem, because most people will just not want to care about this implementation detail.

But one single global fallback is also not enough, because the names of Go types are package-scoped, not universe-scoped.

Maybe one fallback TypeSystem per Go package would be the sweet spot. A similar alternative would be to use fully qualified Go names as schema type names, like `A.T1` instead of just `T1`, though I'm not sure if that's not okay in IPLD land.

node/bindnode/infer.go
mvdan: That said, data-dependent errors aren't particularly nice, either.

Thinking about it some more, this only changes behavior based on Go types, and nearly every bindnode user will be using statically declared Go types. So... I don't think these edge cases would surprise anyone, ever.

node/bindnode/infer.go
willscott: My initial instinct is that a munged version of the fully qualified go name is the right way to register these reflected types. It may need to be `A_T1` rather than `A.T1`, but that seems to have the correct level of disambiguation that's desired.

mvdan: I actually think that one typesystem per Go package just won't work, because it's reasonable to want to call bindnode.Wrap on A.T1 which references B.T2. And the wrapped B.T2 must be in the same typesystem for IPLD to work.

I agree that it seems like "fully qualified names" seems the best approach, be it `pkg/path_T1`, `pkg_path_T1`, `pkg/path.T1`, etc.

node/bindnode/infer.go
```diff
@@ -87,6 +104,78 @@ func fieldNameFromSchema(name string) string {
...
+
+func init() {
+	defaultTypeSystem.Init()
+
+	defaultTypeSystem.Accumulate(schemaTypeBool)
+	defaultTypeSystem.Accumulate(schemaTypeInt)
+	defaultTypeSystem.Accumulate(schemaTypeFloat)
+	defaultTypeSystem.Accumulate(schemaTypeString)
+	defaultTypeSystem.Accumulate(schemaTypeBytes)
+	defaultTypeSystem.Accumulate(schemaTypeLink)
```
warpfork, line 117: I think it may be a good time to plonk all these into a single helper function over in the schema package (or possibly in the near future, `schema/compiler`, if things get factored out that way) called `schema.AccumulateStandardPrelude`.

node/bindnode/infer.go
warpfork: Fwiw, I used "dunder" (*cough* if you will) a la `List__SomeType` for this throughout codegen.  The main reason for this was that saying underscores are a reserved character seems a little aggro to me, and saying double-underscores are reserved does not seem as aggro to me.  (So, roughly the same logic python used, that graphql uses, etc.)

warpfork:
I find myself with pretty minimal opinions on this diff, because I don't know how often I expect to using the total-infer no-schema mode myself.  But nothing looks super out of line to me, either.  So :+1: 

The main thing I'm worried about with total-infer no-schema mode at all is that we'll find there's something that's basically inexpressible in it.  And if we reach that point, then potentially a lot of it might be for naught, because we'll end up recommending users away from that mode anyway because it has this ceiling to its features.  So soon I'd like to increase confidence we're not gonna get in that situation.

I think drafting some examples of how we expect to customize stuff with comment tags might be a good step to take soon.  I know that's mostly fenced behind TODOs for scope control in this diff, and that's fine, but I'm a little worried we'll run into a brick wall there.  For example, because Golang doesn't have a tagging convention for types (just fields), I'm wondering how we can annotate the representation strategy onto a type.

node/bindnode/infer.go
mvdan: I don't feel strongly about this, though I do default to shorter names where possible :) To my mind, this is a fallback/default, not necessarily a reserved name.

node/bindnode/infer.go
mvdan: Separated into https://github.com/ipld/go-ipld-prime/issues/208 :)

mvdan:
We certainly will hit a limit, because Go types are relatively simple. We could always have extra ways of supplying options, though they might be verbose or fiddly enough to a point that people might prefer to just write a proper schema file.

I disagree that hitting such a limit would be a problem or make "schema inferrence" unnecessary, though. I think a significant portion of IPLD users will have simple use cases, just like in Go, the vast majority of people who encode or decode JSON are fine with `encoding/json` and its limits. Think dealbot for example: we literally just want to store a bunch of structs and lists with basic types. The only special schema feature we use is "optional", for the sake of adding extra fields to existing structs in a backwards-compatible manner, and that would fit just fine with Go field tags.

Even for the people who may eventually want complex schemas, it would still be pretty good to start with inferring them. It would halve their initial "IPLD Node" complexity, and we could always add a utility that produces a schema file via `inferSchema` to ease their transition into a manual schema :)

mvdan:
I should also say that I think generics will open some doors in terms of allowing more Go type expressiveness. For example, right now, the best way to represent an IPLD map with a nullable value type might be one of the following:

```
type MyMap map[Key]*Value // nullable iff pointer type

// or

type MyMap map[Key]Value

func (Value) Optional() {} // static method signals this type always tries to be optional
```

With generics, we could do something like:

```
package bindnode

type Nullable[T] struct {
    NonNull bool
    Value T
}

// and then...

type MyMap map[Key]Nullable[Value]
```

mvdan:
I've replied to everything, and I don't think any actionable comment remains. All the thoughts are useful though - I'll be ironing these out in future PRs.

node/bindnode/infer.go
warpfork: > I don't want to force the user to supply a TypeSystem, because most people will just not want to care about this implementation detail.

I think of all the constraints we could choose to drop here: I'd probably pick this one?

How would it feel to have a `bindnode.Analyze(goValue interface{}) (schema.TypeSystem, error)` API?

warpfork: On second thought: the concept of an `Analyze` call like that doesn't necessarily solve the problem, either, because it's possible to have a type system that has more than one "root".  (It's a full directed graph; it's not guaranteed to have a single start node from which a walk will reach all other nodes.)  That seems to point to a more fundamental challenge: either the analysis phase has to take multiple parameters (not easy to use), or we'd have to have mutable `schema.TypeSystem` (which... oof, interesting).

warpfork: Maybe the sweet spot actually is creating one implicit typesystem per Wrap call, and nevermind the waste, because anything else seems to be either complicated or have potential to generate user surprise.

If the typesystem inference path is only for quick-and-dirty rapid development anyway, it seems fine if it has poor performance.

mvdan: We could always have our cake and eat it too: add `...Option` to the bindnode APIs, and have one option where the user can supply a typesystem. If none is supplied, a global fallback is used. I think that would be good enough for most people: those who want to be in charge of their typesystems can do that, and those who don't know what that means will probably be fine with using a global one (with fully qualified type names to avoid name clashes).

I'm not sure what you mean by mutable TypeSystem. For a global one, we'd simply accumulate more types as needed. I think the typesystem supports accumulating some types, validating the graph, using that typesystem, and later on repeating that process if we infer more IPLD schemas.

mvdan: > If the typesystem inference path is only for quick-and-dirty rapid development anyway, it seems fine if it has poor performance.

The current implementation is certainly quick-and-dirty. I'm leaving performance for later, for the sake of getting things working first. But I definitely don't intend schema inferrence as only a temporary step with poor performance; one should be able to get it to a point where it's fast enough for most non-bottleneck use cases. Deduplicating work is certainly part of that, just like how encoding/json does with a global `sync.Map` to avoid redoing reflect work :)

PR #209, opened 2021-07-23
Addtional examples of bindnode usage.

warpfork:
I added a few more examples of bindnode usage while poking around to make sure I knew for myself what it can do :)

Including exercise of:

- stringjoin representations of structs

- stringprefix representations of unions

- schema field names that do not directly equal go struct field names

- deserialization!

I'm not sure if they're ideal documentation; they're just what I ended up plonking out for myself to see.

### DIFF
--- a/node/bindnode/example_test.go
+++ b/node/bindnode/example_test.go
@@ -1,6 +1,8 @@
 package bindnode_test
 
 import (
+	"bytes"
+	"fmt"
 	"os"
 
 	ipld "github.com/ipld/go-ipld-prime"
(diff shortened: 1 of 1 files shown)


node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+			},
+		),
+	))
+	schemaType := ts.TypeByName("FooOrBar")
+
+	// The golang structures for unions don't look as simple as one might like.  Golang has no native sum types, so we do something interesting here.
+	type FooOrBar struct {
+		Index int
+		Value interface{}
+	}
```
willscott, line 158: is this really the best we can do?

node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+		[]schema.StructField{
+			schema.SpawnStructField("foo", "String", false, false),
+			schema.SpawnStructField("bar", "String", false, false),
+			schema.SpawnStructField("baz", "String", false, false),
+		},
+		schema.SpawnStructRepresentationStringjoin(":"),
+	))
+	schemaType := ts.TypeByName("FooBarBaz")
+
+	type FooBarBaz struct {
```
rvagg, line 100: How is this mapping working? Does it just use field order to figure out what to assign where? So the schema could have any arbitrary name but the values get assigned to whatever order is there?

node/bindnode/example_test.go
rvagg: This seems like a reasonable solution to an awkward Go limitation. Other options would probably involve nested structs which starts to detract from the concept of a union, or is there some other option here?
Maybe unions should have some optionality to back them by different strategies depending on user preference.

node/bindnode/example_test.go
mvdan: Yes, when mapping a provided IPLD schema with a provided Go type, it just cares about the "shape" matching, not any names. For a struct, that means that the fields should have matching types in the same order.

Doing the mapping via names or struct field tags would also be an option, though it would require more code and features, so I initially just went for the simplest approach from my point of view.

I haven't documented this, mind you, because I'm not 100% certain that this default behavior is right. I think most users would expect case-insensitive-matching by field name, similar to most other Go libraries. I'm not bothered by this example being public, but the semantics in this case might change.

node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+		Foo: "x",
+		Bar: "y",
+		Baz: "z",
+	}
+	node := bindnode.Wrap(fbb, schemaType)
+
+	// Take the representation of the node, and serialize it.
+	nodeRepr := node.Representation()
+	var buf bytes.Buffer
+	dagjson.Encode(nodeRepr, &buf)
```
mvdan, line 115: check error?

node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+
+	// Output how this was serialized, for the example.
+	fmt.Fprintf(os.Stdout, "json: %s\n", buf.Bytes())
```
mvdan, line 118: nit: for the sake of keeping the example simpler, you could just dagjson.Encode straight to os.Stdout, printing `json:` in the line above or in the same line via Printf

node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+
+	// Now unmarshal that again and print that too, to show it working both ways.
+	np := bindnode.Prototype(&FooBarBaz{}, schemaType)
```
mvdan, line 121: nit: you could just do `node.Prototype()` here, unless you want to showcase `bindnode.Prototype` too

node/bindnode/example_test.go
mvdan: I definitely intend to revisit this. "integer plus empty interface" just was the simplest implementation that let me push forward with the implementation - and it's also what codegen uses.

It's possible that we could support just an interface value, and detect what type is currently stored via the Go type. Though that would require some extra logic, so I didn't want to implement that straight away.

Another option would be to make it arbitrary logic, like:

```
type Union interface {
    Set(idx int, value interface{})
    Get() (idx int, value interface{})
}
```

I think it makes sense to wait seven months for generics to land to revisit this, because generics would potentially offer us a far better option:

```
type FooOrBar interface {
    type Foo, Bar
}
```

Generics would also open more doors in terms of expressing custom map implementations, optional types, nullable types, etc - all without the `interface{}` shenanigans.

node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+	}
+	node := bindnode.Wrap(fob, schemaType)
+
+	// Take the representation of the node, and serialize it.
+	nodeRepr := node.Representation()
+	var buf bytes.Buffer
+	dagjson.Encode(nodeRepr, &buf)
+
+	// Output how this was serialized, for the example.
+	fmt.Fprintf(os.Stdout, "json: %s\n", buf.Bytes())
```
mvdan, line 171: same comments as in the other example: error and buffer :)

node/bindnode/example_test.go
```diff
@@ -80,3 +82,104 @@ func ExamplePrototype_onlySchema() {
...
+
+	// Now unmarshal that again and print that too, to show it working both ways.
+	np := bindnode.Prototype((*FooOrBar)(nil), schemaType)
```
mvdan, line 174: nit: in one example you use `&T{}`, and in the other you use `(*T)(nil)`. the latter is slightly better, as it avoids an alloc. but you should also be consistent.

mvdan (commented):
I wonder if we should join these examples into one, and call it "with custom representation". After all, the examples should show the different features of the bindnode package, not the different features of IPLD schemas :)

node/bindnode/example_test.go
mvdan: On second thought, the current way of dealing with unions is just plain bad. See https://github.com/ipld/go-ipld-prime/issues/210. I'm marking this as resolved because they'll need to be reworked anyway.

mvdan:
I've incorporated a modified version of the union example for https://github.com/ipld/go-ipld-prime/pull/223.

> schema field names that do not directly equal go struct field names

I actually think we should delay exposing an example for this, because it's quite likely I'll change how this works - think Go field tags.

I also want to implement checks that ensure a Go type and a schema type are valid when supplied together. This does not exist right now; bindnode blindly trusts the user. So it's a bit dangerous to encourage the users to do funky things with mismatching field names right now.

> stringjoin representations of structs

There's nothing particularly special about IPLD schema structs, and the other examples already cover structs and custom representation strategies, so I'm inclined to say we don't want this example. We want to keep the number of doc examples relatively low, I'd say between 3-6 ideally. With my unions PR, and with a future "override field names" example, we're already at 5. I don't want to end up with two or three times as many examples, as then the user would just be overloaded with information :)

mvdan:
I think we can close this; I already incorporated most of these in a different form in another PR of mine.

PR #211, opened 2021-07-26
fix: typed links LinkTargetNodePrototype should return ReferencedType

rvagg:
Found in https://github.com/ipfs/go-ipld-git/pull/46 integrating into go-ipfs
which wants to load typed nodes when it encounters typed links. It errors
because it's trying to load the typed node into the type of the typed link
rather than the referenced type.

https://github.com/ipfs/go-ipld-git/pull/46 currently has codegen using this branch and is now working in go-ipfs doing path traversals through nested linked objects 👌.

### DIFF
--- a/node/bindnode/schema_test.go
+++ b/node/bindnode/schema_test.go
@@ -14,6 +14,9 @@ import (
 // In the future, forSchemaTest might return multiple engines.
 
 func forSchemaTest(name string) []tests.EngineSubtest {
+	if name == "Links" {
+		return nil // TODO(mvdan): add support for links
+	}
 	return []tests.EngineSubtest{{
 		Engine: &bindEngine{},
 	}}
--- a/node/tests/schema.go
+++ b/node/tests/schema.go
@@ -8,6 +8,7 @@ var allSchemaTests = []struct {
 	name string
 	fn   func(*testing.T, Engine)
 }{
+	{"Links", SchemaTestLinks},
 	{"ListsContainingMaybe", SchemaTestListsContainingMaybe},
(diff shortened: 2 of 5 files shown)


mvdan (approved):
Hmm, we should have a test for this, because now you're making me wonder if bindnode would have made the same mistake.

Entirely reasonable if you don't have time to write a test right now, but in that case we should file an issue as a reminder :)

rvagg:
@mvdan 👍 I wanted to, but I burned my whole day trying to figure out how so many of the pieces fit together for the ipld-in-ipfs work that I haven't been up to speed on! Just finding this bug took an unreasonably long time that I'm sure someone more familiar would not have had to spend.. a good learning experience though!
Will pick this up tomorrow and figure out what a test would look like.

rvagg:
@mvdan &&/|| @warpfork can you take a look at the latest commit please? Maybe I've gone overboard here but it seemed that to test this properly I had to pull in traversal to the standard node/tests/ to traverse some typed and untyped links and confirm that they pop out from the loader in the right types. It's fine for gen/go but node/bindnode isn't happy but maybe it can't do what I'm asking it to do here so shouldn't have it in the standard list of tests in node/tests/schema.go?

mvdan:
A shared test is definitely right :) let me look at bindnode today, the fix might be easy. If not, I'll make bindnode's tests skip this one with a TODO.

mvdan:
I looked at adding support for this in bindnode for an hour, and made progress, but it needs more work. So I've stashed those changes, and instead pushed a change here to skip the new test in bindnode :) Still LGTM.

PR #214, opened 2021-08-03
Add link conditions for 'stop-at' expression in ExploreRecursive selector

willscott:
allows selectors of the form:
```json
{
		"R": {
			"l": {
				"none": 0
			},
			":>": {
				"a": {
					">": {
						"@": {}
					}
				}
			},
			"!": {
				"/": {
                                       "/":"bafyreifepiu23okq5zuyvyhsoiazv2icw2van3s7ko6d3ixl5jx2yj2yhu"
                                }
			}
		}
	}
```

### DIFF
--- a/traversal/selector/condition.go
+++ b/traversal/selector/condition.go
@@ -28,7 +28,7 @@ const (
 func (c *Condition) Match(n ipld.Node) bool {
 	switch c.mode {
 	case ConditionMode_Link:
-		if n.Kind() != ipld.Kind_Link || c.match.Kind() != ipld.Kind_Link {
+		if n.Kind() != ipld.Kind_Link {
 			return false
 		}
 		lnk, err := n.AsLink()
--- a/traversal/selector/condition.go
+++ b/traversal/selector/condition.go
@@ -1,0 +1,74 @@
+package selector
+
+import (
+	"fmt"
+
+	ipld "github.com/ipld/go-ipld-prime"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+)
+
+// Condition provides a mechanism for matching and limiting matching and
+// exploration of selectors.
+// Not all types of conditions which are imagined in the selector specification
+// are encoded at present, instead we currently only implement a subset that
+// is sufficient for initial pressing use cases.
+type Condition struct {
+	mode  ConditionMode
+	match ipld.Node
+}
+
+// A ConditionMode is the keyed representation for the union that is the condition
+type ConditionMode string
+
+const (
+	ConditionMode_Link ConditionMode = "/"
+)
+
+// Match decides if a given ipld.Node matches the condition.
+func (c *Condition) Match(n ipld.Node) bool {
+	switch c.mode {
+	case ConditionMode_Link:
+		if n.Kind() != ipld.Kind_Link || c.match.Kind() != ipld.Kind_Link {
+			return false
+		}
+		lnk, err := n.AsLink()
+		if err != nil {
+			return false
+		}
+		match, err := c.match.AsLink()
+		if err != nil {
+			return false
+		}
+		cidlnk, ok := lnk.(cidlink.Link)
+		cidmatch, ok2 := match.(cidlink.Link)
+		if ok && ok2 {
+			return cidmatch.Equals(cidlnk.Cid)
+		}
+		return match.String() == lnk.String()
+	default:
+		return false
+	}
+}
+
+// ParseCondition assembles a Condition from a condition selector node
+func (pc ParseContext) ParseCondition(n ipld.Node) (Condition, error) {
+	if n.Kind() != ipld.Kind_Map {
+		return Condition{}, fmt.Errorf("selector spec parse rejected: condition body must be a map")
+	}
+	if n.Length() != 1 {
+		return Condition{}, fmt.Errorf("selector spec parse rejected: condition is a keyed union and thus must be single-entry map")
+	}
+	kn, v, _ := n.MapIterator().Next()
+	kstr, _ := kn.AsString()
+	// Switch over the single key to determine which condition body comes next.
+	//  (This switch is where the keyed union discriminators concretely happen.)
+	switch ConditionMode(kstr) {
+	case ConditionMode_Link:
+		if _, err := v.AsLink(); err != nil {
+			return Condition{}, fmt.Errorf("selector spec parse rejected: condition_link must be a link")
+		}
+		return Condition{mode: ConditionMode_Link, match: v}, nil
+	default:
+		return Condition{}, fmt.Errorf("selector spec parse rejected: %q is not a known member of the condition union", kstr)
+	}
+}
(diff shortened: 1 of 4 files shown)


traversal/selector/exploreRecursive.go
```diff
@@ -149,8 +154,11 @@ func (s ExploreRecursive) replaceRecursiveEdge(nextSelector Selector, replacemen
 	return nextSelector
 }
 
-// Decide always returns false because this is not a matcher
+// Decide if a node directly matches
 func (s ExploreRecursive) Decide(n ipld.Node) bool {
+	if s.stopAt != nil && s.stopAt.Match(n) {
+		return false
+	}
```
warpfork, line 161: I'm wobbling on if this is going to be the desired behaviour.  The semantic question is: does `stopAt` mean "don't explore beyond this", or that _and also_ "don't match this"?

traversal/selector/exploreRecursive.go
```diff
@@ -192,7 +200,16 @@ func (pc ParseContext) ParseExploreRecursive(n ipld.Node) (Selector, error) {
 	if erc.edgesFound == 0 {
 		return nil, fmt.Errorf("selector spec parse rejected: ExploreRecursive must have at least one ExploreRecursiveEdge")
 	}
-	return ExploreRecursive{selector, selector, limit}, nil
+	var stopCondition *Condition
+	stop, err := n.LookupByString(SelectorKey_StopAt)
+	if err == nil {
+		condition, err := pc.PushParent(erc).ParseCondition(stop)
```
warpfork, line 206: This doesn't seem right.  I don't think you want to be doing PushParent twice here, I think you just want `pc.ParseCondition(stop)` and that's it.

Might be good to get a little more testing for this path.  I'd be slightly surprised if this doesn't break something when stopAt is present, as written.

warpfork (changes requested):
Mostly questions; one thing I'm pretty sure is a bug.

traversal/selector/condition.go
```diff
@@ -0,0 +1,74 @@
...
+			return cidmatch.Equals(cidlnk.Cid)
+		}
+		return match.String() == lnk.String()
+	default:
+		return false
+	}
+}
+
+// ParseCondition assembles a Condition from a condition selector node
+func (pc ParseContext) ParseCondition(n ipld.Node) (Condition, error) {
```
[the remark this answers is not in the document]
willscott: this is already in the schema. at the bottom of https://ipld.io/specs/selectors/

traversal/selector/exploreRecursive.go
warpfork: I lean towards thinking the second part is unnecessary.

Reasons:
- "stopAt" to me reads as "i will reach this point and go no further".  That means we can still match here.
- Less dropping the "and also don't match this" makes the overall description of behavior shorter.  Shorter is a heuristic for better.
- It's only even relevant if you have a matcher clause that lands on the same node as the stop.  You can just... choose not to do that?
- ... and since the main practical purpose of the stops is to avoid _loading_ more data, usually, then, matching on some data you've already loaded isn't really the big worry.

In that case it means we don't need the new branch here.

I wouldn't mind a second opinion on this, though.  These reasons aren't strongly held.

traversal/selector/exploreRecursive.go
willscott: removed the pushing of parent context, but i don't think it's used in parsing any of the condition potential union cases so it should be a no-op.

traversal/selector/exploreRecursive.go
willscott: removed

traversal/selector/condition.go
```diff
@@ -0,0 +1,74 @@
...
+
+const (
+	ConditionMode_Link ConditionMode = "/"
+)
+
+// Match decides if a given ipld.Node matches the condition.
+func (c *Condition) Match(n ipld.Node) bool {
+	switch c.mode {
+	case ConditionMode_Link:
+		if n.Kind() != ipld.Kind_Link || c.match.Kind() != ipld.Kind_Link {
```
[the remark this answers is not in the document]
willscott: skipped the check, which should cause the direct assumption that it's a link to fail if the assumption fails

willscott:
Note: this allows stopping recursion at a sub-dag, but doesn't fully de-duplicate 'only new stuff not in the sub-dag i already have' when doing a graph sync which is something you might hope for.

Imagine you're doing something like synchronizing a blockchain, and you want to express 'i already have things up to block <whatever>'. where each block links to a state tree that evolves over blocks. The functionality of this stop-at isn't quite what you might intuitively what which would be 'don't recurse into parts of the state tree on the new blocks that haven't changed since the block i already have'. The problem with that is that you don't know what all those cids are. You could have a selector to stop at 'all the cids', but that's a huge list to transfer, or you could imagine a form of selector to ask the other side to actually form a set out of the subdag represented by the block you have and not send any of those again, but that's asking for a lot of of computational work from the other side (it would have to walk the full history of the blockchain to generate such a set).
This is an initial compromise, but important to realize that how this selector can be used implies the data structure has been created with some care.

warpfork:
Yeah, I think I could still also image a feature someday which might propose something like:

```ipldsch
type TraversalInstructions struct {
    selector Selector
    stopAtAny [Link]
}
```

Such a feature would differ from this by putting all the stop points in one place towards the root of a protocol conversation, and taking global effect for the walk (e.g. you don't have to know where in the topology the stop is at), and also by being a _set_ (whereas to get that by extending the condition feature, we'd be presumably be implementing 'and' clauses, and I suspect that might become... harder to optimize).

This `stopAt` feature using `Condition` is neither the same as, nor in conflict with, such a possible feature.  :)

warpfork:
I am looking one more time at the specification that we have for `Condition` in the Selector system.  It's... not well specified.  I want to remark on this briefly here in case this comes up again in the near future.

---

### it is not well specified

The current degree of specification we have is this excerpt from schema for the `Condition` union:

```
	| Condition_HasValue "=" # will need to contain a kinded union, lol.  these conditions are gonna get deep.
	| Condition_HasKind "%" # will ideally want to refer to the DataModel ReprKind enum...!
	| Condition_IsLink "/" # will need this so we can use it in recursions to say "stop at CID QmFoo".
```

There is _no actual specification, at this time_, of what any of those types (`Condition_IsLink`, etc) ... _are_.

(The same doc also says """The "Condition" system is not fully specified -- it is a placeholder awaiting further design.""" under the "known issues" heading.  It's _highly_ disclaimed.)

---

### why I'm wondering about this

I don't really know why `Condition_IsLink` exists distinctly from those other two.

The name would seem to imply it's a special case of `Condition_HasKind`?

Or, if it's doing equality check, as the comment seems to suggest, I don't know why `Condition_HasValue` wouldn't be able to do that.

(And to be clear: I wrote all these lines.  I'm saying _I don't know what I was thinking_ at the time.  This looks like a sketch kicked out by a tired brain that never ever got reviewed.)

---

### what we seem to have locked in on, with this PR

We seem to have decided that this would be appended to the Selector schema:

```
type Condition_IsLink link
```

And that `Condition_IsLink` is really just an equality check.

We're just quietly ignoring the fact that `Condition_HasValue` (if we bothered to finish specifying it, which, again, we haven't yet) would probably also suffice for doing an equality check.

Okay.  :shrug: 

This is probably fine.  If it is duplicate feature, we're pushing a bit of silly work into future implementors, but... not very much.  There's a lot going on right now, and this is not something I think anyone contributing in the area is willing to chase right now.

PR #217, opened 2021-08-04
Allow parsing of single Null tokens from refmt

rvagg:
For blocks that are simply "null":
 * dag-cbor: `0xf6`
 * dag-json: `null`

### DIFF
--- a/codec/dagcbor/unmarshal.go
+++ b/codec/dagcbor/unmarshal.go
@@ -47,7 +47,7 @@ func unmarshal1(na ipld.NodeAssembler, tokSrc shared.TokenSource, gas *int, opti
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


willscott (approved):
consider a fixture or test? but this seems like a pretty uncontroversial win

rvagg:
tests over @ https://github.com/ipld/codec-fixtures

I'm planning on attempting an auto-gen sharness version of the fixtures to run them through go-ipfs, I could also do that for here as well but it would be doubling-up since I have go-ipld-prime running them over there already

rvagg:
.. or we could git submodule, or wire it up into an Actions script to run on CI, or just do it manually

willscott:
as long as we monitor it somewhere and make sure it's running before releases, i'm in favor of whatever seems least complex but also won't just code-rot away.

PR #221, opened 2021-08-12
codecs: more docs, a terminology guide, consistency in options.

warpfork:
No breaking changes.  In fact I think this should be behaviorally identical to before, despite the size of the diff. There are some functions that are explicitly marked as deprecated now, but I am not removing them in this diff.

One exception: some struct names did change.  However, these have not been included in any tagged release so far, so I believe changing them now is still pretty fair game, and the number of affected people should be slim to none.

The main source of changes is pushing towards nomenclature consistency, which is described in the new readme file also included the diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,7 @@ When a release tag is made, this block of bullet points will just slide down to 
 	- The codecs do not reject other orderings when parsing serial data.
 	  The `ipld.Node` trees resulting from deserialization will still preserve the serialized order.
 	  However, it has now become impossible to re-encode data in that same preserved order.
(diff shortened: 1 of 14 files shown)


warpfork:
It was also discussed in another forum earlier that perhaps it would be nice if the zero values for these config options would actually be the default for that codec.  However, that would conflict with using the same constants to describe things across packages, which seems silly, so I've dropped that as a goal.

mvdan (approved):

mvdan:
Since the default options are not the zero value, we might want to add a `func DefaultOptions() OptionsType` later on - but not a blocker.

PR #223, opened 2021-08-12
node/bindnode: redesign the shape of unions in Go

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/api.go
+++ b/node/bindnode/api.go
@@ -21,6 +21,7 @@ import (
 // the other provided type. For example, we can infer an unnamed Go struct type
 // for a schema struct tyep, and we can infer a schema Int type for a Go int64
 // type. The inferring logic is still a work in progress and subject to change.
+// At this time, inferring IPLD Unions and Enums from Go types is not supported.
 //
 // When supplying a non-nil ptrType, Prototype only obtains the Go pointer type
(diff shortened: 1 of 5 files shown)


warpfork:
This... looks essentially sensible.  :+1: 

The one thing I'm trying to wrap my head around is what the error pathways look like when the golang structure for a union doesn't actually match the union membership in some way.  E.g. if it has too few fields, or too many; or if a field's golang type doesn't actually have the appropriate internal structure for whatever we expect next according to the schema.  Most of these are going to result in panics fairly late in the game, right?

I guess that's probably acceptable for the current level of productionization of `bindnode`.  Just trying to check my understanding and expectations.

mvdan:
That's right. This is what I alluded to earlier in https://github.com/ipld/go-ipld-prime/pull/209#issuecomment-897783756. When this is finished, providing both a schema and a Go type will mean a "sanity check" to ensure that the shapes match properly. If that's happy, then reflect panics later on should be impossible (assuming we don't make mistakes in the implementation).

And indeed, I skipped that for the sake of cutting corners towards a PoC :) It's one of the biggest TODOs in the code in terms of what's required to call this ready for production use.

warpfork (approved):
Alright, I'm good with this.  (And excited to use it!)  Let's merge it!

PR #226, opened 2021-08-14
Bindnode fixes of opportunity

warpfork:
A variety of error message improvements and polish as diffs.

I'm trying to just track these things and do target-of-opportunity work as I try to be a downstream user.  The commit messages still say what and why, but the focus is... whatever I hit. :)

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -749,7 +749,8 @@ func (w *_structAssembler) AssembleValue() ipld.NodeAssembler {
 	}
 	ftyp, ok := w.val.Type().FieldByName(fieldNameFromSchema(name))
 	if !ok {
-		panic("TODO: go-schema mismatch")
+		// It is unfortunate this is not detected proactively earlier during bind.
+		panic(fmt.Errorf("schema type %q has field %q, we expect go struct to have field %q", w.schemaType.Name(), field.Name(), fieldNameFromSchema(name)))
 	}
 	if len(ftyp.Index) > 1 {
 		panic("TODO: embedded fields")
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -551,32 +551,11 @@ func (w *_assemblerRepr) AssignString(s string) error {
 		}
 		return mapAsm.Finish()
 	case schema.UnionRepresentation_Kinded:
-		name := stg.GetMember(ipld.Kind_String)
-		members := w.schemaType.(*schema.TypeUnion).Members()
-		for idx, member := range members {
-			if member.Name() != name {
-				continue
-			}
-
-			// TODO: DRY with behavior below.
-			w2 := *w
-			goType := w.val.Field(idx).Type().Elem()
-			valPtr := reflect.New(goType)
-			w2.val = valPtr.Elem()
-			w2.schemaType = member
-			w2.finish = func() error {
-				if w.finish != nil {
-					if err := w.finish(); err != nil {
-						return err
-					}
-				}
-				unionSetMember(w.val, idx, valPtr)
-				return nil
-			}
-
-			return w2.AssignString(s)
+		w2 := w.asKinded(stg, ipld.Kind_String)
+		if w2 == nil {
+			panic("TODO: GetMember result is missing?")
 		}
-		panic("TODO: GetMember result is missing?")
+		return w2.AssignString(s)
 	case schema.UnionRepresentation_Stringprefix:
 		hasDelim := stg.GetDelim() != ""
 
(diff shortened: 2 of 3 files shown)


node/bindnode/node.go
```diff
@@ -749,7 +749,8 @@ func (w *_structAssembler) AssembleValue() ipld.NodeAssembler {
 	}
 	ftyp, ok := w.val.Type().FieldByName(fieldNameFromSchema(name))
 	if !ok {
-		panic("TODO: go-schema mismatch")
+		// It is unfortunate this is not detected proactively earlier during bind.
+		panic(fmt.Errorf("go type and schema types mismatch: schema type %q has field named %q, so we expected go struct to have field named %q", w.schemaType.Name(), field.Name(), fieldNameFromSchema(name)))
```
mvdan, line 753: I find this error way too long. how about:

```
schema type %q has field %q, we expect go struct to have field %q
```

node/bindnode/repr.go
```diff
@@ -637,6 +664,13 @@ func (w *_structAssemblerRepr) AssembleKey() ipld.NodeAssembler {
 	switch stg := reprStrategy(w.schemaType).(type) {
 	case schema.StructRepresentation_Map:
 		return (*_structAssembler)(w).AssembleKey()
+	case schema.StructRepresentation_Stringjoin,
+		schema.StructRepresentation_StringPairs:
+		// TODO an error-carrying "NodeAssembler" is needed so that this can refrain from panicking.
+		// TODO perhaps the ErrorWrongKind type should also be extended to explicitly describe whether the method was applied on bare DM, type-level, or repr-level.
+		panic(&ipld.ErrWrongKind{TypeName: string(w.schemaType.Name()) + ".Repr", MethodName: "AssembleKey", AppropriateKind: ipld.KindSet_JustMap, ActualKind: ipld.Kind_String})
+	case schema.StructRepresentation_Tuple:
+		panic(&ipld.ErrWrongKind{TypeName: string(w.schemaType.Name()) + ".Repr", MethodName: "AssembleKey", AppropriateKind: ipld.KindSet_JustMap, ActualKind: ipld.Kind_List})
```
mvdan, line 673: multiline expressions please :) should be far more readable too

node/bindnode/repr.go
```diff
@@ -589,10 +617,9 @@ func (w *_assemblerRepr) AssignString(s string) error {
 				unionSetMember(w.val, idx, valPtr)
 				return nil
 			}
-
-			return w2.AssignString(value)
+			return w2.AssignString(remainder)
 		}
-		panic("TODO: GetMember result is missing?")
+		return fmt.Errorf("data doesn't match union type: type %s couldn't match any of its prefixes to the data %q", w.schemaType.Name(), s)
```
mvdan, line 622: I'd also try to make this error shorter. you say "does not match", "data", and "type" twice, so the first sentence could be removed, for example

node/bindnode/repr.go
```diff
@@ -557,22 +557,50 @@ func (w *_assemblerRepr) AssignString(s string) error {
...
+				if w.finish != nil {
+					if err := w.finish(); err != nil {
+						return err
+					}
+				}
+				unionSetMember(w.val, idx, valPtr)
+				return nil
+			}
+
+			return w2.AssignString(s)
```
mvdan, line 577: have you seen asKinded? because this looks extremely similar. we use asKinded in AsString on kinded unions, so it makes sense to use it in AssignString too.

The other AsT/AssignT lack the support for kinded unions simply because I haven't gotten to propagating the code.

mvdan (approved):
I know you have more fixes incoming, but probably best to wrap this up and merge it. Reviewing a five-commit PR is tricky enough as it is :)

node/bindnode/repr.go
warpfork: Oh, nice.  I had not seen it; that works perfectly.

warpfork:
Took all suggestions.

(Though I'm not sure I favour prioritizing error message brevity much, at the moment.  So far, my experience as a user is that I want more info approximately 100% of the time, and I'd much rather get clarity at the expense of brevity.  But, that said, we're also still at the "printf" stage of development on any of the errors, so I'm going to hope things are improved in general in future passes that make error handling more structured.)

node/bindnode/repr.go
mvdan: It's possible some of the code could do with more reuse. I haven't paid *too* much attention to that just yet. I find a bit of copy-pasting to be necessary for the sake of quick iteration - otherwise I spend extra time extracting and inlining common functions :)

warpfork:
Gonna enthusiastically merge this, more can come later :)

PR #229, opened 2021-08-18
Fix ExploreRecursive stopAt condition and add tests

adlrocha:
This PR:
- Fixes the bug in `ExploreRecursive` stopAt condition.
- Adds a bunch of test cases to test the expected functionality of this feature.


traversal/selector/exploreRecursive.go
```diff
@@ -87,8 +87,14 @@ func (s ExploreRecursive) Interests() []ipld.PathSegment {
 
 // Explore returns the node's selector for all fields
 func (s ExploreRecursive) Explore(n ipld.Node, p ipld.PathSegment) Selector {
-	if s.stopAt != nil && s.stopAt.Match(n) {
-		return nil
+	if s.stopAt != nil {
+		target, err := n.LookupBySegment(p)
+		if err != nil {
+			panic(err) // oh dear
```
willscott, line 93: as much as it's unfortunate to drop the error, i wonder if it isn't safer to return nil here rather than panicing

warpfork: Yeah, I didn't know what to do here.

I think there's no answer here that's both "easy" and "correct" -- I think we've finally hit some needs that the current Selector interface isn't ready for.  This is the first time an Explore function has need to actually look into child data; as such, it can error, unlike any of its cousins to date.  It may be appropriate to update the interface to return errors from this method.  This situation here will recur for anything else that starts using the Condition system and peeking at child data, too.

Options:

- A. Bite the bullet and add an error return to the interface: `Explore(Node, PathSegment) (Selector, error)`.  Update all the other types in this package to return nil errors all the time.
- B. Leave the panic in.  It'll be hit by... really only ADLs if they have load errors, I think.  Might not be encountered much in practice right now, but painful if it does come up, of course.
- C. Return nil... and get *very* hurt some day when things start behaving logically incorrectly in complete silence and we have an undebuggable nightmare.

It might be ideal to go straight for Option A.  All the code that changes is in this package, and no where else.  @adlrocha , would that daunt you to do?

adlrocha: I am up for Option A right away, specially if we expect new selectors with this kind of behavior to come in the future. I've faced the "where this panic comes from" situation in a deep dependency before and I personally hate them. 

I'll wrap up what I am doing, and take on this after lunch. If it is as straightforward as it sounds (and the code to change is constraint to this package) I think I can handle it without bothering you too much :)

Out of a curiosity, `Explore` is only used by selectors, right? So it shouldn't break anyone else's code.

warpfork:
Thank you so much for creating test cases for this!  Tests are life.

Can I also beg you to make a PR to our specs and briefly describe how `stopAt` is working now?  Adding another paragraph to the end of the `ExploreRecursive` docs, [here](https://github.com/ipld/ipld/blame/d594fde94d60191fdcd76fcc6e0920cda18f6ba8/specs/selectors/index.md#L132), should be about the right spot.

In particular I think what we've settled on is that the stopAt condition can cause the selector to stop *anywhere* within recursion sequence, not just at its edges.  That's in need of some prose, because I'm not sure whether that would be what people will reliably intuit if it's not discussed.

traversal/selector/exploreRecursive.go
warpfork: Yes, `Explore` is only called by the code in the `traversal` package.  I don't think anything else in the universe uses it directly.

(FWIW, If I had a refactor-wand, I'd probably like to move the `selector.Selector` type to be something like `selector/engine.CompiledSelector`, or other similarly elongated name that makes it clearer that it's not typically a user-facing thing.  Maybe someday soonish. :upside_down_face:  But while that packaging refactor has no reason to be today, hopefully that gives you confidence in the change scope here.)

adlrocha:
@warpfork, it was straightforward as you promised :) I love when "oh come on is trivial" is actually trivial. Let me know if there is anything else missing. Thanks!

warpfork:
I'm thrilled it worked out :dancers: 

This looks good to me!  If you can rebase onto master, so I can click 'merge' and also keep your individual commits, that's ideal.  (Or we can squash if you don't care for the history; no big deal.)

_If_ you really want to go full hulk on this... you can now also add test fixtures as serial data in the langauge-agnostic docs+specs repo, and get them tested in here, as of https://github.com/ipld/go-ipld-prime/pull/231 and https://github.com/ipld/ipld/pull/130 .  I can work through that flow with you if you want, but also I won't hold you to it -- you've done a lot here already.

warpfork (approved):

adlrocha:
Merged

> If you really want to go full hulk on this... you can now also add test fixtures as serial data in the langauge-agnostic docs+specs repo, and get them tested in here, as of #231 and ipld/ipld#130 . 

Cool, I'll look at it tomorrow and ask you for a walkthrough if needed. Thanks!

PR #231, opened 2021-08-18
selector: add tests which are driven by language-agnostic spec fixtures.

warpfork:
This has been LONG overdue.

It depends on a matching PR to the specs, which is:
https://github.com/ipld/ipld/pull/130 .

It adds the specs+docs repo as a git submodule.
(This may be tetchy to some people, but they Solve Problems.)
If you don't check out the submodule, these tests are just skipped.

It adds github.com/warpfork/go-testmark as a new dependency.
This is what parses the fixture format.  It's nothing special;
the fixture format is also... markdown.  This parser just gives us
a good combination of human readability and machine processability.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule ".ipld"]
+	path = .ipld
+	url = https://github.com/ipld/ipld/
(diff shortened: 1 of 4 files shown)


warpfork:
Tests.  Are.  GREEN.

:green_circle: :green_circle: :green_circle: 

This is so badly overdue I'm merging it immediately.

PR #232, opened 2021-08-19
helper methods for encoding and decoding

warpfork:
Add new helper methods that shorten up the number of keystrokes needed for common operations related to encoding and decoding.

These are added in the root `ipld` package, where new users of the library should find them quickly and easily.  (This was made possible by #228 .)

Many common operations are available, but of course not all of them -- for example, if one wants to control a NodeBuilder manually and use its Reset features for allocation avoidance on repeated use, you'd still have to dive deeper and these facades won't help you.  *But*, at the same time, the sheer existence of these facades -- and their copious documentation about subjects exactly like that one! -- means users should find that easier than ever to learn how to do, if it becomes desirable to them to go there.

Some of these functions take shortcuts I never would've considered acceptable before.  For example, `Encode` and `Decode` in this diff... don't let you specify a `NodePrototype`; nada.  They just brazenly and directly use `basicnode.Prototype.Any`, without asking you about it.  Aside from being physically possible now (e.g., without provoking an import cycle, which is again due to the work in #228), this also feels acceptable to me now, as we pivot the semantics of the root package of this repo to "these are where the helpful high-level and getting-started functions go".

Similarly, I've directly exposed `ipld.Marshal` and `ipld.Unmarshal` functions, which provide quick access to some of the features of `bindnode` that a new user is most likely to want to use.  The full `bindnode` package interface, and composing it with our other packages yourself, is still clearly more powerful.  But having this facade -- which shows you _where to find_ the rest of those features, and, if necessary, _how to put them together_ yourself -- should still help first-time users enormously.

You can see the ultimate impacts to what a user sees in the included new example functions.  I think this is the first time I've been able to put example functions that actually _look good_ in one of our central, the-user-sees-it-for-sure packages, and I'm really happy about this.

### DIFF
--- a/codecHelpers.go
+++ b/codecHelpers.go
@@ -1,0 +1,167 @@
+package ipld
+
+import (
+	"bytes"
+	"io"
+	"reflect"
+
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+	"github.com/ipld/go-ipld-prime/schema"
+)
+
+// Encode serializes the given Node using the given Encoder function,
+// returning the serialized data or an error.
+//
+// The exact result data will depend the node content and on the encoder function,
+// but for example, using a json codec on a node with kind map will produce
+// a result starting in `{`, etc.
+//
+// Encode will automatically switch to encoding the representation form of the Node,
+// if it discovers the Node matches the schema.TypedNode interface.
+// This is probably what you want, in most cases;
+// if this is not desired, you can use the underlaying functions directly
+// (just look at the source of this function for an example of how!).
+//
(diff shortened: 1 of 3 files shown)


warpfork:
@willscott , you should also love this -- with these, I finally add the "automatically gearshift into representations the instant we get a whiff of typed nodes" feature.

I'm also happy, because by adding it here in a facade, we leave the possibility for a user to dive into the deeper packages, and do things like apply codecs directly to other layers... if they want, and if they know what they're doing.  But the first and most obvious functions to reach for... will now "DTRT".  (And hopefully teach people about it, since the docs on those functions touch on this behavior, too.)

willscott (approved):
You're right. I do like this addition!

rvagg:
`ENEEDSMORETESTS`

PR #233, opened 2021-08-20
Bindnode fixes continued

warpfork:
Like it says on the tin.

Details in commit messages.

### DIFF
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -226,10 +226,33 @@ func (w *_nodeRepr) MapIterator() datamodel.MapIterator {
 		itr := (*_node)(w).MapIterator().(*_unionIterator)
 		return (*_unionIteratorRepr)(itr)
 	case nil:
-		return (*_node)(w).MapIterator()
+		switch st := (w.schemaType).(type) {
+		case *schema.TypeMap:
+			return &_mapIteratorRepr{
+				schemaType: st,
+				keysVal:    w.val.FieldByName("Keys"),
+				valuesVal:  w.val.FieldByName("Values"),
+			}
+		default:
+			panic(fmt.Sprintf("TODO: mapitr.repr for typekind %s", w.schemaType.TypeKind()))
+		}
 	default:
 		panic(fmt.Sprintf("TODO: %T", stg))
 	}
+}
+
+type _mapIteratorRepr _mapIterator
+
+func (w *_mapIteratorRepr) Next() (key, value datamodel.Node, _ error) {
+	k, v, err := (*_mapIterator)(w).Next()
+	if err != nil {
+		return nil, nil, err
+	}
+	return k.(schema.TypedNode).Representation(), v.(schema.TypedNode).Representation(), nil
+}
+
+func (w *_mapIteratorRepr) Done() bool {
+	return w.nextIndex >= w.keysVal.Len()
 }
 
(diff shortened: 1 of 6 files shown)


warpfork:
I've got one more fairly nasty discovery: our approach to unions is problematic *again*: because it is now required to use pointer fields, if one wants to use a union as a map key, one is going to have a bad day.  (As long as you hold on to literally the the values originally used as keys, you can use them to look things up again; but you'll have a hard time producing any other equivalent value, and that's probably going to result in massive frustration and nasty bugs, etc.)

We'll have to reintroduce the ability to use an integer field as the in-memory discriminator, and allow non-pointer fields for the members.  I think we can make this optional and just engage it if we find such a field.  (And then still probably document some warnings about this, because it'll be in the court of the person authoring the go types to not put themselves in a bad situation.) I'll probably look into doing this tomorrow.

node/bindnode/repr.go
```diff
@@ -53,8 +53,8 @@ func (w *_nodeRepr) Kind() datamodel.Kind {
 		return datamodel.Kind_Map
 	case schema.UnionRepresentation_Kinded:
 		haveIdx, _ := unionMember(w.val)
-		mtyp := w.schemaType.(*schema.TypeUnion).Members()[haveIdx]
-		return mtyp.TypeKind().ActsLike()
+		mtyp := w.schemaType.(*schema.TypeUnion).Members()[haveIdx] // TODO this is fragile: panics with index-out-of-range if the user has nil'd all fields.
```
willscott, line 56: put a `, ok` and an error for out-of-range rather than panic?

node/bindnode/repr.go
```diff
@@ -224,19 +226,57 @@ func (w *_nodeRepr) MapIterator() datamodel.MapIterator {
...
+			}
+		default:
+			panic(fmt.Sprintf("TODO: mapitr.repr for typekind %s", w.schemaType.TypeKind()))
+		}
 	default:
 		panic(fmt.Sprintf("TODO: %T", stg))
 	}
 }
 
+type _mapIteratorRepr _mapIterator
```
willscott, line 244: comment to explain why this wrapping and not just calling w's MapIterator is needed would be helpful

willscott (approved):
the additions here look reasonable

warpfork:
I was too optimistic in saying I'd work on the union thing immediately.  I'm going to extract that to an issue instead.

node/bindnode/repr.go
warpfork: Nontrivial.  This method can't return errors.  It's a core interface method, and a break-the-universe-badly one if we changed its return type.

warpfork:
More tests for this would be desirable, but I'm going to hope we can backfill those in the future.  (And also, it should eventually be easier to do that if we get schema-schema wired up to be data-driven; then we can make purely data-driven language-agnostic tests, and gosh that would be nice.)  Meanwhile, as a stopgap, I promise I've at least tested all these myself in downstreams.

I'm also somewhat attached to these commits already, for the same reason (I've started depending on them by commit hash in a downstream already), so I'm going to boldly merge them, and we can continue improving in future PRs.

mvdan:
All of these fixes make sense to me. I think we should start thinking about more table-driven tests before we continue doing more fixes, otherwise we're just building on quicksand.

PR #238, opened 2021-09-02
Printer feature

warpfork:
I've started developing a `printer` feature, which provides a human-readable debugging output describing a tree of nodes.

This is an answer to https://github.com/ipld/go-ipld-prime/issues/88 .

- :heavy_check_mark: It's reasonably human readable, and rich in information.
- :heavy_check_mark: It can describe typed or untyped data, or a mixture of it.
- :heavy_check_mark: Everything works generically over the `datamodel.Node` interface, and feature-detects `schema.TypedNode` as appropriate.
- :heavy_check_mark: No additional methods are needed from nodes.
- :heavy_check_mark: There are a number of configuration options already (and probably more are desired).
- :non-potable_water: Nongoal: it's not a codec.  It's not intended to be parsable!  (This is important: I want this to provide _more_ information than data model alone, which means it's definitionally _not_ a codec.)
- :shrug: Does not currently log anything about the node implementation type (e.g. bindnode vs basicnode vs custom/codegen'd).  We could add this in the future if it seems useful?
- :shrug: It's not standardized.  (I don't see any reason to aggressively push this to be a cross-language thing.  If it evolves well and seems desirable?  Okay, sure.  But we don't need to start there.)  On the plus side: that also means it's not frozen to this format.

The syntax is somewhat impulsively chosen.  Roughly, it's `kindname{"value"}` (so: `string{"foo"}` and `int{12}` and `bool{true}` etc), or for typed information, `typekindname<TheTypeName>{"value"}`.  You can check out the tests for a few more examples of how it looks.  We could bikeshed this syntax, I suppose, but so far, I'm finding it tolerably readable.

I've started using this in debugging myself already as I work using go-ipld-prime as a downstream user in other projects, and it's saving me tons of time, and also starting to show up in my docs there to explain features to people.  I'm pretty happy about it.  (Also: it means I'm already relatively attached to some of these commit hashes and will try to roll forward and resist a need to rebase if possible.)

In the future, I'd like to add some functions to the root `ipld` package which call these printer features, for rapid easy use and for discoverability, but that's not included in this PR yet.

Review welcome.  This seems like the sort of thing I'd rather merge relatively quickly, and continue iterating on master, though.

### DIFF
--- a/printer/doc.go
+++ b/printer/doc.go
@@ -1,0 +1,94 @@
+/*
+Printer provides features for printing out IPLD nodes and their contained data in a human-readable diagnostic format.
+
+Outputs should look like...
+
+	map{
+		"foo": "bar"
+		"zot": struct<Foobar>{
+			someFieldName: list{
+				0: "this"
+				1: "is untyped"
+				2: int{400}
+				3: bool{true}
+			}
+			otherField: list<ANamedListType>{
+				0: "mind you: 'ANamedListType' is the name of the *list type*."
+				1: "it is not the name of the value types.  those are not actually shown here."
+				2: "you'd have to look at the schema for that information."
+			}
+			moreField: list<[nullable String]>{
+				0: "this is a typed list"
+				1: "but anonymous"
+				2: null
+			}
+		}
+		"frog": map<{String:String}>{
+			"as you have probably imagined": "this is a typed (but anonymous type) map"
+		}
+		"numbers": int{1}
+		"binary": bytes{ABCDEF0123456789}
+		"typed numbers": int<MyNamedTypeInt>{9000}
+		"typed string": string<MyNamedTypeString>{"okay, this one needed some marker prefixes."}
+		"map with typed keys": map<{MyNamedTypeString:MyNamedTypeString}>{
+			"despite being typed": string<MyNamedTypeString>{"map keys still never need prefixes; there's no ambiguity"}
+			"but for the values": string<MyNamedTypeString>{"we still use them"}
+			"well, maybe": string<MyNamedTypeString>{"this might actually be worth debating; it doesn't seem necessary (unless the map value is a variant type, but that's then already addressed by another case)"}
+		}
+		"structs": struct<FooBar>{
+			foo: "do not need to have quoted field names"
+			bar: "because (unlike map keys) their character range is already restricted"
+		}
+		"unit types": unit<TheTypeName>
+		"notice unit types": "have no braces at all, because they have literally no further details.  they're all type info."
+		"variants": variant<TheUnionName>{string<TheInhabitant>{
+			"that was wild, wasn't it.  Check out these double closing braces, coming up, too!  also the string got forced to a new line, even though it usually would've clung closer to its type and kind marker."
+		}}
+		"enums": enum<TheEnumName>{"inhabitant name"}
+		"typed bools": bool<TheBoolName>{true}
+		"map with struct keys": map<{FooBar:String}>{
+			struct<FooBar>{foo:"foo", bar:"bar"}: "that one probably surprised you, didn't it?"
+			struct<FooBar>{foo:"hmmm", bar:"maybe"}: "we might be able to get away without the kind+type marker, actually.  but we need the one-liner struct content printing, at least, for sure."
+		}
+		"map with really wicked keys": map<{WickedNestedUnion:String}>{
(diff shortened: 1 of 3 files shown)


printer/printer_test.go
```diff
@@ -0,0 +1,179 @@
...
+			schema.SpawnString("String"),
+			schema.SpawnStruct("FooBar", []schema.StructField{
+				schema.SpawnStructField("foo", "String", false, false),
+				schema.SpawnStructField("bar", "String", false, false),
+			}, nil),
+		)
+		n := bindnode.Wrap(&FooBar{"x", "y"}, ts.TypeByName("FooBar"))
+		qt.Check(t, Sprint(n), qt.CmpEquals(), wish.Dedent(`
+			struct<FooBar>{
+				foo: string<String>{"x"}
```
rvagg, line 61: `string<String>` is getting weird, I don't suppose it's reasonable to special-case the typed data model kinds like `String` to omit the typing here?

printer/doc.go
```diff
@@ -0,0 +1,94 @@
...
+			"but for the values": string<MyNamedTypeString>{"we still use them"}
+			"well, maybe": string<MyNamedTypeString>{"this might actually be worth debating; it doesn't seem necessary (unless the map value is a variant type, but that's then already addressed by another case)"}
+		}
+		"structs": struct<FooBar>{
+			foo: "do not need to have quoted field names"
+			bar: "because (unlike map keys) their character range is already restricted"
+		}
+		"unit types": unit<TheTypeName>
+		"notice unit types": "have no braces at all, because they have literally no further details.  they're all type info."
+		"variants": variant<TheUnionName>{string<TheInhabitant>{
```
rvagg, line 44: so `variant` is the name for "specific variant of a union"? introduction of new terminology is a concern I reckon, it doesn't read as smoothly if there's disconnect with user-consumed schema lingo

rvagg:
I think I prefer `string{"x"}` to the shorthand version you're proposing in here; I'd advocate dropping that option entirely

aside from comments left inline, I think I like this. I did something a little bit similar in a JS experiment I was working on but this reads much better and you've integrated type information which is great, so I might have to copy this.

printer/doc.go
warpfork: oh oof, I should `s/variant/union/` this.  It's a stretchmark from how I think during development: I was trying out "variant" as an alternate name for "union" and trying to get a feel for if I liked it when seeing and using it for a while.

The result is "yes", I do kinda come to prefer that term.  I think maybe we should consider doing a mass rename of "union" to "variant".

But that's not a discussion I want to force through in this PR; it deserves its own space.

printer/printer_test.go
warpfork: Yeah I'm on the fence on this.  At first I thought I'd strip it, as you suggest.  Then I looked at the printout the first time I got one to come out, and was like... hum: typed vs untyped string seems like the kind of distinction that could actually trip someone up someday, so maybe that's useful.

I'm _still_ on the fence, a couple days later.  I can't think of _how_ typed-vs-untyped string would trip someone up, exactly, either.  It's just a nagging fear.

Maybe a config option for "`SmartElidePreludeTypeInfo bool`"?

warpfork: I think I'd propose we leave it in verbose mode for now, someday add that config flag, and then if enough people are perturbed by the verbosity we flip the default of that flag to true.  We should probably get some impressions from people further out from us before deciding to elide things that might be obvious to us but nonobvious to new users, is my thinking.

warpfork:
:raised_hands: I'm thrilled you like it.

And you're right, I think `string{"x"}` consistently is a good idea too now.  I'll remove the references to the shorthand idea.

printer/printer_test.go
warpfork: Added a config placeholder for this and attached documentation for it so someone looking for such a feature can at least find that we've considered it.  Haven't implemented it immediately, however.

warpfork:
Though this contains some TODOs, I'm going to go ahead and merge it.  Since this seems directionally acceptable, it should be no problem to add the remaining features in subsequent PRs.

(I've gotten myself in the awkward position of wanting both this, and some other features recently landed, at the same time, in a downstream, and yet don't have the time block today to finish the remaining feature wishlist in this diff.  So: onward!)

PR #239, opened 2021-09-02
bindnode: fix for stringjoin struct emission when first field is the empty string

warpfork:
In the case of an empty string as the first field, the buffer length
is not a valid proxy for whether we're on the first field or not.

This means if we have some type like:
`type Foo struct {a String; b String} representation stringjoin(":")`,
and the value of it is `{"", "b"}`, then the string of that
should still be ":b".

Before this fix, it would incorrectly be emitted as "b" (no joiner),
which would not round-trip.

### DIFF
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -364,6 +364,7 @@ func (w *_nodeRepr) AsString() (string, error) {
 	case schema.StructRepresentation_Stringjoin:
 		var b strings.Builder
 		itr := (*_node)(w).MapIterator()
+		first := true
 		for !itr.Done() {
 			_, v, err := itr.Next()
 			if err != nil {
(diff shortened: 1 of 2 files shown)


rvagg (approved):
you know you want to add a test for this

mvdan (approved):
Nice catch, oops.

FYI Rod, with Eric we've been discussing testing. One half of it will be extending the schema test suite, and another half will be adding table-driven bindnode tests for anything that's specific to how it handles Go types.

In this particular case, I agree with Rod: this could just be an added test in the schema test suite, so there's no reason to delay it.

warpfork:
You're both right, this was easily added to the existing schema test suite, so I was indefensibly lazy not to add a regression test.  Fixed.

(I yearn to get to data-driven tests that we can put in the ipld meta repo, but that's no excuse for not doing the best we can along the way until we get there.)

Issue #237, opened 2021-09-02
selector for limiting recursion over a field


willscott:
In cases of a dag in a git-like dag
<img width="425" alt="image" src="https://user-images.githubusercontent.com/428199/131820915-72f05b88-7a57-47a7-a8bf-12720f430b45.png">

It would be useful to have a selector that's able to get the chain (the nodes labeled 'vn', 'vn-1', etc), including the CIDs of the roots pointed to in each object, but not pulling in the full data dags linked in each.

This can be mostly achieved today with a combination of the field-recursion and stop-at, but it seems like that might skip including the CID of the link to the root as well.

warpfork:
This might seem like kind of a low-brow question, but... Can't you just obtain the desired effect by listing the fields you *do* want to recurse over, and not the ones that you... don't?

I ask because -- while I'm also not going to say the proposal in https://github.com/ipld/go-ipld-prime/pull/240 is nonviable -- I think you'll get better performance out of saying what you want, rather than asking the selector to look at everything and then using further logic to toss things back out of consideration again.  In other words, even if we had both these features -- I'd probably council someone to use the simpler more direct one unless they're really, really sure they can't for some reason.

warpfork:
In the git example, the recursion you describe wanting keeps following "parent" fields, and just never askes to follow a "tree" field, and... that's it, right?

I think this is pretty common in most protocols I've seen -- there's some fields that we want to walk, that point to more data (often, data of the same "type", not that we explicitly know about that in selectors), and there's some we just don't ever want to walk (because it's going to point to different "types" of stuff, e.g. file trees in git instead of commits).

So if we have an example where that logic doesn't hold, it'd be kinda fun to look at it in a bit more detail as a case study.  I'd probably want to attach those case study notes to docs somewhere for "why would you want to use this feature".

rvagg:
closing this for now but this can be picked up by anyone who feels this is really important and useful and wants to do an implementation because current options are unsatisfactory for various reasons (it'll also probably be important to clarify _why_ as per the comments above)

PR #240, opened 2021-09-03
Add bypass condition to ExploreRecursive selector

adlrocha:
This PR implements https://github.com/ipld/go-ipld-prime/issues/237

The idea is to have a `bypass` condition in `ExploreRecursive` selector to specify fields that shouldn't be recursed deeper (similar to the `RecursionLimit` condition but for specific fields).

Thus, the EploreRecursive selector turns into:
```js
type ExploreRecursive struct {
	sequence Selector (rename ":>")
	limit RecursionLimit (rename "l")
	stopAt optional Condition (rename "!") 
        bypass optional Condition (rename ":x")
}
```

And the representation would be like this:
```
{
                "R": {
                        "l": {
                                "none": 0
                        },
                        ":>": {
                                "a": {
                                        ">": {
                                                "@": {}
                                        }
                                }
                        },
                        "!": {
                                "/": {
                                       "/":"bafyreifepiu23okq5zuyvyhsoiazv2icw2van3s7ko6d3ixl5jx2yj2yhu"
                                }
                        },
                        ":x":  {
                               "hasField": "field_that_shouldn't_be_recursed",
                         }
                }
}
```
//cc @willscott

### DIFF
--- a/traversal/selector/condition.go
+++ b/traversal/selector/condition.go
@@ -21,7 +21,8 @@ type Condition struct {
 type ConditionMode string
 
 const (
-	ConditionMode_Link ConditionMode = "/"
+	ConditionMode_Link     ConditionMode = "/"
+	ConditionMode_HasField ConditionMode = "hasField"
 )
 
 // Match decides if a given datamodel.Node matches the condition.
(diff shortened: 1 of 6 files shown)


warpfork:
If we do choose to go forward with this -- do you think I could beg you to contribute test fixtures in data-driven form over into https://github.com/ipld/ipld/tree/master/specs/selectors/fixtures ?

I see you already added tests here, and that's good.  I'd like pure data-driven tests that can live in the specs&meta repo even more, though.  I find the data-driven tests easier to read and review.  It'll be more helpful to people implementing selectors in other languages.  And it's closer to being documentation for end users.  (It's also... probably actually *easier* to write the data-driven tests now than this kludgey golang stuff!)

Happy to walk through how to bump the relationship between those repos to make those kind of tests fly, if it's helpful.  (It might be nonobvious; and if it is, we haven't written docs for that workflow yet.)

adlrocha:
> If we do choose to go forward with this -- do you think I could beg you to contribute test fixtures in data-driven form over into https://github.com/ipld/ipld/tree/master/specs/selectors/fixtures ?

Sure! I'll go through the docs see if I am able to do it myself. If not I may ask for a quick walkthrough.

PR #241, opened 2021-09-04
codec: make cbor and json codecs use ErrUnexpectedEOF

mvdan:
(see commit message)

### DIFF
--- a/codec/dagcbor/unmarshal.go
+++ b/codec/dagcbor/unmarshal.go
@@ -71,6 +71,9 @@ func Unmarshal(na datamodel.NodeAssembler, tokSrc shared.TokenSource, options De
 func unmarshal1(na datamodel.NodeAssembler, tokSrc shared.TokenSource, gas *int, options DecodeOptions) error {
 	var tk tok.Token
 	done, err := tokSrc.Step(&tk)
+	if err == io.EOF {
+		return io.ErrUnexpectedEOF
+	}
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


rvagg (approved):

warpfork:
I think I'm probably mostly happy with this, but do also want to walk through one scenario:

What would I do if want to:
1. open a file
2. hand it to unmarshal
3. and then be able to clearly and calmly accept if the file existed and was zero-length?

Note that this is nontrivial because it's kind of a pain to check if the file is zero length or not without pre-reading it.

As a result, I think in this story, I would want to be able to differentiate between "the file has zero bytes" and "the file contained one byte and it was '{'".  One of those is much less valid than the other.  Can I differentiate those?

mvdan:
I think the answer there would be rich errors. Something like:

```
package codec

type SyntaxError struct {
    Offset int64
    Error error
    // more fields if we want, perhaps what codec was used
}

func (e *SyntaxError) Error() string {
    return fmt.Sprintf("syntax error at byte %d: %v", e.Offset, e.Error)
}
func (e *SyntaxError) Unwrap() error {
    return e.Error
}
```

Then, in your code:

```
err := Decode(...)
if stxErr := (codec.SyntaxError{}); errors.As(err, &stxErr) {
    if stxErr.Offset == 0 && errors.Is(stxErr.Error, io.ErrUnexpectedEOF) {
        // you hit an empty file here
    }
}
```

You could expose helpers for stuff like this, such as `func IsEmptyInputError(err error) bool`, but I lean towards thinking that's overkill. Special casing this feels rare.

warpfork (approved):

mvdan:
Thanks for the reviews! I've removed the raw codec TODO, since Rod confirmed it should support decoding zero bytes. That's a small comment removal, so I don't think we need a re-approval.

PR #245, opened 2021-09-07
WIP: improved schema-schema support

mvdan:
DO NOT MERGE


mvdan:
For clarity: this is blocked because the schema-schema definition uses inline type definitions in places like map keys, and the current schema package doesn't support those, as far as I can see.

There's also another edge case: I did not see how to spawn a map type while specifying a custom representation strategy, which the schema DMT seems to support.

mvdan:
I deleted schema2 just to remind myself not to use it btw, but that probably does not belong in this PR.

warpfork:
I'm probably going to have to review this in rounds, so forgive me as the comments trickle in :)

> If a TypeSystem remembered the order in which type names were accumulated,

+1, I 100% want this, I consider it a bug that I didn't make it do that the first time around.  Stable iteration order is always, always, always good for sanity.

> can probably merge this before adding missing features in the schema package, by keeping the Load API unfinished for now

:heavy_check_mark: If we can, we probably should.

> I did not see how to spawn a map type while specifying a custom representation strategy

Yep, I probably missed this on my last go.  The absense is a bug.

> blocked because the schema-schema definition uses inline type definitions in places like map keys

Yes.  I've become pretty fully convinced we should move the schema-schema to not use the inline strategy for unions, and use keyed unions instead.  That will unblock us in addition for being good for {{laundry list of subtle reasons}}.

... Wait.  I read that again.  You mean the situation around InlineDefn.

> blocked because the schema-schema definition uses inline type definitions in places like map keys

Yeah there's also some currently unwritten code around inline type definitions.

I think I had some work in schema2 about TypeReference vs TypeName and some third type, which seemed like it was going the right direction for clarifying some of the issues around identifying InlineDefn types, at least.  Might be worth grabbing that back out of the bin (even if much of the rest can still head into the bin).

On the other hand: I think we could route around this, too.  We could offer a variant of the schema-schema simply doesn't use the inline type definitions feature.  (This would be surprisingly neat, because we'd only be naming more types in the schema-schema:  the actual shape of document they'll parse would remain *exactly the same*.)  This would seem reasonable to do, in light of acknowledging "the schema-schema should be something you can parse relatively early in the journey to bootstrap an IPLD Schemas implementation in a new language" as a goal.

(Maybe worth noting, maybe a tangent: I've also recently realized that the idea of having types sometimes have a TypeReference instead of TypeName (e.g. be able to include spaces, braces, etc) is sometimes a recurring bugaboo.  With codegen in go, we had to invent the idea of using double-underscores as delimiters (and honestly, it still probably gets ambiguous at some point).  I've just recently discovered this _again_ when trying to make some Starlark bindings to IPLD data.  It's... a sufficiently recurring topic that it might be worth standardizing how we do that munge.)

node/bindnode/infer.go
```diff
@@ -108,6 +112,124 @@ func inferGoType(typ schema.Type) reflect.Type {
...
+			fmt.Fprintf(&b, "%s ", fieldNameFromSchema(string(ftyp.Name())))
+			fmt.Fprintf(&b, "*%s\n", ftypGo)
+		}
+		fmt.Fprintf(&b, "\n}")
+		return b.String()
+	}
+	panic(fmt.Sprintf("%T\n", typ))
+}
+
+func ProduceGoTypes(w io.Writer, schemaTypes map[schema.TypeName]schema.Type) error {
```
warpfork, line 203: oh, WOW!  Minicodegen!  Sweet!  I had been thinking this would be maybe kinda neat to have in our back pockets, and now here it is!

I wonder if it should be shifted to a subpackage?  Obviously it's not mechanically necessary to do so, but also it seems like a sufficiently different kind of feature than all the other publicly exported functions of this package that it might be worth it for organization and messaging.

mvdan: I wrote it in a couple of hours because I'm too lazy to spend the two hours instead writing the types by hand :)

I was also thinking about where to put this. Ideally it would be a program, like `node/bindnode/bindnodegen` or something like that. I don't think that's particularly useful until we have a "parse schema file" library API, though. For now, I think we can just put this behind a `// +build bindnodegen` to not expose it in the bindnode API, but still allow accessing it form other packages like:

```
//go:generate go test -run TestGenerate -tags bindnodegen
```

mvdan:
> heavy_check_mark If we can, we probably should.

Should be fairly easy to add to TypeSystem. The only big question in my mind is whether spending time on the schema package is worthwhile, which might not be if we want to replace it with schema2.

Maybe we can marry schema and schema2 this week? :)

schema/dmt/build.go
```diff
@@ -0,0 +1,71 @@
+package schemadmt
+
+import (
+	"fmt"
+
+	"github.com/ipld/go-ipld-prime/schema"
+)
+
+func Load(ts *schema.TypeSystem, node *Schema) error {
```
warpfork, line 9: So this function is the big crux where things connect -- it mutates the given `*schema.TypeSystem` pointer.

Just gonna talk my way through what the design approach is here: The way this function is approaching things is: we've got this `schemadmt` package, and we're still expecting the `schema` package to expose the `Spawn*` functions.  The dependency arrow order is `schemadmt -> schema`, and that's it.  Users would have to import the `schemadmt` package, use it to parse and load the data, and use this function to compile it.  And we've left all the compilation validation logic in the `schema` package.  (Implicitly, it is still very possible right now to create `schema.*` values that are nonsensical, and the interfaces don't stop you at all.)  And the amount of mostly-repetitive hand-wrought golang types and accessor methods in the `schema` package remains high.  (But, in this approach, since it's using bindnode instead of my heavy-duty (and I don't mean that in a good way) codegen, we don't have any of those accessor methods on the dmt types, so the repetitivity level, as it manifests in SLOC, is a lot lower.)  Okay.  Cool.

warpfork: There are a couple small things I'd probably change, just at the level of how we call this:

- I'd call this `Compile`.  It's going from DMT (something pretty legible, transcodable, etc) to the form _used_ to do work, which is not reversable back to DMT.  An operation that goes from legibility, irreversibly, to illegible-but-useful sounds like a "compile" operation to me.

- Not sure if thrilled with the mutate-the-argument strategy.  My first inclination would be to have it look more like a constructor and return a new `schema.TypeSystem`.  I don't think we have a lot of reasons to think of appending a TypeSystem repeatedly.  (Like, even if it's currently mechanically possible, A) I don't think it's a good API contract for TypeSystem to be treated as mutable, and B) just... why?  It's something one would only be able to do by writing Interesting golang code.  When we think of how we'll have schema-driven tooling work purely by APIs (think: IPLD Explorer with a Schema parameter), none of those Interesting ways one could compose golang code will be relevant, so istm it's somewhere between not useful and mildly bad for us to have API features that won't make sense when we go data-driven.)

mvdan: I totally made up the name Load, fully expecting we'd replace it :) Compile sounds fine. I also thought of others like Accumulate or Spawn.

Regarding append to vs return a new TypeSystem: I thought I'd leave the API being more flexible, just like Accumulate. I imagine it could be reasonable to use a combination of two schemas at once, for example. If that's something we never want to allow, then SGTM to return a new typesystem.

schema/schema2/typesystem.go
```diff
@@ -1,441 +0,0 @@
...
-	//  Any type that has some kind of recursion (maps, lists, structs, unions, links with target type info) causes a lookup to see if the referenced types exist.
-	//   FUTURE: we'll need most recursives to additionally do some checks for correct composition of nullability after the introduction of unit types.
-	//  We manage to avoid the need for a two-pass system because we just give all reified types a pointer to the typesystem aggregate;
-	//   this means each of them still just stores the type names, and if asked for another type pointer, looks it up on the fly (by which time it's available).
-	//  Some kinds of types involve other more specific checks,
-	//   such as maps verifying that their keys are stringable (which is a rule we enforce for reasons relating to pathing),
-	//   and unions verifying that all their discriminant tables are complete (which is a rule that's necessary for sanity!),
-	//   and etc.
-	typesdmt := schdmt.FieldTypes()
-	for itr := typesdmt.Iterator(); !itr.Done(); {
```
warpfork, line 37: I know you toss out most of this package immediately for clarity, but of the things worth dredging out of the ruins: this, and the code below, is probably actually good.  I think this is more complete validation logic than we've had anywhere else.

I guess with the approach taken in this PR, this logic should move into the `schema.ValidateGraph` function, probably.

schema/dmt/build.go
mvdan: Your understanding of the dependency graph is right. That's what I arrived at with my paper sketches some weeks ago. When we end up with a fully end-to-end Parse API that takes a file and gives you a TypeSystem, that will probably live in schemadmt, or a package that imports schemadmt.

And you're right that there's still glue code to walk the "AST" nodes and spawn schema types. But like you say, this is pretty painless with "just Go types" :)

schema/schema2/typesystem.go
mvdan: probably best if we do this (and remove schema2) in a PR before or after this one. I can undo this removal for now.

schema/schema2/typesystem_test.go
```diff
@@ -1,212 +0,0 @@
...
-	"strings"
-	"testing"
-
-	. "github.com/warpfork/go-wish"
-
-	"github.com/ipld/go-ipld-prime/codec/dagjson"
-	schemadmt "github.com/ipld/go-ipld-prime/schema/dmt"
-)
-
-func TestBuildTypeSystem(t *testing.T) {
```
warpfork, line 14: Similar to [above](https://github.com/ipld/go-ipld-prime/pull/245/files#r704275918), I'd like to flag these tests as one of few things still worth dredging out of here.

There's not much to them, but it's some of the only tests we've got so far which test the logical validation rules.  (Also, test that things can flunk them!)  Even if the coverage of these is skeletal so far, it would be nice to make sure we know where we're going to put things like this.

(Or, better yet, if we can port them to data-driven stuff now, and maybe put them in the ipld/ipld meta repo, that'd be..... really awesome.)

schema/schema2/typesystem.go
warpfork: Not complaining about bold deletions (deleting code is good!), just plopping some surveyor flags for stuff that might be worth keeping.  (Because most of it, indeed, is _not_, heh.)

node/bindnode/infer.go
warpfork: > I wrote it in a couple of hours because I'm too lazy to spend the two hours instead writing the types by hand :)

:raised_hands: 
Sounds like it was extremely worth it.

I'm super happy about this; it'll make this immediately less of a pain to maintain if we tweak the schema-schema.

> node/bindnode/bindnodegen

I advocate you do this immediately :)

Even if we don't wrap a full CLI around it today, I can already think of another project where I'd very likely use this even in its nascent form.

warpfork:
> The only big question in my mind is whether spending time on the schema package is worthwhile, which might not be if we want to replace it with schema2.
> 
> Maybe we can marry schema and schema2 this week? :)

Yeah!

I think maybe we should do some brief phase of "readme driven development" at the beginning of that, and make sure we know what our goalposts are.

I guess the biggest questions are how aggressive we want to be about fixing things along the way.  I don't think there's a lot from `schema2` we can take (`schema2` was trying to work by refering to the DMT, and we're going basically the opposite direction now, so none of the `schema2.Type*` stuff is recyclable at all).  The tests and the validate logic that I flagged are worth porting and that's about it.  But if there are things we want to fix about `schema`, now might be a time to consider at least some of them, too.

ISTM we're settling around:

- Yes, the `schema` package has its own manually written types, and accessors.  We're just accepting this.
  - Upside: the `schema` package still has zero dependencies.  Probably a good thing.
  - Downside: maintainability.  A fair amount of manual grind here.
    - Especially: manually maintaining map ordering, anywhere that comes up and could affect determinism :sob: 
- The `schema` package is still structs, not interfaces.
  - Means we'll never be able to do the thing where a sufficiently fancy node package can use the same value for Type and Prototype.  (Oh well!  We already have `TypedPrototype` to at least make it clear how to step from the latter to the former, and that solves most problems.)
  - We probably could get aggressive enough to change this.  Not sure if it feels compellingly important anymore, though.
- The `schema` package has some guardrails, but not many.
  - The `schema` package still has the bare constructor methods, which allow people to create values of `schema.Type*` that are logically broken.  We give up on preventing that and will go with the "we're all consenting adults here" reasoning.
  - Assuming we keep things mostly as-is: all the `schema.Type*` types are mostly immutable, once spawned, at least.
  - wish: I'd like to make `TypeSystem` immutable.  (It's not right now, but that kinda flusters me.)  Then *force* things to go through a Validate step before returning a `TypeSystem`.
  - Optionally, if making `TypeSystem` immutable: introduce a quick simple `TypeSystemBuilder`.  Gives us a clear path to port any existing code using "accumulate" style, for whatever that's worth.
- We're doing validation in the `schema` package, and not trying to have validation work directly on data in `schemadmt`.
  - So if someone wanted to write a schema validator tool, it's gonna fully compile the thing, and flip all the data through both packages.  Well, okay.  Fine.
  - (I have vague memories of finding some error messages were harder to phrase usefully when I had to detect the error farther from the DMT form, but maybe that was in my head, and we'll find it surmountable if we try again.)

Does that sound about right?

I've got wishlist items in there, but it also seems like a mostly conservative approach is possible.

The thing that remains the most frustrating to me is the `schema` package still having so many of its own manually written types.    I can't really see a good way around this.  But it would certainly be nice if we could make it easier to maintain, so maybe it's worth thinking about that for a moment.  Thinking out loud:

- It's somewhat frustrating because the types in `schema` feel *very* similar to the ones in `schemadmt`.  It's just that they're... not.  (One reason is because the ones in `schema` are mostly aiming to be immutable, and thus have unexported fields, whereas `schemadmt`, because it's bindnode, is required to have the opposite: all exported fields.  The other reason is that the `schema.Type*` structures have pointers back up to `schema.TypeSystem` (and sometimes other precomputed indexes that wouldn't be in the DMT form).)  So: manual work appears.
- We were reminded that TypeSystem should remember order, because stability is good.  _This comes up in a variety of places in the schema-schema_; there are lots of definitions that occur inside maps and stability would be good.  The DMT package, because it's Nodes, and Bindnode (and the new supporting minicodegen) supports that properly, naturally retains order.  But we can't use it in the `schema` package because import cycles.  Is there anything we can do to make it easier for the `schema` package to have order-maintaining maps and also be less of a PITA to maintain?
- I've found it's easy to let typos creep into the `schema.*` APIs when they're manually written.  (Especially around capitalization: is it `Stringjoin` or `StringJoin` when translating the concept (which is all lowercase, in the IPLD Schema DSL) to golang?)  This can be addressed with "just be more diligent" but it's worth noting as, well, something we have to be diligent about when there's all this near-duplicated code.
- Is there a possible outcome, somehow, where the majority of the `schema` package is also pretty much generated?  It would be neat if we could get it set up so that if we added something to the schema-schema, the feature instantly becomes described by the `schema` package.  I think we'd always have some more accessor methods to write manually, but even keeping the type skeletons shipshape by codegen would possibly be worth it.  Maybe?

mvdan:
> Especially: manually maintaining map ordering, anywhere that comes up and could affect determinism sob

Reminder that generics will likely make this less painful soon(ish).

> The `schema` package is still structs, not interfaces.

I think this is fine. `go/types` is mostly structs, too.

> The `schema` package has some guardrails, but not many.

I think this is fine in practice - when we finish the end-to-end API, very few users would choose to stick with creating/filling typesystems directly. But I'd be in favour of deprecating `ValidateGraph` in favor of something like `Finalize`, which validates and forbids any further modification.

> We're doing validation in the `schema` package, and not trying to have validation work directly on data in `schemadmt`.

We could always add validation at the DMT level, if we think that would give us better error messages. Another option would be to tell `TypeSystem.Accumulate` a position, so that errors could point back to input line positions. All of this seems wishlist-like to me, though.

> I've got wishlist items in there, but it also seems like a mostly conservative approach is possible.

Yes. We don't want to break users unless we really want to revamp how the schema package works. And I don't want to add another month's worth of work to this already-large chunk of work to get `schemadmt.Parse`.

> Is there anything we can do to make it easier for the `schema` package to have order-maintaining maps and also be less of a PITA to maintain?

Generics :) And bindnode will hopefully use those too.

> The thing that remains the most frustrating to me is the `schema` package still having so many of its own manually written types.

On one hand, yes, it's repetitive. On the other, I do think separate types are worthwhile. Having a TypeStruct at the DMT/AST level is very different from having it at the TypeSystem level. And, like you say, the TypeSystem level needs to attach extra information to nodes, as well as avoiding some forms of mutability.

I don't have a strong opinion here. But I do think that optimizing for maintainability is probably the wrong target. Hopefully we'll get to a point where the schema-schema is mostly stable, and we have two or three more orders of magnitude of IPLD users than IPLD developers. At that point, a thousand lines of repetitive code is insignificant when compared to nicer APIs :)

warpfork:
Alright.  I'm convinced on all fronts :)

mvdan:
Alright, I'll undo the schema2 removal and clean this up to get it green.

mvdan:
@warpfork this should be ready. Renamed Load to Compile, hid bindnode.ProduceGoTypes behind a build tag, organized the code a bit better, added some TODOs. And wrote a nice commit message, I think.

mvdan:
Oh, and of course I disabled the last part of the "schema parse" test that was using Compile, as that's unfinished and just errors right now.

schema/dmt/schema.go
```diff
@@ -261,7 +272,7 @@ func main() {
 	ts.Accumulate(schema.SpawnString("FieldName"))
 	ts.Accumulate(schema.SpawnStruct("StructField",
 		[]schema.StructField{
-			schema.SpawnStructField("type", "TypeNameOrInlineDefn", false, false),
+			schema.SpawnStructField("type", "TypeTerm", false, false),
```
warpfork, line 275: I think that I actually wanted to carry the rename `s/TypeTerm/TypeNameOrInlineDefn/g` back _up_ to the schema-schema.  But I haven't done it yet, so if you'd rather go for consistency for today, fair 'nuff.

mvdan: Yep. Renames should go there first :) Otherwise it's going to be easy to get out of sync and forget. The source of truth is there, not here.

warpfork (approved):
I still kinda want the `bindnodegen` as a package asap.  I might very well use that later _today_ if it is :angel: 

But yeah, let's gooo :shipit:

PR #247, opened 2021-09-08
schema: keep TypeSystem names ordered

mvdan:
(see commit message)

### DIFF
--- a/schema/tmpBuilders.go
+++ b/schema/tmpBuilders.go
@@ -139,13 +139,21 @@ func (ts *TypeSystem) Init() {
 }
 func (ts *TypeSystem) Accumulate(typ Type) {
 	typ._Type(ts)
-	ts.namedTypes[typ.Name()] = typ
+	name := typ.Name()
+	if _, ok := ts.namedTypes[name]; ok {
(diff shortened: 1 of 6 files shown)


schema/tmpBuilders.go
```diff
@@ -139,14 +139,23 @@ func (ts *TypeSystem) Init() {
 }
 func (ts *TypeSystem) Accumulate(typ Type) {
 	typ._Type(ts)
-	ts.namedTypes[typ.Name()] = typ
+	name := typ.Name()
+	if _, ok := ts.namedTypes[name]; ok {
+		// Should we error or panic here?
+		// We'll overwrite a type, and end up with duplicate names.
+	}
```
warpfork, line 146: I think a panic is fine here.  This is not an interface that we're expecting to be heavily directly used by the time we've finished all the parsing features we want to have.

warpfork (approved):
LGTM!

I'll be very happy to have this rough spot fixed.  Stability of order is very nice to have.

schema/tmpBuilders.go
mvdan: alright, will do

PR #250, opened 2021-09-13
schema: turn TypeName into an alias

mvdan:
(see commit message)

### DIFF
--- a/fluent/qp/qp.go
+++ b/fluent/qp/qp.go
@@ -182,7 +182,7 @@ type linkParam struct {
 }
 
 func (l linkParam) Assemble(na datamodel.NodeAssembler) {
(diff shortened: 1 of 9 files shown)


warpfork:
I _guess_ I'm fine with this?

The commit message walks through the consequences and tradeoffs accurately.

I guess I'm unfamiliar with whether it's nowadays considered a golang norm to use aliases like this.  But no objections.  And it's not like we've been shy putting the alias features to use elsewhere, so, that's certainly no barrier!

mvdan:
It is not a very common norm to use type aliases this way, but I've definitely seen it in a bunch of places :) I'm hoping it will gain more traction to, for example, make `func() map[string]string` more self-descriptive.

PR #251, opened 2021-09-14
Don't abort block processing when encountering SkipMe

rvagg:
Loading a link from within a block and getting a SkipMe back from the loader
will `return` out of the node iteration. If using SkipMe to deal with block
load memoisation, this can lead to incomplete traversal by avoiding any
potential non-skippable links that follow a skipped one.

_(still needs a test or two)_


warpfork (approved):
Yep, that looks right

rvagg:
@warpfork I'm cringing at how much this is going to grind your gears, but I've pushed a couple more commits, including tests for `SkipMe`, but along the way I found a codec-related inconsistency that I needed to fix in order to make a more complex DAG to walk..

The fixtures from focus_test.go use blocks that have been `encode()`d with dag-json, so they have the lexical sorting applied. _But_ `rootNode` is used in its original assembled form in most places, even though we `encode()` it and use the link that results from that. So, when I try to use `rootNodeLnk` in a new parent Node, I get inconsistent results doing a walk through its contents compared to if I walk through its contents using `rootNode`.

There's a few ways to deal with this obviously, the one I went for was:

1. do a full round-trip with `encode()` to ensure stability
2. do it with a custom JSON encoder that removes lexical sorting, so we don't have to touch many of the existing tests and the field ordering, where it's being checked, looks like the ordering during construction
3. rework the newish chain walk tests (which were added _after_ the dag-json lexical sorting fix) to have more explicit visit order checking

Another option was to just go with dag-json and rewrite the order dependent assertions and add some docs. I can back up and do that if you think that's preferable—the custom encoder is already pretty weird so maybe it'll be less weird to just fix up the ordering and explain why they are not in the original construction order.

Added walk tests for a few different selectors, including the manually constructed one used commonly in Filecoin and a few other places.

warpfork:
Yeah, this does grind my gears.

We should not delete tests that test that order is maintained in the data model.  The reasoning for this is straightforward:

1. Selectors should work over the Data Model; and as another way of saying the same thing, should work over all codecs.
2. Not all codecs, and certainly not the Data Model, has this sorted order idea.
3. Therefore it would be incorrect to delete tests that just because they aren't representable in, for example, round-tripped dag-json (which is an "incomplete" codec, as it does not successfully round-trip the total set of information the Data Model can describe).

This means I don't think we should do the encoding round-tripping, either.

(This is a problem we could've continued to just not have.  Ya'll wanted to walk face-first into it... ┐_(ツ)_┌  This is the kind of stuff we have to deal with, all the dang time now, if dag-json is going to be incomplete on ordering grounds.  We'll probably eventually want to... make another codec that doesn't have that problem, to avoid the sheer irritation of it all, I suspect.  Meanwhile, we also have to support all this stuff unambiguously if only for the sheer reason we _also_ still decode un-sorted dag-json data in the wild without complaint.  The worst of all worlds!  Who could've possibly seen this coming!)

rvagg:
k, I've reverted the special codec and encode round-trip, leaving testing of a non-encoded node intact.

I accept your disgruntlement with the codec shuffling, but this is also the world of codecs we swim in, it's going to be the case with all current codecs we deal with, especially the ones with wide real-world use - dag-pb, git, bitcoin, eth, etc. They all have special rules about how the bytes should be laid out, some stricter than others. So creating `Node` forms of them without passing through an encoder is going to have this same problem of difference of ordering. It's just a reality we have to accept and make sure we (a) test for and (b) educate users about in cases like traversal where the rubber hits the road in surprising ways.

One of the things I experimented with in JS, for dag-pb at least, was a `prepare()` function that would take your in-memory form and ensure that it matches what it'll look like after a round-trip, so if that's important then you have the ability to do it.

dag-jose has a really weird version of this where it'll accept a "compact" form which is just a single colon-delimited string that can be expanded out to be the full form. The codec takes both this compact and the full form but only encodes the full form. So a round-trip can get really surprising in that case.

This is the world of messy data and opinionated layers.

If you want a codec that preserves order then we should just do that. Or, just lean on lists to make the problem disappear entirely, `listpairs` is quite a nice abstraction for this.

rvagg:
Includes a working version of #252 with tests for it doubling what I already have for `SkipMe`.

warpfork (approved):
Nice tests.  :heavy_check_mark:

PR #254, opened 2021-09-17
Add hasField condition

adlrocha:
As part of this PR https://github.com/ipld/go-ipld-prime/pull/240 I implemented the `hasField` selector. This condition is part of the [spec](https://ipld.io/specs/selectors/) but hasn't been implemented yet. As the aforementioned PR is still frozen until we figure out if we really need a `bypass` condition (and it may not be ever merged), I thought it would be a good idea to at still get the `hasField` condition merged.

### DIFF
--- a/traversal/selector/condition.go
+++ b/traversal/selector/condition.go
@@ -21,7 +21,8 @@ type Condition struct {
 type ConditionMode string
 
 const (
-	ConditionMode_Link ConditionMode = "/"
+	ConditionMode_Link     ConditionMode = "/"
+	ConditionMode_HasField ConditionMode = "hasField"
 )
 
 // Match decides if a given datamodel.Node matches the condition.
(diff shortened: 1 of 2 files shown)


willscott (commented):
looks good to me

warpfork:
Yep, no complaints from me either.

Other than perhaps it would be more ideal if tests start coming in via the `ipld/ipld` repo, so that they're cross language.  (And also a lot more legible.  I honestly can't really eyeball our golang tests very well.  It's lot less well-forged than the data-driven system (and it's only around because it's older; someday I'd like to hammer it right out of existence, probably).)

warpfork (approved):

BigLep:
@adlrocha : do you want this?  If so, can you please rebase and please see the latest developments in IPLD test fixtures.

adlrocha:
Hey, @BigLep. I think the team ended up implementing the feature for which we needed `hasField` with an alternative approach, so we can probably close this PR as it no longer applies. Feel free to re-open if this is not the case. Thanks!

Issue #236, opened 2021-08-26
refactor selector package layout


warpfork:
The current `traversal/selector` package works, but I think at this point, after seeing how users interact with it (and what confuses them), we could make some changes that would make it auto-explain itself to users much better.

## Problem Identification

The main problem with the current arrangement is that it makes it too easy for a new reader to think the `selector.Selector` type is something they should interact with... which is *not* the case.

The `selector.Selector` type is a "complied" thing that's used by other logic to get stuff done; it's often not the data type a user should be doing anything but pass around.  Users shouldn't be implementing it; users shouldn't be calling its methods directly themselves; it's not meant to be part of a protocol API itself (its precursors, before "compilation", are!); etc.  And yet people see it first, so they keep assuming it will be those things.

## Proposed Solution

It would be better to have:

- `traversal/selector/engine` -- contains the `engine.Selector` interface, all its implementors, and the `Compile` functions
- `traversal/selector` -- contains the user-facing walk functions, and user-facing `Compile` functions (it mostly delegates)
- `traversal/selector/parse` -- contains things like the json-to-dmt jumpers
- `traversal/selector/builder` -- sure, this can come along too

The packages like `selector/parse` and `selector/builder` should just be returning an `ipld.Node` of the DMT of the Selector.

The whole `traversal/selector.*` namespace can be filled with things a typical user *should* see at first glance.  Like: functions that glue the `parse` or `builder` features directly the `engine.Compile` and then pass all that straight through to starting a `Walk`.  Having complete end-to-end features like that visible in the root package about selectors should increase ease-of-use and discoverability drastically.

The `traversal/selector` package also becomes free to import other packages like `dagjson` without fear, and free to use those to make nice one-step helper functions... because if some user is really dependency-conscious and doesn't want that, they are free to do the additional work to make their code use the low level subpackages directly.  This frees us up to be much more helpful to the most typical users.

The huge majority of concrete types in the current selector package would all disappear into `selector/engine` and I think that would be for the best.  Users need to be aware of those approximately "never".

## Timeline and Change Strategy

Not immediate.  At this time, I'm just making notes on the wish.

This would probably result in some API changes that are technically "breaking", so we'll want to decide when it's time for that.  That said: the breakages I would expect to make will be extremely clear at compile time and almost certainly addressable by "sed"-scale updates in downstream consumers.

[...]

Issue #249, opened 2021-09-12
Common selectors are in the wrong place


willscott:
https://github.com/ipld/go-ipld-prime/blob/master/traversal/selector/parse/selector_parse.go#L66-L94

Why are they in the `parse` package? They should be in `traversal` or `traversal/selector`, and not there.

warpfork:
Yeah.  https://github.com/ipld/go-ipld-prime/issues/236 would be how I want to solve this.

(Things are currently in the parse package because I wanted to make an executed example of how to write a Selector in JSON, but was unwilling to make _all_ use of selectors depend on the JSON parser package.  A reorg of packages to get the engine stuff in a less central place would make it possible to both have and eat the cake.)

warpfork:
I think I'll close this because #236 has the more actionable plan in it already, but it remains good evidence that there's a user need for these packages to be less confusingly organized, so thanks for opening it.

PR #255, opened 2021-09-20
Use datamodel.Node for "Common" selector variants

rvagg:
Ref: https://github.com/ipld/go-ipld-prime/issues/236

The selector.Selector type used for these selector variants are not useful if
you're interacting with an API that wants `datamodel.Node` selector forms.
Rather than trying to re-use builder.SelectorSpec here, just return the
`datamodel.Node` form since it's trivial to compile the selector for ourselves
on demand.

I considered returning a `builder.SelectorSpec` but `Selector()` on it has an `error` return which is unnecessary for this use since we're checking comparability here. And #236 suggests that users shouldn't be interacting with `selector.Selector` anyway so let's just go with the DMT here eh?

I'd also vote for removing `ParseAndCompileJSONSelector` completely too if this is the course of action we want to take with this API and really push `selector.Selector` into the background.

### DIFF
--- a/traversal/selector/parse/selector_parse.go
+++ b/traversal/selector/parse/selector_parse.go
@@ -56,7 +56,7 @@ func ParseAndCompileJSONSelector(jsonStr string) (selector.Selector, error) {
 	}
 }
 
-func must(s selector.Selector, e error) selector.Selector {
+func must(s datamodel.Node, e error) datamodel.Node {
 	if e != nil {
 		panic(e)
 	}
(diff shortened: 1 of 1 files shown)


traversal/selector/parse/selector_parse.go
```diff
@@ -67,13 +67,13 @@ func must(s selector.Selector, e error) selector.Selector {
 // It doesn't walk anywhere at all.
 //
 // This is not a very useful selector, but is an example of how selectors can be written.
-var CommonSelector_MatchPoint = must(ParseAndCompileJSONSelector(`{".":{}}`))
+var CommonSelector_MatchPoint = must(ParseJSONSelector(`{".":{}}`))
```
willscott, line 70: I'd really like these to not be in the `parse` package at all. see #249 
Needed to point other people to them rather than re-implementing from a builder, and it took way too long for either of us to find where they'd been defined.

warpfork: I'm not thrilled with it either but the fix for both {make things discoverable from the more root-wards package} plus {selectors are possible to use without a strict dependency on dag-json} is going full ham on https://github.com/ipld/go-ipld-prime/issues/236 (which... might be a one-day effort, actually, come to think of it.  But might be more than that by the time we talk about change propagation).

rvagg: Yeah, I just figured that instead of me attempting to pull off #236 I'd just do a piece of it that's currently getting in my way. If it would help that effort to move them as part of this PR then I'm happy to change.

warpfork (approved):
Yep, seems fine to me.

rvagg:
> I'd also vote for removing `ParseAndCompileJSONSelector` completely too if this is the course of action we want to take with this API and really push `selector.Selector` into the background.

Maybe we do this as part of #236?

PR #256, opened 2021-09-20
Cleanup error wrapping in traversal package

### LINKED ISSUE #58: Can we wrap errors where possible? (go 1.13 feature)
when calling external dependencies like loaders, rather than generating a new error via a formatted string, it would be nice to use go 1.13's error wrapping, so that we can for example match the errors with errors.Is(). This may be as simple as switching %v to %w where needed.

For example:
https://github.com/ipld/go-ipld-prime/blob/master/traversal/walk.go#L203

By changing:
```
return nil, fmt.Errorf("error traversing node at %q: could not load link %q: %s", prog.Path, lnk, err)
```
to
```
return nil, fmt.Errorf("error traversing node at %q: could not load link %q: %w", prog.Path, lnk, err)
```
This could allow someone inspecting the results of WalkAdv to verify what went wrong in the loader more precisely

hannahhoward:
# Goals

fix #58 
Savor the sweet joy of using go >= 1.13

# Implementation

- Replace some %s's with %w's in calls to fmt.Errorf
- The one reason this is actually nice is that you can run errors.Is against the return value of a selector traversal or focus to compare against an error returned from a block load

### DIFF
--- a/traversal/focus.go
+++ b/traversal/focus.go
@@ -94,7 +94,7 @@ func (prog *Progress) get(n datamodel.Node, p datamodel.Path, trackProgress bool
 		case datamodel.Kind_Map:
 			next, err := n.LookupByString(seg.String())
 			if err != nil {
-				return nil, fmt.Errorf("error traversing segment %q on node at %q: %s", seg, p.Truncate(i), err)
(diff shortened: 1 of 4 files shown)


mvdan (approved):
:ok_hand:

mvdan:
Going to assume merging this is fine :) I also seem to recall Eric's code review stack is pretty large right now.

PR #257, opened 2021-09-20
schema,tests,gen/go: more tests, gen union fixes.

warpfork:
Increase the coverage produced by the schema test suites markedly, fixing a few old todos in the test system for typed nodes.

The additional exercise picks up some bugs in the codegen for unions.
(This was also reported by Adin, and I think possibly by Ian earlier,
though I only got around to reproducing it fully now.)  Those bugs --
off-by-ones; uuf -- are now fixed.

Some todos were encountered in bindnode by the new coverage too.
Only one thing.  But it's a bit more than I knew how to fix in one
sitting, so I've made that test skipped for now, hopefully to be
fixed soon.

Some error messages that confused me along the way are tweaked.

The datamodel.DeepEquals method is slightly less fragile on nils.

### DIFF
--- a/datamodel/equal.go
+++ b/datamodel/equal.go
@@ -25,6 +25,9 @@ package datamodel
 // so an error should only happen if a Node implementation breaks that contract.
 // It is generally not recommended to call DeepEqual on ADL nodes.
 func DeepEqual(x, y Node) bool {
+	if x == nil || y == nil {
+		return false
+	}
 	xk, yk := x.Kind(), y.Kind()
 	if xk != yk {
 		return false
--- a/datamodel/equal.go
+++ b/datamodel/equal.go
@@ -26,7 +26,7 @@ package datamodel
 // It is generally not recommended to call DeepEqual on ADL nodes.
 func DeepEqual(x, y Node) bool {
 	if x == nil || y == nil {
-		return false
+		return x == y
 	}
 	xk, yk := x.Kind(), y.Kind()
 	if xk != yk {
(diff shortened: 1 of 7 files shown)


datamodel/equal.go
```diff
@@ -25,6 +25,9 @@ package datamodel
 // so an error should only happen if a Node implementation breaks that contract.
 // It is generally not recommended to call DeepEqual on ADL nodes.
 func DeepEqual(x, y Node) bool {
+	if x == nil || y == nil {
```
rvagg, line 28: you don't want to `true` if _both_ are `nil`?

rvagg (approved):
👌

datamodel/equal.go
mvdan: I guess this goes into whether an untyped `nil` node is valid. I think `DeepEqual(nil, nil)` should indeed return true, and that's consistent with `reflect.DeepEqual` and what most Go users would expect.

You don't even need another `if` branch for this; you can just do:

```
if x == nil || y == nil {
    return x == y
}
```

mvdan: Also, I did write table-driven tests for DeepEqual, so please add coverage for this :)

mvdan (approved):

warpfork:
Added tests and made nils equal nils as suggested.  Thanks for those.  Will probably merge this shortly if not further objections; I think that addressed everything.

mvdan:
go ahead!

mvdan (approved):

PR #260, opened 2021-09-28
traversal: implement monotonically decrementing budgets.

warpfork:
This patch adds budgets to traversals.  Setting a budget allows you to make a traversal halt and return within a fixed number of steps.

It's optional, and currently not on by default, but easy to set.

Budgets can be set in terms of either number of nodes visited, or number of links loaded.

A few other minor target-of-opportunity improvements to internal documentation are included, while I was passing through.

### DIFF
--- a/traversal/fns.go
+++ b/traversal/fns.go
@@ -36,12 +36,25 @@ type Progress struct {
 		Path datamodel.Path
 		Link datamodel.Link
 	}
+	Budget *Budget // If present, tracks "budgets" for how many more steps we're willing to take before we should halt.
 }
 
 type Config struct {
 	Ctx                            context.Context                // Context carried through a traversal.  Optional; use it if you need cancellation.
 	LinkSystem                     linking.LinkSystem             // LinkSystem used for automatic link loading, and also any storing if mutation features (e.g. traversal.Transform) are used.
 	LinkTargetNodePrototypeChooser LinkTargetNodePrototypeChooser // Chooser for Node implementations to produce during automatic link traversal.
+}
+
+type Budget struct {
+	// Fields below are described as "monotonically-decrementing", because that's what the traversal library will do with them,
+	// but they are user-accessable and can be reset to higher numbers again by code in the visitor callbacks.  This is not recommended (why?), but possible.
+
+	// If you set any budgets (by having a non-nil Progress.Budget field), you must set some value for all of them.
+	// Traversal halts when _any_ of the budgets reaches zero.
+	// The max value of an int (math.MaxInt64) is acceptable for any budget you don't care about.
+
+	NodeBudget int64 // A monotonically-decrementing "budget" for how many more nodes we're willing to visit before halting.
+	LinkBudget int64 // A monotonically-decrementing "budget" for how many more links we're willing to load before halting.  (This is not aware of any caching; it's purely in terms of links encountered and traversed.)
 }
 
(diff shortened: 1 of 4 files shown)


warpfork:
(Oh, hey, I guess this addresses https://github.com/protocol/web3-dev-team/pull/27 .  Cool.)

warpfork:
This has unit tests, and typed errors now.

@ribasushi you seemed to have opinions about what kind of info you wanted to see in errors -- can you give that bit in particular a hairy eyeball?

traversal/fns.go
```diff
@@ -67,3 +81,17 @@ type SkipMe struct{}
...
 }
+
+type ErrBudgetExceeded struct {
+	BudgetKind string // "node"|"link"
+	Path       datamodel.Path
+	Link       datamodel.Link // only present if BudgetKind=="link"
+}
+
+func (e *ErrBudgetExceeded) Error() string {
+	msg := fmt.Sprintf("traversal budget exceeded: budget for %ss reached zero as we reached path %q", e.BudgetKind, e.Path)
```
ribasushi, line 92: Perhaps
```suggestion
	msg := fmt.Sprintf("traversal budget exceeded: budget for %ss reached zero while on path %q", e.BudgetKind, e.Path)
```

The rest of the PR looks 👍

traversal/walk_test.go
```diff
@@ -257,3 +258,67 @@ func TestWalkMatching(t *testing.T) {
...
+			case 2:
+				qt.Assert(t, n, qt.CmpEquals(), basicnode.NewString("beta"))
+				qt.Assert(t, prog.Path.String(), qt.Equals, "2")
+			}
+			order++
+			return nil
+		})
+		qt.Check(t, order, qt.Equals, 3)
+		qt.Assert(t, err, qt.Not(qt.Equals), nil)
+		qt.Check(t, err.Error(), qt.Equals, `traversal budget exceeded: budget for links reached zero as we reached path "3" (link: "baguqeeyexkjwnfy")`)
```
ribasushi, line 322: @warpfork shouldn't there be a leading slash to indicate "pah starts here" in path stringifications?

ribasushi (approved):
Left a couple nit-comments, good to ship as-is without changes.

rvagg:
OK, so you _must_ set a value for both budgets even if you only care about one - and most likely use-cases only care about one. I see a lot of code with max int64 for the node budget in our future.

tbh I'm not super enthused about this because it's all just arbitrary numbers, but in lieu of a more intelligent approach this will have to do for now.

traversal/walk_test.go
warpfork: There hasn't been anywhere in this implementation yet.  :shrug: 

(And you and I both know we've _tried_ to get some sort of quorum formed to have that discussion in full so we could decisively finalize something well-specified for path encoding... Alas...)

warpfork:
Took the proposed wording changes.

Agree it's arbitrary numbers, and that non-total functions aren't fun, but budgets are also the simplest (and thus most solid and reliable) way to ensure resources aren't being consumed beyond reason.

Also true that we can probably expect a lot of code with max int64 for the node budget in the future.  However, I don't really have a problem with that.  It's the simplest and clearest API I can think of.  (Alternative ideas seem to end up with either more bools or more pointers, and don't seem to get easier to explain or shape up more usable.)

warpfork:
Sounds like we're happy enough with this, so, moving to merge.

PR #264, opened 2021-10-07
Improve docs for AssignNode; and datamodel.Copy function.

warpfork:
Revisited the docs on the NodeAssembler interface, and described `AssignNode`'s contract and purpose better.

Also created a `datamodel.Copy` function.  This is related because just having that code gave me something to refer to which helped further clarify `AssignNode`.  (We should also probably start calling this function from codegen output, and perhaps other places, where similar logic is currently probably repeated, but I've left that out of this diff for now.)

### DIFF
--- a/datamodel/copy.go
+++ b/datamodel/copy.go
@@ -1,0 +1,110 @@
+package datamodel
+
+import (
+	"fmt"
+)
+
+// Copy does an explicit shallow copy of a Node's data into a NodeAssembler.
+//
+// This can be used to flip data from one memory layout to another
+// (for example, from basicnode to using using bindnode,
+// or to codegenerated node implementations,
+// or to or from ADL nodes, etc).
+//
+// The copy is implemented by ranging over the contents if it's a recursive kind,
+// and for each of them, using `AssignNode` on the child values;
+// for scalars, it's just calling the appropriate `Assign*` method.
+//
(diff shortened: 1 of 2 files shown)


datamodel/copy.go
```diff
@@ -0,0 +1,110 @@
...
+				return err
+			}
+		}
+		return ma.Finish()
+	case Kind_List:
+		la, err := na.BeginList(n.Length())
+		if err != nil {
+			return err
+		}
+		itr := n.MapIterator()
```
smrz2001, line 93:
```suggestion
		itr := n.ListIterator()
```

rvagg:
I'm wondering about the implications of just having a shallow `Copy`. How are users going to use this, what are their expectations, and what is the potential for mismatch between expectations and reality when it comes to just doing a shallow? Same goes for `AssignNode` too I suppose since the same set of concerns apply there too, no?

warpfork:
Yeah: we just have to try to solve that with docs docs docs, I think.  These are my first try, but maybe you can prose it even better.

We could also name this function `ShallowCopy`, if that makes anything better?

Fwiw, I don't think I'd be entirely surprised if we also someday want to have some kind of `DumbCopy` function which explicitly doesn't ever dip back into `AssignNode`, and does all the recursion itself.  But it's a little hard to defend that without saying "this is for working with node implementations that have broken contracts", so I'm on the fence about it.

PR #265, opened 2021-10-11
A more Featureful Approach to Storage APIs

warpfork:
(This is a PR of the "big, fun one" sort.)

**I've revisited storage APIs.**

In this PR, I set out some new interfaces, which I think improve on limitations of both the very old `go-datastore.Datastore` API, and the slightly less old `linking.BlockWriteOpener` and `linking.BlockReadOpener`.

It also provides a very smooth compatibility road between all three (both of the old ones, and the new one) -- smoother than I think we've ever had all in one place before.  It's now easier than ever to connect either brand new storage implementations to a `LinkSystem`, or, an older `go-datastore` implementation up to `LinkSystem` instead.

And it provides a lot (_lot_) of improvements in DX, dependency minimization, extensibility, usability, and, well... just about everything.  Let's see the whole list:

### list of improvements

1. improved over `go-datastore` ⇨ slimmer core; more feature-detection.
1. improved over `linking.Block*Opener` ⇨ more feature-detection.
1. improved over both `go-datastore` and `linking.Block*Opener` ⇨ feature-detection design means it's easy for end-users to pick the feature they need and use the package-scope functions -- and rest assured that the operation _will work_ regardless of the backing implementation; it's just a matter of whether it gets fastpaths or not.
1. improved over both `go-datastore` and `linking.Block*Opener` ⇨ a consistent strategy for feature-detection and optional fastpaths means it's easier to offer performance specializations than ever before.
1. improved over `linking.Block*Opener` ⇨ a consistent strategy for feature-detection means it's now clear where we'd expect storage system implementations to offer more other features (e.g. size estimates, or other non-core behaviors).  (Previously this was very very difficult, since the `linking.Block*Opener` types were function types, not interfaces -- meaning you can't feature-detect on them at all!)
1. improved over both `go-datastore` and `linking.Block*Opener` ⇨ **only standard library types are needed to implement the API** -- meaning it's much easier to write (and maintain) storage implementations without getting tangled in a complex dependency web.
1. improved over `go-datastore` ⇨ golang's modern standard `context.Context` is used everywhere.  (`go-datastore` predated `Context`; times have changed.)
1. improved over `go-datastore` ⇨ clear support for streaming operation.  (`go-datastore` just didn't have this, at all.)
1. improved over `linking.Block*Opener` ⇨ clear support for non-streaming operation.  (heh.)
1. improved over `linking.Block*Opener` ⇨ easier to implement!  Creating a minimum viable thing is now possible with just the most basic `Put`/`Get` methods (vs previously streaming operation was the minimum, which would generally be a bit more work).
1. improved over `go-datastore` ⇨ the read and write directions of operation are separated.  You can implement both features on one type if you like, but you're not required to.  (That means a read-only system can just _not have_ write methods, rather than being required to have them, but then being stuck having to error when they're used.)  Whenever storage handles are provided as configuration to some other system, the read and the write handles are distinct, and you can choose to set both or just one or the other (this is the same as how `linking.Block*Opener` already worked; it was good, we're keeping that part).
1. improved over `go-datastore` ⇨ the key type is much, _much_ simpler (and does not do strange internal manipulations that cause useless allocations!)
1. improved over `go-datastore` ⇨ the key type is now _binary_ and that's that.  If the storage implementation can't handle binary keys, the storage implementation should have some configuration system that accepts a function for escaping.  (In many `go-datastore` implementations, these boundaries were unclear, and some implementations would just reject some keys, forcing higher levels of application to deal with the problem, which could be confusing to reason about.)
1. improved over `go-datastore` ⇨ conceptually clearer and more explicit about intent to be used with one-write-per-key (because that's what a content-addressed system needs)... which results in implementations being allowed to make design choices that drastically simplify things and often increase performance.

### the new style of usage

As a user, if you're just doing IPLD things -- e.g. working mostly at the data model level -- then you can now use `LinkSystem.SetReadStorage` and `LinkSystem.SetWriteStorage` to set up storage.

As a user, if you want to work with storage directly, you should usually pass around the `storage.ReadableStorage` and `storage.WritableStorage` types, and use the always-available functions at package scope to actually operations on them.  (These package-scope functions mirror the ones available on the storage implementations themselves, and will "do the right thing" in the most efficient way possible, by using feature detection internally on your behalf, for ease-of-use.)

As an implementer: start off implementing `storage.ReadableStorage` or `storage.WritableStorage` first... and then pick and choose any of the additional feature-detection interfaces in the storage package that you think you can do especially well, or more optimally or more resource-efficiently than the fallback behaviors, and implement those too at your leisure.

### construction vs wiring

All of the APIs here are about connecting storage systems.

There's (still) no APIs attempting to standardize configuration for storage systems, nor standardize constructor funcs, on offer here.  (That was the story for `go-ipld-prime/linking.Block*Opener`, too, and also the story for `go-datastore` before that.)

(I might probe towards such configuration systems in https://github.com/ipld/go-ipldtool, but I don't think it's going to be appropriate to try to over-standardize that here -- I don't want to hem things in too much; letting things evolve is important.)

### connection to `go-datastore`

Notice that the new interfaces explicitly do not acknowledge the interfaces from the `github.com/ipfs/go-datastore` module.  We'll offer wrappers that bridge back to the `go-datastore` style API, because there's lots of things implemented today using it, but not be letting any of the concrete types from that old API leak through the new API.

(Was a new API necessary to achieve the above improvements?  The answer is: Yes, absolutely.  It would not have been feasible to reach the successes above by incrementally evolving that older interface; it's vastly less prohibitively difficult to get there by simply starting anew.)

The bridge to using `go-datastore` code as the new `go-ipld-prime/storage` API is already done: it's in the `storage/dsadapter` folder.  It's in this repo (because the code is so small!) but in a new go module -- which means you don't have to pull it (or its transitive dependency baggage) into your project unless you need it.

### connection to `go-ipld-prime/linking.Block*Opener`

`linking.BlockWriteOpener` and `linking.BlockReadOpener` stay as they are for now -- these are still the functions you use to configure a `LinkSystem`.

The new `storage.ReadableStorage` and `storage.WritableStorage` APIs can be glued to a `LinkSystem` easily, though: use `LinkSystem.SetReadStorage` and `LinkSystem.SetWriteStorage`.  With this, configuration is a one-liner.  (Or two-liner, if you are setting up both read and write.)

Note that the new storage APIs do _not_ receive the `linking.LinkContext` type.  That means: if you did want to do something like "please store things of type 'foo' in a different place" or "if I got here by a path that contains 'frobnoz', load info from a different place", you can't do that with these new storage APIs: you can't see the `LinkNode` or `LinkPath` anymore like you could in the `linking.BlockWriteOpener` and `linking.BlockReadOpener` functions.  If you need that?  You can still use the `linking.BlockWriteOpener` and `linking.BlockReadOpener` APIs... and maybe just shell out to the new `storage.*` inside of that.

(It's possible there will be further eventual revisits to `linking.BlockWriteOpener` and `linking.BlockReadOpener` in the future.  However, it's up in the air -- due to things like the user story in the previous paragraph -- for now.  Regardless of what happens: the eventual migration story should be smooth; we'll probably support both styles at once for a while, and graceful transitions can be done behind the `LinkSystem` abstraction, which should remain unaffected.)

### whew

I think this represents a pretty significant improvement to the state of storage APIs.  There's a few things to polish here, but the direction feels good.  Overall I hope, and think, this feels like a more comprehensive and inclusive solution than what we've had before.  Feedback and refinements welcome.

I'm currently fuelling this by turning around and immediately using it again downstream in the [go-ipldtool](https://github.com/ipld/go-ipldtool) repo.  So far, so good.  (This does make me a bit more averse to rebasing much, though -- the go mod file in that repo is referring to some of these commit hashes.)

If this goes well: I think this API might justify the tagging of a `v0.14.0` release after it lands.

### DIFF
--- a/adl/rot13adl/example_test.go
+++ b/adl/rot13adl/example_test.go
@@ -65,8 +65,8 @@ func ExampleReify_loadingToADL() {
 	}}
 	linkSystem := cidlink.DefaultLinkSystem()
 	storage := &memstore.Store{}
-	linkSystem.StorageReadOpener = storage.OpenRead
-	linkSystem.StorageWriteOpener = storage.OpenWrite
+	linkSystem.SetReadStorage(storage)
+	linkSystem.SetWriteStorage(storage)
 	linkSystem.NodeReifier = func(_ linking.LinkContext, nd datamodel.Node, _ *linking.LinkSystem) (datamodel.Node, error) {
 		return rot13adl.Reify(nd)
 	}
--- a/linking/linkingExamples_test.go
+++ b/linking/linkingExamples_test.go
@@ -36,8 +36,8 @@ func ExampleLinkSystem_Store() {
 	// We want to store the serialized data somewhere.
 	//  We'll use an in-memory store for this.  (It's a package scoped variable.)
 	//  You can use any kind of storage system here;
-	//   you just need a function that conforms to the datamodel.BlockWriteOpener interface.
-	lsys.StorageWriteOpener = (&store).OpenWrite
+	//   or if you need even more control, you could also write a function that conforms to the linking.BlockWriteOpener interface.
+	lsys.SetWriteStorage(&store)
 
 	// To create any links, first we need a LinkPrototype.
 	// This gathers together any parameters that might be needed when making a link.
@@ -103,8 +103,8 @@ func ExampleLinkSystem_Load() {
 	//  We'll use an in-memory store for this.  (It's a package scoped variable.)
 	//   (This particular memory store was filled with the data we'll load earlier, during ExampleLinkSystem_Store.)
 	//  You can use any kind of storage system here;
-	//   you just need a function that conforms to the datamodel.BlockReadOpener interface.
-	lsys.StorageReadOpener = (&store).OpenRead
+	//   or if you need even more control, you could also write a function that conforms to the linking.BlockReadOpener interface.
+	lsys.SetReadStorage(&store)
 
 	// We'll need to decide what in-memory implementation of datamodel.Node we want to use.
 	//  Here, we'll use the "basicnode" implementation.  This is a good getting-started choice.
--- a/node/tests/schemaLinks.go
+++ b/node/tests/schemaLinks.go
@@ -31,7 +31,7 @@ func encode(n datamodel.Node) (datamodel.Node, datamodel.Link) {
 		MhLength: 4,
 	}}
 	lsys := cidlink.DefaultLinkSystem()
-	lsys.StorageWriteOpener = (&store).OpenWrite
+	lsys.SetWriteStorage(&store)
 
 	lnk, err := lsys.Store(linking.LinkContext{}, lp, n)
 	if err != nil {
@@ -96,7 +96,7 @@ func SchemaTestLinks(t *testing.T, engine Engine) {
 
 		var order int
 		lsys := cidlink.DefaultLinkSystem()
-		lsys.StorageReadOpener = (&store).OpenRead
+		lsys.SetReadStorage(&store)
 		err = traversal.Progress{
 			Cfg: &traversal.Config{
 				LinkSystem: lsys,
--- a/storage/memstore/memstore.go
+++ b/storage/memstore/memstore.go
@@ -2,54 +2,91 @@ package memstore
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
-
-	"github.com/ipld/go-ipld-prime/datamodel"
-	"github.com/ipld/go-ipld-prime/linking"
 )
-
-// TODO: this is implementing the linking APIs, and it should be updated to just implement the storage APIs, and let LinkSystem construction figure out the rest of the connections.
 
 // Store is a simple in-memory storage.
 // (It's little more than a map -- in fact, the map is exported,
 // and you can poke it directly.)
 //
-// The OpenRead method conforms to linking.BlockReadOpener,
-// and the OpenWrite method conforms to linking.BlockWriteOpener.
-// Therefore it's easy to use in a LinkSystem like this:
+// Store conforms to the storage.ReadableStorage and storage.WritableStorage APIs.
+// Additionally, it supports storage.PeekableStorage and storage.StreamingReadableStorage,
+// because it can do so while provoking fewer copies.
 //
-//		store := memstore.Store{}
-//		lsys.StorageReadOpener = (&store).OpenRead
-//		lsys.StorageWriteOpener = (&store).OpenWrite
+// If you want to use this store with streaming APIs,
+// you can still do so by using the functions in the storage package,
+// such as storage.GetStream and storage.PutStream, which will synthesize the correct behavior.
+//
+// You can use this storage with a linking.LinkSystem easily,
+// by using the LinkSystem.SetReadStorage and/or LinkSystem.SetWriteStorage methods.
+//
+// There are no construction parameters for sharding functions nor escaping functions.
+// Any keys are acceptable.
 //
 // This storage is mostly expected to be used for testing and demos,
 // and as an example of how you can implement and integrate your own storage systems.
+// It does not provide persistence beyond memory.
 type Store struct {
-	Bag map[datamodel.Link][]byte
+	Bag map[string][]byte
 }
 
 func (store *Store) beInitialized() {
 	if store.Bag != nil {
 		return
 	}
-	store.Bag = make(map[datamodel.Link][]byte)
+	store.Bag = make(map[string][]byte)
 }
 
-func (store *Store) OpenRead(lnkCtx linking.LinkContext, lnk datamodel.Link) (io.Reader, error) {
+// Get implements go-ipld-prime/storage.ReadableStorage.Get.
+//
+// Note that this internally performs a defensive copy;
+// use Peek for higher performance if you are certain you won't mutate the returned slice.
+func (store *Store) Get(ctx context.Context, key string) ([]byte, error) {
 	store.beInitialized()
-	data, exists := store.Bag[lnk]
+	content, exists := store.Bag[key]
 	if !exists {
 		return nil, fmt.Errorf("404") // FIXME this needs a standard error type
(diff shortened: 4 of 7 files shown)


datamodel/link.go
```diff
@@ -34,6 +34,19 @@ type Link interface {
...
+	// the golang string type is used for immutability and for ease of use as a map key.
+	// As with the String method, the returned value may not elide any parts of the hash.
+	//
+	// Note that there is still no contract that the returned value be able to be parsed back into a Link value;
+	// not even in the case of `lnk.Prototype().BuildLink(lnk.Binary()[:])`.
+	// This is because the value returned by this method may contain data that the LinkPrototype would also restate.
+	// (For a concrete example: if using CIDs, this method will return a binary string that includes
+	// the cid version indicator, the multicodec and multihash indicators, etc, in addition to the hash itself --
+	// whereas the LinkPrototype.BuildLink function still expects to receive only the hash itself alone.)
+	Binary() string
```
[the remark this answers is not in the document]
mvdan: I have to side with Will here; strings are technically like read-only byte slices, but using them as such in return values is a bad idea. I think they're only reasonable in cases where the bytes being shoved into a string is necessarily a short-lived state, such as `map[string]T` or `string(input) == string(expected)`.

Like Will says, if the objective here is preventing corruption/races, make a copy - converting to a string makes a copy anyway. Converting to a string is arguably worse in all scenarios - if the user wants to use the bytes like a byte slice, they'll need to make a second copy for the conversion back.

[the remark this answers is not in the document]
warpfork: [The natural resident form of this data is usually `string`.](https://github.com/ipfs/go-cid/blob/5640b0169f6284562b1b8636b10e32dd85ea0a8a/cid.go#L207)

I think we would be encountering significantly more conversions (and probably, heap-escaping allocations) if this interface required `[]byte`.  (I was being flippant about the mutability thing; while that does bother me, it's not the most essential reason for this.)

warpfork:
For an end-to-end example of how this working if you want to use it through `LinkSystem`, here's a draft usage of it: https://github.com/ipld/go-ipldtool/blob/0b5f562654c21c3f881bc8c5d11c9d4ba904d999/app/basic/put.go#L26-L64

storage/dsadapter/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/ipld/go-ipld-prime/storage/dsadapter
```
mvdan, line 1: Why an extra module? If the reason is that you don't want ipld-prime to depend on go-datastore, then perhaps put this code in go-datastore.

If I want to use ipld-prime with datastore, I want to find the wrapper solution in either module - not have to discover an extra module. The IPLD/IPFS ecosystem already has enough Go modules as it is, too :)

storage/api.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"context"
+	"io"
+)
+
+// --- basics --->
+
+type ReadableStorage interface {
+	Get(ctx context.Context, key string) ([]byte, error)
```
mvdan, line 11: What's the precedent for using `key string` rather than `key []byte`? Basically every key-value database in Go (badger, bolt, pebble, etc etc) use `[]byte` keys, precisely because it helps clarify that they can contain anything - not just valid utf8.

If the point is to be more like go-datastore, perhaps that's enough precedent. But I would hope that we would make the interface be what is nicest for Go and as a generic abstraction layer, not just for IPFS in particular.

storage/api.go
```diff
@@ -0,0 +1,95 @@
...
+	PutVec(ctx context.Context, key string, blobVec [][]byte) error
+}
+
+// PeekableStorage is a feature-detection interface which a storage implementation can use to advertise
+// the ability to look at a piece of data, and return it in shared memory.
+// The PeekableStorage.Peek method is essentially the same as ReadableStorage.Get --
+// but by contrast, ReadableStorage is expected to return a safe copy.
+// PeekableStorage can be used when the caller knows they will not mutate the returned slice.
+type PeekableStorage interface {
+	Peek(ctx context.Context, key string) ([]byte, error)
```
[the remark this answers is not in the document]
mvdan: Another option, if you want to avoid callbacks, is also returning an `io.Closer` to signal that you're done using the returned bytes: https://pkg.go.dev/github.com/cockroachdb/pebble#Batch.Get

storage/dsadapter/go.mod
mvdan: I just read the README, which confirms my suspicion. I still think this code belongs in either of the two modules.

warpfork: :thinking:  I guess I can't really defend this, other than I think it's going to be operationally easier for me to deal with it here, and I kind of like the idea of having a few other things also in-repo here as sibling packages, so I can say "look in `storage/*` for implementations" in the docs.

(I have another freshly written filesystem-based storage system to add in a sibling directory, shortly, which would make for three of them here, including this one.)

You're right that this code could technically go in the go-datastore repo and module, due to the the success of the key design goal here of not having the new API require any types outside of the standard library.

warpfork: Alright, if someone wants to push this into `go-datastore` be my guest, but I want to see the PR landed in <24 hours, and then you also get to deal with whatever else comes up as I try to add benchmarks in the coming days that reach across old and new code :upside_down_face: Sounds like the road of greater hassle to me, but I dunno.  :shrug: 

Otherwise, I'd suggest we keep it here for now, and revisit in the future if it causes pain or discovery failure.  I don't think it should be difficult to move (by copying to a new home, and eventually removing the older location, after a transition period), if we find it necessary.  Leaving it here is a very "type 2 / revisitable" decision, in other words.

storage/api.go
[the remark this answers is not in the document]
raulk: Request: please review the interface of candidate backing implementations (like the one @whyrusleeping posted) before choosing a direction here. Getting creative with interfaces, while fun, may lead to poor choices. io.Closer is an interface that I would despise using, as it precludes the calling code from using closures.

PR #266, opened 2021-10-13
Add a cidlink.Memory storage option

willscott:
the current `storage.Memory` does not provide cid semantics where data can be stored under e.g. cidV0 and retrieved as cidV1.

### DIFF
--- a/linking/cid/memorystorage.go
+++ b/linking/cid/memorystorage.go
@@ -12,6 +12,11 @@ import (
 
 // Memory is a simple in-memory storage for cidlinks. It's the same as `storage.Memory`
 // but uses typical multihash semantics used when reading/writing cidlinks.
+//
+// Using multihash as the storage key rather than the whole CID will remove the
+// distinction between CIDv0 and their CIDv1 counterpart. It also removes the
+// distinction between CIDs where the multihash is the same but the codec is
+// different, e.g. `dag-cbor` and a `raw` version of the same data.
 type Memory struct {
 	Bag map[string][]byte
 }


rvagg (approved):
sgtm
`beInitialized()` is a bit of a naming oddity though, and are you intending for `Bag` to be public? seems like something that should remain private and be mediated through accessors if you need to do something fancy with it.

linking/cid/memorystorage.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/linking"
+)
+
+// Memory is a simple in-memory storage for cidlinks. It's the same as `storage.Memory`
+// but uses typical multihash semantics used when reading/writing cidlinks.
```
rvagg, line 14:
```suggestion
// Memory is a simple in-memory storage for cidlinks. It's the same as `storage.Memory`
// but uses typical multihash semantics used when reading/writing cidlinks.
//
// Using multihash as the storage key rather than the whole CID will remove the
// distinction between CIDv0 and their CIDv1 counterpart. It also removes the
// distinction between CIDs where the multihash is the same but the codec is
// different, e.g. `dag-cbor` and a `raw` version of the same data.
```

I do wonder what the implications might be throughout the loading process in that latter case, where you request a `raw` block but it comes back as a `dag-cbor`. I know we're now doing this kind of thing IRL now (esp Estuary) and probably should make sure we're good with this kind of thing, but I wonder if there will be some surprises for us along the way.

warpfork:
yeah, the memory storage type that Will's riffing off of here was really only intended for testing and demo usage, so it's very yolo/we-are-all-consenting-adults-here/unsafe.  I shan't defend it much :)

PR #267, opened 2021-10-14
linking: add LoadRaw and LoadPlusRaw functions to LinkSystem.

warpfork:
The new `LoadRaw` and `LoadPlusRaw` functions allow getting raw byte slices back along with the processed Node, or just the raw byte slice alone.

These behaviors can be useful if you're building an application which both loads the data and processes it, and also needs to pass on the serial form at the same time.

(Also added a great deal of documentation, including to existing functions which were lacking it.)


rvagg (approved):
👌 is good, should solve @willscott's problem @ https://github.com/ipfs/go-unixfsnode/pull/12/files#diff-e7601536d9fbbbc73c6e0384fbd5db1d2acc4cdf253ecc100bddbbbd15c3d799R155

PR #270, opened 2021-10-19
fully support parsing, encoding, and decoding the schema-schema

mvdan:
(see commit message)


schema/dsl/parse_test.go
```diff
@@ -1,27 +1,79 @@
...
+	srcJSON, err := ioutil.ReadFile(inputJSON)
+	qt.Assert(t, err, qt.IsNil)
+
+	testParse(t, string(src), string(srcJSON), func(updated string) {
+		err := ioutil.WriteFile(inputJSON, []byte(updated), 0o777)
+		qt.Assert(t, err, qt.IsNil)
+	})
+}
+
+func testParse(t *testing.T, inSchema, inJSON string, updateFn func(string)) {
```
mvdan, line 38: FYI, I wrote this helper because the intent is to add a table-driven tests to reuse @rvagg's previous test input suite at https://github.com/ipld/go-ipld-schema/tree/master/test/fixtures/bulk

schema/dmt/compile.go
```diff
@@ -134,9 +161,42 @@ func spawnType(ts *schema.TypeSystem, name schema.TypeName, defn TypeDefn) (sche
...
+						sumVal = schema.ImplicitValue_String(*imp.String)
+					default:
+						panic("TODO: implicit value kind")
+					}
+					implicits[name] = sumVal
+				}
+
+			}
+			repr = schema.SpawnStructRepresentationMap2(renames, implicits)
+		default:
```
warpfork, line 194: fwiw, if you do feel like continuing to strike hereabouts while the iron is hot, I think the tuple representation strategy is a relatively highly used one.  It's used in the filecoin protocol, heavily, certainly, at least.

This error is fine and explicit for now though.  Things are still moving in the right direction with this diff.

warpfork (approved):
All LGTM! :+1: :+1: :+1:

schema/dsl/parse_test.go
mvdan: Or that :) I'll probably do that as part of the next PR.

schema/dmt/compile.go
mvdan: Right, there are quite a lot of repr TODOs left, but we can incrementally work on those.

mvdan:
Updated the submodule to point to the new master. Merging :) Thanks for the speedy review!

PR #273, opened 2021-10-21
node/bindnode: more support for kinded unions, polish some panics

mvdan:

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -414,7 +414,7 @@ func (w *_node) AsLink() (datamodel.Link, error) {
 }
 
 func (w *_node) Prototype() datamodel.NodePrototype {
-	panic("TODO: Prototype")
+	return &_prototype{schemaType: w.schemaType, goType: w.val.Type()}
 }
 
 type _builder struct {
@@ -623,7 +623,7 @@ func (w *_assembler) AssignLink(link datamodel.Link) error {
 			// Unbox a cidlink.Link to assign to a go-cid.Cid value.
 			newVal = newVal.FieldByName("Cid")
 		} else {
-			// The target value cannot be assigned a ipld.Link or go-cid.Cid.
+			// The target value cannot be assigned a datamodel.Link or go-cid.Cid.
 			return datamodel.ErrWrongKind{
 				TypeName:        w.schemaType.Name(),
 				MethodName:      "AssignLink",
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -190,7 +190,7 @@ func unionMember(val reflect.Value) (int, reflect.Value) {
 	for i := 0; i < val.NumField(); i++ {
 		elemVal := val.Field(i)
 		if elemVal.Kind() != reflect.Ptr {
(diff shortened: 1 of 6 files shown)


mvdan:
Added one more commit. Now this adds support for *all* kinds in kinded unions :)

node/bindnode/node.go
```diff
@@ -189,6 +189,9 @@ func unionMember(val reflect.Value) (int, reflect.Value) {
 	// The first non-nil field is a match.
 	for i := 0; i < val.NumField(); i++ {
 		elemVal := val.Field(i)
+		if elemVal.Kind() != reflect.Ptr {
+			panic("found unexpected non-pointer in a union field")
```
warpfork, line 193: Do you think it might be time to start trying to adopt a consistent prefix -- something like perhaps `"bindnode cannot match golang type to schema:"` -- to errors like this?"

It wouldn't be load-bearing, but I think it'd be nice to start aiming towards a slightly more structured world.

node/bindnode/repr.go
```diff
@@ -53,7 +53,11 @@ func (w *_nodeRepr) Kind() datamodel.Kind {
 		return datamodel.Kind_Map
 	case schema.UnionRepresentation_Kinded:
 		haveIdx, _ := unionMember(w.val)
-		mtyp := w.schemaType.(*schema.TypeUnion).Members()[haveIdx] // TODO this is fragile: panics with index-out-of-range if the user has nil'd all fields.
+		if haveIdx < 0 {
+			// Invalid state; no union member is set.
+			return datamodel.Kind_Invalid
+		}
```
warpfork, line 59: I think this _could_ panic, fwiw, if it likes.  It should just say something explicit about "bindnode: node has invalid state (did you modify it directly, incorrectly?)".

I've tried really hard to avoid using `datamodel.Kind_Invalid` almost anywhere at all (the one exception being [schema.TypeUnion.RepresentationBehavior()](https://github.com/ipld/go-ipld-prime/blob/3d82a69fa632144c64f7e205f3b0f2ee28d27c27/schema/typeMethods.go#L40), because it has literally no other choice)... so I think I might actually prefer a panic here over pushing around invalid states to blow up somewhere else that will end up with less information about why.

Also: I think I used one-indexed numbering for union members in the gengo stuff in order to make sure uninitialized zero values would stand out like a sore thumb (as they should).  Might be worth doing here too.

node/bindnode/node.go
mvdan: Sure, will do. The panic trace should still point to `node/bindnode/node.go`, but if the panic gets recovered, then it's just the message string.

node/bindnode/repr.go
mvdan: Agreed. I actually added this because I ran into the panic while adding test cases. In particular, because AssignBytes would fail to do anything, so I'd end up with an invalid union state.

After this change, the dag-json encode gives an "encountered absent" error, which is slightly misleading at best. A clear panic seems better.

mvdan:
Applied both of your suggestions to the right commits. Merging on green. Thanks again for the reviews!

PR #274, opened 2021-10-21
add LoadSchema APIs to the root package

mvdan:
(see commit message)

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -1,0 +1,43 @@
+package ipld
+
+import (
+	"bytes"
+	"io"
+	"os"
+
+	"github.com/ipld/go-ipld-prime/schema"
+	schemadmt "github.com/ipld/go-ipld-prime/schema/dmt"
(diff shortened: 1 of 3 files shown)


schema.go
```diff
@@ -0,0 +1,43 @@
...
+
+// LoadSchemaBytes is a shortcut for LoadSchema for the common case where
+// the schema is a file on disk.
+func LoadSchemaFile(path string) (*schema.TypeSystem, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	return LoadSchema(path, f)
```
willscott, line 27: do you want `path.BaseName` here?

mvdan: I figured not, e.g. if I run it on `foo/bar/baz.ipldsch`, the full path seems useful.

One possibility would be to do like most compilers, and make the path relative to the current directory for readability. But "current directory" feels like a weird concept for a library.

PR #284, opened 2021-11-04
node/bindnode: support links in schema-type verification

mvdan:
(see commit message)


masih (approved):
Thank you for sorting this out so quickly 🚀 🍻

willscott (approved):

PR #287, opened 2021-11-04
fix: use `Repr` types in representation assembly methods

smrz2001:
A follow-up PR to my discussion with Eric regarding using schema-generated "representation" types for correctly encoding/decoding `dag-jose` while ignoring missing optional fields in the input data.

Changing to the "representation" type still didn't work because the generated code didn't seem to be correctly including the appropriate `*Repr` types in the `*ReprAssembler` and `*ReprBuilder` structures and functions.

By making the changes in this PR, I was able to get encode/decode (mostly) working correctly using code gen.

```
func (nb *_JOSE__ReprBuilder) Reset() {
-	var w _JOSE
+	var w _JOSE__Repr
	var m schema.Maybe
	*nb = _JOSE__ReprBuilder{_JOSE__ReprAssembler{w: &w, m: &m}}
}

type _JOSE__ReprAssembler struct {
-	w     *_JOSE
+	w     *_JOSE__Repr
	m     *schema.Maybe
	state maState
	s     int
	f     int
```

@warpfork

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,4 +1,3 @@
-github.com/creack/pty v1.1.9 h1:uDmaGzcdjhF4i/plgjmEsriH11Y0o7RKapEf/LDaM3w=
 github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
 github.com/frankban/quicktest v1.14.0 h1:+cqqvzZV87b4adx/5ayVOaYZ2CrvM4ejQvUdBzPPUss=
 github.com/frankban/quicktest v1.14.0/go.mod h1:NeW+ay9A/U67EYXNFA1nPE8e/tnQv/09mUdL/ijj8og=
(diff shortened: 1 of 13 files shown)


warpfork:
I don't think these diffs should be necessary.

What's going on in the code here can be a little tricky to follow, but it's relying heavily on the fact that golang will let you switch pretty freely between two types that have identical memory layouts... so, usually the code uses the non-repr types, when defining memory layout, and then "casts" to the repr variant (which has the *same memory layout*) (it's not really a cast, either, so much as telling the compiler "I want to use a different methodset on this memory") on demand.

smrz2001:
> I don't think these diffs should be necessary.
> 
> What's going on in the code here can be a little tricky to follow, but it's relying heavily on the fact that golang will let you switch pretty freely between two types that have identical memory layouts... so, usually the code uses the non-repr types, when defining memory layout, and then "casts" to the repr variant (which has the _same memory layout_) (it's not really a cast, either, so much as telling the compiler "I want to use a different methodset on this memory") on demand.

Ok, gotcha, thanks @warpfork.

The reason I ended up doing this was that when the assembly actually happens, the `_JOSE__Assembler` uses `_JOSE` (instead of  `_JOSE__Repr`), which not only has the wrong count of fields (it assumes all fields are included), it also encodes them as null. That always caused the encode to include absent fields.

There are method overrides on the `*Repr` structures that make their behavior different. For example, here are two such methods:

```
func (JOSE) Length() int64 {
	return 9
}
```
```
func (rn *_JOSE__Repr) Length() int64 {
	l := 9
	if rn.aad.m == schema.Maybe_Absent {
		l--
	}
	if rn.ciphertext.m == schema.Maybe_Absent {
		l--
	}
	if rn.iv.m == schema.Maybe_Absent {
		l--
	}
	if rn.payload.m == schema.Maybe_Absent {
		l--
	}
	if rn.protected.m == schema.Maybe_Absent {
		l--
	}
	if rn.recipients.m == schema.Maybe_Absent {
		l--
	}
	if rn.signatures.m == schema.Maybe_Absent {
		l--
	}
	if rn.tag.m == schema.Maybe_Absent {
		l--
	}
	if rn.unprotected.m == schema.Maybe_Absent {
		l--
	}
	return int64(l)
}
```

smrz2001:
And whoops (regarding test cases) 😅 

Will put up some test fixtures in the `go-dag-jose` repo shortly.

smrz2001:
I know how to fix the failing tests @warpfork, but I'm going to pause till we can figure out if I'm just doing something wrong entirely and these ^ diffs are unnecessary.

smrz2001:
I figured it out. Closing this PR.

PR #288, opened 2021-11-08
readme: introduce a migration guide.

warpfork:
This document discusses interfaces, APIs, and concept names from other libraries in golang, and how they map to this library.

Some references are to storage systems which are appearing in https://github.com/ipld/go-ipld-prime/pull/279 .

### DIFF
--- a/README_migrationGuide.md
+++ b/README_migrationGuide.md
@@ -1,0 +1,59 @@
+A short guide to migrating to go-ipld-prime from other repos
+============================================================
+
+Here are some quick notes on APIs that you might've been using
+if you worked with IPLD in golang before go-ipld-prime,
+and a pointer to what you should check out for an equivalent
+in order to upgrade your code to use go-ipld-prime.
+
+(Let us know if there are more pointers we should add to this list to ease your journey,
+or someone else's future journey!)
+
+- Were you using [ipfs/go-datastore](https://pkg.go.dev/github.com/ipfs/go-datastore) APIs?
+	- You can wrap those in `storage/dsadapter` and keep using them.
+	  You can also plug that into `linking.LinkSystem` to get higher level IPLD operations.
+	- Or if you were only using datastore because of some specific implementation of it, like, say `flatfs`?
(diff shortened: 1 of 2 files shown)


README_migrationGuide.md
```diff
@@ -0,0 +1,59 @@
...
+		- Be judicious about whether you actually want to do this.
+		  Plugging in the potential to experience unknown network latencies into code that's expecting predictable local lookup speeds
+		  may have undesirable performance outcomes.  (But if you want to do it, go ahead...)
+- Were you using [ipfs/go-ipld-format.DAGService](https://pkg.go.dev/github.com/ipfs/go-ipld-format#DAGService)?
+	- If you're using the `Add` and `Get` methods: those are now on `linking.LinkSystem`, as `Store` and `Load`.
+	- If you're using the `*Many` methods: those don't have direct replacements;
+	  we don't use APIs that force that style of future-aggregation anymore, because it's bad for pipelining.
+	  Just use `Store` and `Load`.
+- Were you using [ipfs/go-ipld-format.Node](https://pkg.go.dev/github.com/ipfs/go-ipld-format#Node)?
+	- That's actually a semantic doozy -- we use the word "node" _much_ differently now.
```
rvagg, line 33: whoa boy, yeah, is it any wonder this continues to confuse people if we have multiple uses of the word that mean something entirely different but so closely?

rvagg (approved):
:thumbsup: good start
there's probably a slew of stuff in go-merkledag worth mentioning too, including the dag-pb stuff (already here) and even DAG "walking" which should be replaced with selectors now

README_migrationGuide.md
warpfork: Yeah, I wish... this was different, somehow.  But most of those wishes seem to involve time machines at this point.  Alas.

warpfork:
Okay, I'm in cleanup mode today, so I'm gonna merge this.  Agree more pointers on migrating code about walking would probably be good in the future too though.

Issue #219, opened 2021-08-04
wishlist: port all tests to quicktest


warpfork:
https://pkg.go.dev/github.com/frankban/quicktest is now clearly superior to https://pkg.go.dev/github.com/warpfork/go-wish .

We've started using `quicktest` in some parts of this repo already and it's working great.

It would be nice to port all the tests to using `quicktest` and see one more dependency drop from the `go.mod` file.

This should be relatively easy.  The style used by `quicktest` and `wish` are syntactically and semantically extremely similar.

mvdan:
I'd usually recommend to do a slow transition as the codebase is modified, to avoid too much churn and conflicts.

But in this case, given that go-wish and quicktest are pretty similar, it probably doesn't make sense to leave the repository using both for months or years. So maybe it is worth the effort to try to automate the refactor.

Maybe it could be done in groups of packages, to make the reviews less painful.

PR #289, opened 2021-11-09
Port `node` package tests to quicktest

masih:
Port the tests in `node` package to quicktest; use:
 - `qt.Check` for `wish.Wish`
 - `qt.IsTrue` for ShouldEqual` over `true`

Port `closeEnough` Wish checker to its equivalent in quicktest.

Relates to:

### DIFF
--- a/node/tests/schemaStruct.go
+++ b/node/tests/schemaStruct.go
@@ -3,7 +3,7 @@ package tests
 import (
 	"testing"
 
-	. "github.com/warpfork/go-wish"
+	qt "github.com/frankban/quicktest"
 
 	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/fluent"
@@ -43,8 +43,8 @@ func SchemaTestRequiredFields(t *testing.T, engine Engine) {
 		ma, _ := nb.BeginMap(0)
 		err := ma.Finish()
 
-		Wish(t, err, ShouldBeSameTypeAs, schema.ErrMissingRequiredField{})
-		Wish(t, err.Error(), ShouldEqual, `missing required fields: a,b`)
+		qt.Check(t, err, qt.ErrorAs, &schema.ErrMissingRequiredField{})
+		qt.Check(t, err.Error(), qt.Equals, `missing required fields: a,b`)
 	})
 	t.Run("building-representation-without-required-fields-errors", func(t *testing.T) {
 		nrp := engine.PrototypeByName("StructOne.Repr")
@@ -53,8 +53,8 @@ func SchemaTestRequiredFields(t *testing.T, engine Engine) {
 		ma, _ := nb.BeginMap(0)
 		err := ma.Finish()
 
-		Wish(t, err, ShouldBeSameTypeAs, schema.ErrMissingRequiredField{})
-		Wish(t, err.Error(), ShouldEqual, `missing required fields: a,b`)
+		qt.Check(t, err, qt.ErrorAs, &schema.ErrMissingRequiredField{})
+		qt.Check(t, err.Error(), qt.Equals, `missing required fields: a,b`)
 	})
 	t.Run("building-representation-with-renames-without-required-fields-errors", func(t *testing.T) {
 		nrp := engine.PrototypeByName("StructTwo.Repr")
@@ -63,8 +63,8 @@ func SchemaTestRequiredFields(t *testing.T, engine Engine) {
 		ma, _ := nb.BeginMap(0)
 		err := ma.Finish()
 
-		Wish(t, err, ShouldBeSameTypeAs, schema.ErrMissingRequiredField{})
-		Wish(t, err.Error(), ShouldEqual, `missing required fields: a,b (serial:"z")`)
+		qt.Check(t, err, qt.ErrorAs, &schema.ErrMissingRequiredField{})
+		qt.Check(t, err.Error(), qt.Equals, `missing required fields: a,b (serial:"z")`)
 	})
 }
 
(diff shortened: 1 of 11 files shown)


node/tests/schemaStruct.go
```diff
@@ -43,8 +43,8 @@ func SchemaTestRequiredFields(t *testing.T, engine Engine) {
 		ma, _ := nb.BeginMap(0)
 		err := ma.Finish()
 
-		Wish(t, err, ShouldBeSameTypeAs, schema.ErrMissingRequiredField{})
-		Wish(t, err.Error(), ShouldEqual, `missing required fields: a,b`)
+		qt.Check(t, err, qt.ErrorAs, &schema.ErrMissingRequiredField{})
```
masih, line 46: Please pay attention to this modification; I'd love to make sure the changes are equivalent to what `ShouldBeSameTypeAs` is meant to do. 🍻

mvdan: I think the semantics here are slightly different, as errors.As unwraps error values - but that's fine, because the new semantics look better. The tests should enforce `errors.As`, not the error's top-level type.

node/tests/testcase.go
```diff
@@ -262,60 +263,71 @@ func testMarshal(t *testing.T, n datamodel.Node, data string) {
...
 // If the expected value is a primitive string, it'll AsStrong on the Node; etc.
 //
 // Using a datamodel.Kind value is also possible, which will just check the kind and not the value contents.
 //
-// If a datamodel.Node is the expected value, a full deep ShouldEqual is used as normal.
-func closeEnough(actual, expected interface{}) (string, bool) {
+// If a datamodel.Node is the expected value, a full deep qt.Equals is used as normal.
+var closeEnough = &closeEnoughChecker{}
+
+var _ qt.Checker = (*closeEnoughChecker)(nil)
```
mvdan, line 295: clever :) I like how this turned out, even if implementing a custom checker with quicktest is slightly more verbose.

you could almost use qt.CmpEquals with https://pkg.go.dev/github.com/google/go-cmp/cmp#Comparer here, except that Comparer expects `func(T, T) bool`, so it only kicks in when the two types being compared are exactly equal. The logic here is different, as it is more flexible.

node/tests/testcase.go
```diff
@@ -155,17 +156,17 @@ func (tcase testcase) Test(t *testing.T, np, npr datamodel.NodePrototype) {
...
 			for _, entry := range tcase.typeItr {
-				Wish(t, itr.Done(), ShouldEqual, false)
+				qt.Check(t, itr.Done(), qt.Equals, false)
 				k, v, err := itr.Next()
-				Wish(t, k, closeEnough, entry.key)
-				Wish(t, v, closeEnough, entry.value)
-				Wish(t, err, ShouldEqual, nil)
+				qt.Check(t, k, closeEnough, entry.key)
+				qt.Check(t, v, closeEnough, entry.value)
+				qt.Check(t, err, qt.Equals, nil)
```
mvdan, line 163: nit: search-replace `qt.Equals, nil` for `qt.IsNil`.

node/tests/testcase.go
```diff
@@ -155,17 +156,17 @@ func (tcase testcase) Test(t *testing.T, np, npr datamodel.NodePrototype) {
 			//  because the only kind of thing that needs this style of testing are some instances of maps and some instances of structs.
 			itr := n.MapIterator()
 			for _, entry := range tcase.typeItr {
-				Wish(t, itr.Done(), ShouldEqual, false)
+				qt.Check(t, itr.Done(), qt.Equals, false)
```
mvdan, line 159: nit: search-replace `qt.Equals, false` for `qt.IsFalse`

node/tests/testcase.go
```diff
@@ -197,8 +198,8 @@ func (tcase testcase) Test(t *testing.T, np, npr datamodel.NodePrototype) {
 		if n.(schema.TypedNode).Representation().Kind() == datamodel.Kind_Map {
 			t.Run("repr-create with AK+AV", func(t *testing.T) {
 				n3, err := shallowCopyMap(npr, n.(schema.TypedNode).Representation())
-				Wish(t, err, ShouldEqual, nil)
-				Wish(t, datamodel.DeepEqual(n, n3), ShouldEqual, true)
+				qt.Check(t, err, qt.Equals, nil)
+				qt.Check(t, datamodel.DeepEqual(n, n3), qt.Equals, true)
```
mvdan, line 202: mental TODO: in the future, we might be able to expose datamodel.DeepEqual as a go-cmp Comparer. Definitely out of scope for this refactor, though.

mvdan (approved):
Excellent work!

By the way, you might want to encode those "shorter qt expression" nits as sed lines, so you can automatically apply those cleanups throughout your PRs.

mvdan:
BTW, not sure if you're aware, but qt.Check does a t.Error, whereas qt.Assert does a t.Fatal. I think in general we want the latter. For example:

```
foo, err := bar()
qt.Check(t, err, qt.IsNil) // assume this errors because err != nil

qt.Check(t, foo.Baz, qt.Equals, "xyz") // panic! foo is nil
```

If you use `qt.Assert` on both lines, then we stop when `err != nil`, and we don't panic.

`qt.Check` is for the few times where we do want to continue running a test even when it errors once. That's commonly used e.g. when you check multiple properties in a row, so you want to report all the errors, and not just the first one.

masih:
Thank you for pointing that out @mvdan; I interpreted tests as `wish.Require` should do `t.fatal` and `wish.Wish` should do `t.Error` then used the equivalent `qt` function to keep the behaviour consistent with what's there just now.

Do you think the behaviour of tests should change?

masih:
Thank you; I intentionally wanted to keep changes as net-zero as possible

node/tests/testcase.go
warpfork: One could also use the `printer` package and then do string diff on the results.

That would simultaneously be very complete, and produce very readable debugable output, and also happily ignore the actual node implementation strategy (which is usually what we want).

masih: `quicktest` comes with [`CmpEquals`](https://pkg.go.dev/github.com/frankban/quicktest#hdr-CmpEquals) which looks pretty flexible; I wonder if we could use that here?

mvdan: That is what I was alluding to. But you need a custom comparer, because by default, CmpEquals will just loop infinitely. This is because bindnode ipld.Node implementations contain schema types, which have cyclic pointers.

mvdan: I originally arrived there via the flat datamodel.DeepEqual, which works, but also doesn't give good errors. I think either a custom go-cmp comparer, or comparing encoded versions like "printer" or "dagjson" like Eric mentions, would work better.

warpfork (approved):
(Wow, we have a lot of tests!)

Nice.  It'll be great to have one fewer testing system in the dependency tree by the end of this (and the similar PRs).

node/tests/testcase.go
masih: I attempted to write a checker that works with `printer` [here](https://github.com/ipld/go-ipld-prime/pull/291/commits/3c062db77fe81b0db0237f419be5b52ac9316005) but looks like there are TODOs in the printer that tests hit. Going to use `datamodel.DeepEqual` going forward.

PR #291, opened 2021-11-10
Port `codec` package tests to quicktest

masih:
Implement `NodeContentEquals` quicktest checker that checks two given
nodes have equal content by comparing their printout using
`printer.Sprint`. This simplifies the tests that use
`datamode.DeepEqual` by performing an equivalent check while producing a
human-readable error when nodes are not equal. The naming for this check
is inspired by a similar equality check in quicktest, named
`ContentEquals`.

Port the tests in `codec` package to quicktest; use:
 - `qt.Check` for `wish.Wish`
 - `qt.IsTrue` for `ShouldEqual` over `true`
 - `NodeContentEquals` for `ShouldEqual` over nodes
 - `NodeContentEquals` for `datamodel.DeepEqual` over nodes

Update `NodeContentEquals` for `datamodel.DeepEqual` over nodes in
`node` package while at it.

Port `node/mixins/TestSplitExact` over to quicktest missed out in
earlier PRs. Note, to assert equality `CmpEqual` is used with an
`Exporter` that exports all unexpored fields.

Address TODO in node tests by using `NodeContentEquals` to check node
equality.

Relates to:

Depends on:
 - https://github.com/ipld/go-ipld-prime/pull/294

### DIFF
--- a/node/tests/checkers.go
+++ b/node/tests/checkers.go
@@ -1,0 +1,41 @@
+package tests
+
+import (
+	"errors"
+
+	qt "github.com/frankban/quicktest"
+
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/printer"
+)
+
+// NodeContentEquals checks whether two nodes have equal content by first encoding them via
+// printer.Sprint, then checking that the generated encodings are identical.
+//
+// See: printer.Sprint.
+var NodeContentEquals = &nodeContentEqualsChecker{}
+
+type nodeContentEqualsChecker struct{}
+
+func (n *nodeContentEqualsChecker) Check(got interface{}, args []interface{}, note func(key string, value interface{})) error {
+	want := args[0]
+	if want == nil {
+		return qt.IsNil.Check(got, args, note)
+	}
+	wantNode, ok := want.(datamodel.Node)
+	if !ok {
+		return errors.New("this checker only supports checking datamodel.Node values")
+	}
+	wantPrint := printer.Sprint(wantNode)
+
+	gotNode, ok := got.(datamodel.Node)
+	if !ok {
+		return errors.New("this checker only supports checking datamodel.Node values")
+	}
+	gotPrint := printer.Sprint(gotNode)
+	return qt.Equals.Check(gotPrint, []interface{}{wantPrint}, note)
+}
+
+func (n *nodeContentEqualsChecker) ArgNames() []string {
+	return []string{"got", "want node"}
+}
--- a/node/tests/testcase.go
+++ b/node/tests/testcase.go
@@ -126,7 +126,7 @@ func (tcase testcase) Test(t *testing.T, np, npr datamodel.NodePrototype) {
 		}
 		if n2 != nil {
 			t.Run("type-create and repr-create match", func(t *testing.T) {
-				qt.Check(t, datamodel.DeepEqual(n, n2), qt.IsTrue)
+				qt.Check(t, n, NodeContentEquals, n2)
 			})
 		}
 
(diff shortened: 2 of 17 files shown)


node/tests/testcase.go
```diff
@@ -199,10 +199,7 @@ func (tcase testcase) Test(t *testing.T, np, npr datamodel.NodePrototype) {
 			t.Run("repr-create with AK+AV", func(t *testing.T) {
 				n3, err := shallowCopyMap(npr, n.(schema.TypedNode).Representation())
 				qt.Check(t, err, qt.IsNil)
-				// TODO: Improve checking mechanism; hits:
-				// -  expose datamodel.DeepEqual as a go-cmp Comparer, or
-				// - use the printer package and then do string diff on the results
-				qt.Check(t, datamodel.DeepEqual(n, n3), qt.IsTrue)
+				qt.Check(t, n3, NodeContentEquals, n)
```
masih, line 202: Note, there are other tests in `node` package that use `datamodel.DeepEqual` to assert equality of two nodes. I thought I check that reviewers are happy with this change before changing all assertions to use the new checker.

node/tests/checkers.go
```diff
@@ -0,0 +1,50 @@
...
+	want := args[0]
+	if want == nil {
+		return qt.IsNil.Check(got, args, note)
+	}
+	wantNode, ok := want.(datamodel.Node)
+	if !ok {
+		return errors.New("this checker only supports checking datamodel.Node values")
+	}
+	var buf bytes.Buffer
+	if err := dagjson.Encode(wantNode, &buf); err != nil {
```
warpfork, line 31: I think I'd suggest replacing this (and the other one below) with [`printer.Sprint(n)`](https://github.com/ipld/go-ipld-prime/blob/b9a89e847312334af141788b303fd78b4f076fe3/printer/printer.go#L23).

It's going to be behaviourally awfully similar in all the key goals, but I think often better:
- better on readability (the printer produces multiline output),
- will sometimes give slightly more coverage (it's going to tell you if there's schema types involved, and what they are, if applicable),
- and it just plain gives me a bit less fear about the dependency tree getting kinky than depending on a codec package here.

node/tests/testcase.go
mvdan: SGTM, and this definitely leaves the tests in a better place long-term

node/tests/checkers.go
masih: I had that implemented at first but unfortunately tests hit the unimplemented `TODO`s in printer which is a blocker.

masih: Another alternative is to use `datamodel.DeepEqual` in the checker and call the checker `NodeDeepEquals`

warpfork: :facepalm: :facepalm: :facepalm: 

Those might be fast enough to fill in that it's worth it.  I can try to take a look later, or maybe talk you into it?

mvdan: IMO this is precisely the kind of thing that could be a small follow-up PR :)

masih: Happy to be talked into stuff 🚀

masih: For reference, `printer` TODOs that tests hit are now implemented [here](https://github.com/ipld/go-ipld-prime/pull/294). Once that's merged I'll revert back to using `printer` to implement the checker here. 🍻

masih:
Tests fail because #294 is not merged yet. Once that's merged, this branch needs a rebase and we are good to go 🚀

mvdan:
Said PR merged, if you want to rebase.

masih:
OK now we are hitting this:
* https://github.com/ipld/go-ipld-prime/issues/295

node/tests/checkers.go
masih: Reverting back to using dagjson, since printer panics, caused by a TODO in test generator; see #295. Two things need to happen before we can use `printer` for the checker:

1. Implement TODO in test generator, which #295 should address
1. Be more defensive in `printer` implementation and handle `nil`s gracefully, captured in #296

The key thing in all of this is to make sure checker used in tests is correct since tests rely on it to assert content equality.

masih:
Printer issues fixed in: 
* #297 
* #298

warpfork (approved):
Sweet!

PR #292, opened 2021-11-10
Port `datamodel` package tests to quicktest

masih:
Port the tests in `datamodel` package to quicktest; use:
 - `qt.Check` for `wish.Wish`
 - `qt.CmpEquals` with custom `Exporter` that exports all fields for
   `ShouldEqual` over `PathSegment`

Relates to:

### DIFF
--- a/datamodel/path_test.go
+++ b/datamodel/path_test.go
@@ -1,35 +1,39 @@
 package datamodel
 
 import (
+	"reflect"
 	"testing"
 
-	. "github.com/warpfork/go-wish"
+	qt "github.com/frankban/quicktest"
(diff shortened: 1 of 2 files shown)


masih:
I have accidentally pushed the changes here to master (excluding the go.mod change)... my apologies; What can I do to fix this?

mvdan:
It had only been ~15m so I removed the broken, unintended commit from master via force pushing. FYI @warpfork, we should really add a protection rule for master to prevent future unintended pushes from non-admins.

If it had been longer than an hour, I'd definitely gone for a revert commit.

warpfork (approved):
Okay but now you can put it back anyway :)

PR #294, opened 2021-11-10
Implement handling of `Link` and `[]byte` in `printer`

masih:
Implement textual representation for types `Link` and `[]byte` by
`printer`. Represent `Link` as its string value, and `[]byte` as its hex
encoded value.

### DIFF
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -1,6 +1,7 @@
 package printer
 
 import (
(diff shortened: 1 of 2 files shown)


warpfork (approved):
LGTM!   Thanks for this!

mvdan (approved):

PR #297, opened 2021-11-11
Gracefully print typed nodes with `nil` type

### LINKED ISSUE #296: Print defensively in `printer`
Ideally `printer` should handle `nil`s gracefully even if the interface it's interacting with should not practically return `nil`. Instead consider printing something that highlights the misschief, e.g. `<?nil>` 

Relates to #295

masih:
Given an invalid `TypedNode` with `nil` type gracefully print a
representation of its content by using:
- `?!nil` as type name
- `schema.TypeKind_Invalid` as its `TypeKind`

Fixes #296

### DIFF
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -175,11 +175,21 @@ func (z *printBuf) doString(indentLevel int, printState uint8, n datamodel.Node)
 	//  Note: this can be somewhat overbearing -- for example, typed strings are going to get called out as `string<String>{"value"}`.
 	//   This is rather agonizingly verbose, but also accurate; I'm not sure if we'd want to elide information about typed-vs-untyped entirely.
 	if tn, ok := n.(schema.TypedNode); ok {
-		z.writeString(tn.Type().TypeKind().String())
+		var tnk schema.TypeKind
+		var tntName string
+		// Defensively check for nil node type
+		if tnt := tn.Type(); tnt == nil {
+			tntName = "?!nil"
+			tnk = schema.TypeKind_Invalid
+		} else {
+			tntName = tnt.Name()
+			tnk = tnt.TypeKind()
+		}
+		z.writeString(tnk.String())
 		z.writeString("<")
-		z.writeString(tn.Type().Name())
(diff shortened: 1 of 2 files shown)


printer/printer.go
```diff
@@ -175,11 +175,21 @@ func (z *printBuf) doString(indentLevel int, printState uint8, n datamodel.Node)
...
+		}
+		z.writeString(tnk.String())
 		z.writeString("<")
-		z.writeString(tn.Type().Name())
+		z.writeString(tntName)
 		z.writeString(">")
-		switch tn.Type().TypeKind() {
+		switch tnk {
 		case schema.TypeKind_Invalid:
 			z.writeString("{?!}")
```
warpfork, line 194: this is getting into odd territory, but I think it should be fine to fall through and let the data-model driven behavior still do whatever it can?

warpfork: ... on second thought, this is so very far from relevance that I don't know if there's a reason to care one way or the other, and I don't want to make you rewrite the tests you already made. It's fine as is.

warpfork (approved):
:ship:

printer/printer.go
masih: That's what I was thinking; initially I thought introducing a special implementation for `Type` and using that when the original is `nil` would leave the `printer` logic cleaner and more consistent. But doing that now is probably an overkill. 

I checked the rest of the logic and I think we are good to let it follow on. Happy to change if you think we shouldn't.

masih: OK I take all my optimism back 🤦 nil type when printing content still causes trouble

masih: Fixed here: https://github.com/ipld/go-ipld-prime/pull/298 

Apologies

rvagg:
out of interest, is `?!` a Go thing, or are we just fishing for likely-non-conflicting tokens for these?

mvdan:
I believe the latter.

masih:
I think `?!` here is just to draw attention to the output to say "hey this is not normal"

PR #299, opened 2021-11-11
Port `traversal` package tests to quicktest

### LINKED ISSUE #219: wishlist: port all tests to quicktest

masih:
Port the tests in `traversal` package to quicktest; use:
 - `qt.Check` for `wish.Wish`
 - `qt.IsTrue` for `ShouldEqual` over `true`
 - `qt.HasLen` for `ShouldEqual` over slice length
 - `qt.DeepEquals` for `ShouldEqual` over slices
 - `CmpEquals` for `ShouldEqual` over types with unexported fields
    excluding `datamodel.Node` with custom exporter to allow checking
    of all unexported fields
      - Defined checker once and reused across packages. Did not add
        this to `node/tests` to avoid cycle package dependency.
 - `nodetests.NodeContentEquals` over `datamodel.DeepEqual` and
   `ShouldEqual` over `datamodel.Node`

Fixes #219

### DIFF
--- a/traversal/focus_test.go
+++ b/traversal/focus_test.go
@@ -1,12 +1,13 @@
 package traversal_test
 
 import (
-	"fmt"
+	"reflect"
 	"testing"
 
-	. "github.com/warpfork/go-wish"
-
+	qt "github.com/frankban/quicktest"
+	"github.com/google/go-cmp/cmp"
 	"github.com/ipfs/go-cid"
+
 	_ "github.com/ipld/go-ipld-prime/codec/dagjson"
 	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/fluent"
@@ -14,12 +15,15 @@ import (
 	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	"github.com/ipld/go-ipld-prime/must"
 	"github.com/ipld/go-ipld-prime/node/basicnode"
+	nodetests "github.com/ipld/go-ipld-prime/node/tests"
 	"github.com/ipld/go-ipld-prime/storage/memstore"
 	"github.com/ipld/go-ipld-prime/traversal"
 )
 
 // Do some fixture fabrication.
 // We assume all the builders and serialization must Just Work here.
+
+var deepEqualsAllowAllUnexported = qt.CmpEquals(cmp.Exporter(func(reflect.Type) bool { return true }))
 
 var store = memstore.Store{}
 var (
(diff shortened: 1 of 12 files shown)


masih:
Last one in the series; once this is merged all tests should be using `quickest`. The only remaining dependency to `go-wish`  is `Wish.Dedent`.

traversal/select_links_test.go
```diff
@@ -1,25 +1,30 @@
...
+	qt "github.com/frankban/quicktest"
+	"github.com/google/go-cmp/cmp"
 
 	"github.com/ipld/go-ipld-prime/datamodel"
 	"github.com/ipld/go-ipld-prime/traversal"
 )
 
 func TestSelectLinks(t *testing.T) {
+
+	linkEquals := qt.CmpEquals(cmp.Exporter(func(reflect.Type) bool { return true }))
```
mvdan, line 16: document why this is needed? I would assume that qt.DeepEquals would be equivalent to ShouldEqual

traversal/focus_test.go
```diff
@@ -326,18 +330,18 @@ func TestFocusedTransformWithLinks(t *testing.T) {
...
 			nb.AssignString("new string!")
 			return nb.Build(), nil
 		}, false)
-		Wish(t, err, ShouldEqual, nil)
-		Wish(t, n.Kind(), ShouldEqual, datamodel.Kind_Map)
+		qt.Check(t, err, qt.IsNil)
+		qt.Check(t, n.Kind(), qt.Equals, datamodel.Kind_Map)
 		// there should be a new object in our new storage!
-		Wish(t, len(store2.Bag), ShouldEqual, 1)
+		qt.Check(t, len(store2.Bag), qt.Equals, 1)
```
mvdan, line 344: nit: qt also has qt.HasLen, which is shorter, and also allows giving better errors (e.g. include the overall value) when the check fails. probably worth checking if this occurs elsewhere across the repo as well.

mvdan (approved):
Smashing work!

traversal/select_links_test.go
masih: Fat fingers on my part; this should not be there. thank you for pointing that out 🍻

PR #300, opened 2021-11-11
Remove dependency to `go-wish`

masih:
Fork `wish.Dedent` code into `testutil` package along with its test and
replace references to `go-wish` with forked function.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,6 @@ require (
 	github.com/polydawn/refmt v0.0.0-20201211092308-30ac6d18308e
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	github.com/warpfork/go-testmark v0.3.0
-	github.com/warpfork/go-wish v0.0.0-20200122115046-b9ea61034e4a
+	github.com/warpfork/go-wish v0.0.0-20200122115046-b9ea61034e4a // indirect
 	gopkg.in/yaml.v2 v2.4.0
 )
--- a/printer/printer_test.go
+++ b/printer/printer_test.go
@@ -4,7 +4,6 @@ import (
 	"testing"
 
 	qt "github.com/frankban/quicktest"
-	"github.com/warpfork/go-wish"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime/datamodel"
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -11,6 +11,6 @@ require (
 	github.com/polydawn/refmt v0.0.0-20201211092308-30ac6d18308e
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	github.com/warpfork/go-testmark v0.3.0
-	github.com/warpfork/go-wish v0.0.0-20200122115046-b9ea61034e4a
+	github.com/warpfork/go-wish v0.0.0-20200122115046-b9ea61034e4a // indirect
```
masih, line 14: I am not sure where this is coming from; this is what the dependency graph tells me:

```
$ go mod graph | grep go-wish
github.com/ipld/go-ipld-prime github.com/warpfork/go-wish@v0.0.0-20200122115046-b9ea61034e4a
```

mvdan:
```
$ go mod why -m github.com/warpfork/go-wish
# github.com/warpfork/go-wish
github.com/ipld/go-ipld-prime/codec/dagcbor
github.com/polydawn/refmt/cbor
github.com/polydawn/refmt/cbor.test
github.com/warpfork/go-wish
```

Not sure why go mod graph shows that edge, to be honest.

warpfork (approved):
Seems fine to me.  "A little copying is better than a little dependency" is often true, and I think applies here.

(Mildly unfortunate it doesn't make go-wish drop entirely out of our transitive dependency tree, but ah well.  We can hope a future enclevering of go-mod might make it so; we've got it closer to happening, anyway.)

go.mod
mvdan: Also, the reason why the indirect line stays is because refmt is not a Go module yet, so then its dependencies have to be tracked here.

mvdan: FYI @warpfork if you're interested in actually dropping go-wish from the dependency graph. Most ipld-prime users shouldn't need to download nor link go-wish though, if they are only using the ipld-prime library.

masih: > Also, the reason why the indirect line stays is because refmt is not a Go module yet, so then its dependencies have to be tracked here.

Bingo; that explains the `go mod graph` output thanks @mvdan  🍻

PR #301, opened 2021-11-15
signaling ADLs in selectors

willscott:
* [x] Provide the `ExploreInterpretAs` selector  for specifying an ADL
* [x] Allow linksystem to hold a registry of known ADL reifiers
* [x] Tooling for registration of ADLs?
* [x] Testing
* [ ] Validating all traversal paths are covered.


traversal/walk.go
```diff
@@ -120,6 +120,25 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
 }
 
 func (prog Progress) walkAdv_iterateAll(n datamodel.Node, s selector.Selector, fn AdvVisitFn) error {
+	if rs, ok := s.(selector.Transformable); ok {
```
mvdan, line 123: in terms of defensive programming, using `s, ok := ...` here would make sure you only have one version of the typed selector in scope

traversal/walk.go
```diff
@@ -120,6 +120,25 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
...
+		adl := rs.Transform()
+		if prog.Cfg.LinkSystem.KnownReifiers == nil {
+			return fmt.Errorf("adl requested but not supported by link system: %s", adl)
+		}
+		reifier, ok := prog.Cfg.LinkSystem.KnownReifiers[adl]
+		if !ok {
+			return fmt.Errorf("unregistered adl requested: %s", adl)
```
mvdan, line 130:
```suggestion
			return fmt.Errorf("unregistered adl requested: %q", adl)
```

traversal/walk.go
```diff
@@ -120,6 +120,25 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
 }
 
 func (prog Progress) walkAdv_iterateAll(n datamodel.Node, s selector.Selector, fn AdvVisitFn) error {
+	if rs, ok := s.(selector.Transformable); ok {
+		adl := rs.Transform()
+		if prog.Cfg.LinkSystem.KnownReifiers == nil {
+			return fmt.Errorf("adl requested but not supported by link system: %s", adl)
```
mvdan, line 126:
```suggestion
			return fmt.Errorf("adl requested but not supported by link system: %q", adl)
```

traversal/selector/exploreInterpretAs.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Transform indicates how this selector expects to transform the current datamodel.Node.
+func (s ExploreInterpretAs) Transform() string {
+	return s.adl
+}
+
+// Transformable provides for a feature detection interface on selectors to understand when
+// and if Transformation of the datamodel.node should be attempted when performing traversals.
+type Transformable interface {
+	Transform() string
```
mvdan, line 38: I would have expected this method name to contain "reifier" or "ADL"; it's not really a generic "transform" given that it only supports whatever is registered via a string registry.

What do we call those strings? If, for example, those are "named reifiers", then I'd perhaps go with:

```
type Reifiable interface {
    NamedReifier() string
}
```

traversal/walk.go
```diff
@@ -120,6 +120,25 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
...
+		if !ok {
+			return fmt.Errorf("unregistered adl requested: %s", adl)
+		}
+
+		rn, err := reifier(linking.LinkContext{
+			Ctx:      prog.Cfg.Ctx,
+			LinkPath: prog.Path,
+		}, n, &prog.Cfg.LinkSystem)
+		if err != nil {
+			return err
```
mvdan, line 138: this seems like one of the cases where we should add context to the error :) for instance:

```
return fmt.Errorf("failed to reify node as %q: %v", adl, err)
```

traversal/selector/exploreInterpretAs.go
mvdan: Alternatively, if your argument is that this supports altering nodes in any way, not just reification, then perhaps NamedTransform.

traversal/walk.go
willscott: %q is "a single-quoted character literal " - i don't think the adl is guaranteed to be a single character?

traversal/selector/exploreInterpretAs.go
willscott: I think the use of transform here was to not overload the current `Reify` so that the two uses were named distinctly and there wasn't confusion in talking about them. I'll switch to your suggested naming though, since in this design it really is the same concept/operation

traversal/walk.go
mvdan: It has different meanings depending on the argument type. For strings:

> %q	a double-quoted string safely escaped with Go syntax

willscott:
Added a test. I think this should be ready for review

rvagg (approved):
🤷 it seems that all of the approaches to the ADL signalling problem that we've come up with have a smell about them, the main real benefit of this one is that there's actual code to do it. So maybe we just move the discussion forward by trying this out and seeing if its limitations are acceptable and leaning from it.

traversal/walk_test.go
```diff
@@ -491,3 +491,41 @@ func TestWalkBlockLoadOrder(t *testing.T) {
...
+
+	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype.Any)
+	ss := ssb.ExploreFields(func(efsb builder.ExploreFieldsSpecBuilder) {
+		efsb.Insert("linkedList", ssb.ExploreInterpretAs("linkjumper", ssb.Matcher()))
+	})
+	s, err := ss.Selector()
+	qt.Check(t, err, qt.IsNil)
+	lsys := cidlink.DefaultLinkSystem()
+	lsys.KnownReifiers = make(map[string]linking.NodeReifier)
+	lsys.KnownReifiers["linkjumper"] = customReifier
```
mvdan, line 512:
```suggestion
	lsys.KnownReifiers = map[string]linking.NodeReifier{"linkJumper": customReifier}
```

traversal/walk.go
```diff
@@ -155,6 +155,32 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
 		}
 		prog.Budget.NodeBudget--
 	}
+
+	// refiy the node if advised.
+	if rs, ok := s.(selector.Reifiable); ok {
+		adl := rs.NamedReifier()
+		if prog.Cfg.LinkSystem.KnownReifiers == nil {
```
mvdan, line 162: do we need to treat a nil KnownReifiers as a special case? it really isn't any different from a non-nil map with zero entries, if we're just reading.

if you really want to capture the "no known reifiers" case, I'd suggest `len(...) == 0`. the easiest approach is to just let it get to the "not supported" error below.

traversal/walk.go
```diff
@@ -155,6 +155,32 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
...
+		if !ok {
+			return fmt.Errorf("unregistered adl requested: %q", adl)
+		}
+
+		rn, err := reifier(linking.LinkContext{
+			Ctx:      prog.Cfg.Ctx,
+			LinkPath: prog.Path,
+		}, n, &prog.Cfg.LinkSystem)
+		if err != nil {
+			return fmt.Errorf("failed to reify node as %q: %v", adl, err)
```
mvdan, line 175: nit: use %w to wrap errors

mvdan (approved):
Sounds good to me. I'll leave it to @warpfork to give the final seal of approval, as I'm not fully up to speed with the design and plans for the selector language and API.

traversal/walk.go
willscott: i think this check is here because deference of nil map can panic

traversal/selector/fieldKeys.go
```diff
@@ -21,5 +22,6 @@ const (
 	SelectorKey_LimitNone            = "none"
 	SelectorKey_StopAt               = "!"
 	SelectorKey_Condition            = "&"
+	SelectorKey_As                   = "c"
```
warpfork, line 25: I realize this would also be a specs change, since we merged that in the `ipld/ipld` repo already, but since this will be the first implementation actually using it --

I still think we should change this key.  There's no way I would either look at the spec and expect it to use the constant "c" here, nor look at the constant "c" in a message and guess remotely correctly what it stands for.  (To wit: I just had to look it up again.)

Can we use "as"?  I don't think a second byte is going to bother anyone, and my surprise factor would go down drastically.

willscott: i'm very scared to go back to step one given that we're now 2 months in to a very simple change that has now missed 2 shipping deadlines because of the spec then implementation then wait long time for review approach.

traversal/selector/exploreInterpretAs.go
```diff
@@ -0,0 +1,65 @@
...
+// NamedReifier indicates how this selector expects to Reify the current datamodel.Node.
+func (s ExploreInterpretAs) NamedReifier() string {
+	return s.adl
+}
+
+// Reifiable provides a feature detection interface on selectors to understand when
+// and if Reification of the datamodel.node should be attempted when performing traversals.
+type Reifiable interface {
+	NamedReifier() string
+}
```
warpfork, line 39: (Reviewing out loud, not a change request)  This interface is implemented only once and used in only one place, so I wonder if it's necessary.  Since the set of types for the compiled selectors[^1] is considered closed, it seems like we could just as well use a concrete type switch?

I'm okay leaving it this way too, because it certainly works, and I don't think it gets in the way, and perhaps someone has an idea that there will be more than one selector clause in the future which asks for a reification to be performed.

[^1] - (I wish we had done https://github.com/ipld/go-ipld-prime/issues/236 by now so I could refer to what I mean by `selectorengine.*`...)

traversal/selector/exploreInterpretAs.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ParseExploreInterpretAs assembles a Selector
+// from a ExploreInterpretAs selector node
+func (pc ParseContext) ParseExploreInterpretAs(n datamodel.Node) (Selector, error) {
+	if n.Kind() != datamodel.Kind_Map {
+		return nil, fmt.Errorf("selector spec parse rejected: selector body must be a map")
+	}
+	adlNode, err := n.LookupByString(SelectorKey_As)
+	if err != nil {
+		return nil, fmt.Errorf("selector spec parse rejected: fields in ExploreFields selector must be present")
```
warpfork, line 49:
```suggestion
		return nil, fmt.Errorf("selector spec parse rejected: the 'as' field must be present in ExploreInterpretAs clause")
```

traversal/selector/exploreInterpretAs.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	next, err := n.LookupByString(SelectorKey_Next)
+	if err != nil {
+		return nil, fmt.Errorf("selector spec parse rejected: next field must be present in ExploreAll selector")
```
warpfork, line 53:
```suggestion
		return nil, fmt.Errorf("selector spec parse rejected: the 'next' field must be present in ExploreInterpretAs clause")
```

warpfork (changes requested):
Error message typo fixes.

Also tried to push them to slightly more self-similarity while at it (but honestly this whole package needs an error consistency passover, somewhat badly).

warpfork:
Replaying a conversation that @willscott and I had elsewhere, for posterity:

My initial thought was to hope we could put most logic like this -- anything that wants an ADL naming table -- into the middle of a `NodeReifier` callback, and thus push off having strong opinions about it just a little while longer.  (While, nonetheless, expecting something semistandard to still be able to emerge over time.)  This would also keep the `linking` package from having to know anything at all about ADLs, which would be pleasing.

This turns out not to fly together with this idea of Selectors carrying ADL signalling information.  `NodeReifier` doesn't get information about Selectors.

Putting the ADL naming table in the `LinkSystem` struct forces us to start making more standards about the naming.  But means the traversal code can have the knowledge of the Selectors and that naming table at the same time, and thus accomplish the goal of letting information in Selectors do the ADL signalling.

Perhaps a third option would've been adding a field to `linking.LinkContext` that allows carrying the selector there.  Then, `NodeReifier` callbacks would've gotten access to the info needed.  I'm not sure we considered this earlier.  It seems dubious though, because teaching the `linking` package about `selector`... I'm not sure if that would compile.  Maybe it even would.  But it still seems like it would be rather more of an arc through the dependency graph than I'd expect the `linking` package to have.

So we end up with this.

I'm a little nervous about getting a map in the `LinkSystem` type, when we've avoided that for every other thing `LinkSystem` needs to wrangle so far, but all the other indirections examined so far don't seem better, and we seem to have run out of patience for noodling about it any further right now.  Alright.  :ship:

warpfork (approved):
Alright, let's go

warpfork:
Future wish: more test fixtures that exercise this feature in a data-driven way and live in the `ipld/ipld` repo.

I don't feel able to hold folks up on that at the moment, but it would probably be easier to review those, and also we'll _need_ them if we expect other languages and other implementations to be able to confidently match this implementation.

traversal/selector/fieldKeys.go
warpfork: https://github.com/ipld/ipld/pull/170

warpfork:
(Followup note: we actually got fixtures that happen to exercise this, recently: they appeared as part of the specification for range clauses for bytes in https://github.com/ipld/ipld/pull/184/.  Fabulous!)

PR #302, opened 2021-11-15
Builder improvements

hannahhoward:
# Goals

This is a draft a more ipld native selector builder. It's based on the previous: https://github.com/ipld/go-ipld-prime/pull/194 with some more improvements for readability.

# Implementation

as an example, `foo/bar/**` could be built with:

```golang
Build(Path(datamodel.ParsePath("foo/bar"), ExploreRecursive(selector.RecursionLimitNode, ExploreAll(ExploreRecursiveEdge()))))
```

I keep improving this in my spare time but then never have time to finish it.... so maybe someone wants to pick it up?


rvagg:
cool, the simple path parsing in here would probably make https://github.com/ipld/go-ipld-selector-text-lite redundant

BigLep:
@hannahhoward : what would you like to do here?  There is no IPLD maintainer to finish this, but you are welcome to add tests so this could get reviewed/merged.  Alternatively, we can close.

BigLep:
Closing due to no input and no one to take it across the line.

PR #304, opened 2021-11-18
Add toInterface

### LINKED ISSUE #178: fluent: provide an API to convert an ipld.Node to a generic Go value
In other words, the reverse of `fluent.Reflect`. `Int` gives an `int64`, `Map` gives a `map[string]interface{}`, `List` gives a `[]interface{}`, and so on.

This is somewhat similar to how code-generated scalar nodes have methods like `Int() int64` as well as the usual `AsInt() (int64, error)`, but applied more generally for any node implementation and kind.

The main use case is when one has small IPLD nodes, such as any scalar kind or a small list/map, and one wants to obtain the plain Go value for easy use. Especially for debugging or testing, too.

This could potentially allocate a lot for large maps or lists, but we assume noone would do that.

Unfortunately this does mean that calling this API on a `List_String` would return a `[]interface{}` instead of `[]string`, which is unfortunate and not terribly helpful. But I don't think we can reasonably do better right now. Perhaps we can revisit that with generics.

cc @warpfork @willscott for the recent discussion
cc @gammazero since your recent question made me reconsider proposing this

MarcoPolo:
Fixes #178 

Does this seem like the general idea you were thinking of @mvdan? If so I'll add more documentation and maybe a couple more test cases.

### DIFF
--- a/fluent/toInterfaceValue.go
+++ b/fluent/toInterfaceValue.go
@@ -1,13 +1,18 @@
 package fluent
 
 import (
+	"errors"
+
 	"github.com/ipld/go-ipld-prime/datamodel"
 )
 
-func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
+var ErrInvalidKind = errors.New("invalid kind")
+var ErrUnknownKind = errors.New("unknown kind")
+
+func ToInterface(node datamodel.Node) (interface{}, error) {
 	switch k := node.Kind(); k {
 	case datamodel.Kind_Invalid:
-		panic("invalid node")
+		return nil, ErrInvalidKind
 	case datamodel.Kind_Null:
 		return nil, nil
 	case datamodel.Kind_Bool:
@@ -23,7 +28,7 @@ func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
 	case datamodel.Kind_Link:
 		return node.AsLink()
 	case datamodel.Kind_Map:
-		outMap := make(map[string]interface{})
+		outMap := make(map[string]interface{}, node.Length())
 		for mi := node.MapIterator(); !mi.Done(); {
 			k, v, err := mi.Next()
 			if err != nil {
@@ -33,15 +38,11 @@ func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
 			if err != nil {
 				return nil, err
 			}
-			vVal, err := ToInterfaceValue(v)
+			vVal, err := ToInterface(v)
 			if err != nil {
 				return nil, err
 			}
 			outMap[kVal] = vVal
-
-			if mi.Done() {
-				break
-			}
 		}
 		return outMap, nil
 	case datamodel.Kind_List:
@@ -51,17 +52,14 @@ func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
 			if err != nil {
 				return nil, err
 			}
-			vVal, err := ToInterfaceValue(v)
+			vVal, err := ToInterface(v)
 			if err != nil {
 				return nil, err
 			}
 			outList = append(outList, vVal)
-
-			if li.Done() {
-				break
-			}
 		}
 		return outList, nil
+	default:
+		return nil, ErrUnknownKind
 	}
-	panic("unhandled case in switch")
 }
--- a/fluent/toInterfaceValue.go
+++ b/fluent/toInterfaceValue.go
@@ -23,13 +23,13 @@ func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
 	case datamodel.Kind_Link:
 		return node.AsLink()
 	case datamodel.Kind_Map:
-		outMap := make(map[interface{}]interface{})
+		outMap := make(map[string]interface{})
 		for mi := node.MapIterator(); !mi.Done(); {
 			k, v, err := mi.Next()
 			if err != nil {
 				return nil, err
 			}
-			kVal, err := ToInterfaceValue(k)
+			kVal, err := k.AsString()
 			if err != nil {
 				return nil, err
 			}
--- a/fluent/toInterfaceValue.go
+++ b/fluent/toInterfaceValue.go
@@ -6,13 +6,13 @@ import (
 	"github.com/ipld/go-ipld-prime/datamodel"
 )
 
-var ErrInvalidKind = errors.New("invalid kind")
-var ErrUnknownKind = errors.New("unknown kind")
+var errInvalidKind = errors.New("invalid kind")
+var errUnknownKind = errors.New("unknown kind")
 
 func ToInterface(node datamodel.Node) (interface{}, error) {
 	switch k := node.Kind(); k {
 	case datamodel.Kind_Invalid:
-		return nil, ErrInvalidKind
+		return nil, errInvalidKind
 	case datamodel.Kind_Null:
 		return nil, nil
 	case datamodel.Kind_Bool:
@@ -60,6 +60,6 @@ func ToInterface(node datamodel.Node) (interface{}, error) {
 		}
 		return outList, nil
 	default:
-		return nil, ErrUnknownKind
+		return nil, errUnknownKind
 	}
 }
--- a/fluent/toInterfaceValue.go
+++ b/fluent/toInterfaceValue.go
@@ -1,0 +1,67 @@
+package fluent
+
+import (
+	"github.com/ipld/go-ipld-prime/datamodel"
+)
+
+func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
+	switch k := node.Kind(); k {
+	case datamodel.Kind_Invalid:
+		panic("invalid node")
+	case datamodel.Kind_Null:
+		return nil, nil
+	case datamodel.Kind_Bool:
+		return node.AsBool()
+	case datamodel.Kind_Int:
+		return node.AsInt()
+	case datamodel.Kind_Float:
+		return node.AsFloat()
+	case datamodel.Kind_String:
+		return node.AsString()
+	case datamodel.Kind_Bytes:
+		return node.AsBytes()
+	case datamodel.Kind_Link:
+		return node.AsLink()
+	case datamodel.Kind_Map:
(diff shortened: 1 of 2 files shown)


fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
+package fluent
+
+import (
+	"github.com/ipld/go-ipld-prime/datamodel"
+)
+
+func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
+	switch k := node.Kind(); k {
+	case datamodel.Kind_Invalid:
+		panic("invalid node")
```
mvdan, line 10: Panicking here seems a bit harsh :) An error woudl be better. Otherwise one would have to wrap most calls to this func with a recover, because it's hard to tell in advance if some input data or edge case program behavior could lead to an invalid kind.

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
+package fluent
+
+import (
+	"github.com/ipld/go-ipld-prime/datamodel"
+)
+
+func ToInterfaceValue(node datamodel.Node) (interface{}, error) {
```
mvdan, line 7: After having slept on my suggestion of ToInterfaceValue, I wonder if just ToInterface would be enough. All functions that return `(T, err)` already return a value.

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
...
+	case datamodel.Kind_Float:
+		return node.AsFloat()
+	case datamodel.Kind_String:
+		return node.AsString()
+	case datamodel.Kind_Bytes:
+		return node.AsBytes()
+	case datamodel.Kind_Link:
+		return node.AsLink()
+	case datamodel.Kind_Map:
+		outMap := make(map[string]interface{})
```
mvdan, line 26: nit: I think you can also use a size hint here

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
...
+		for mi := node.MapIterator(); !mi.Done(); {
+			k, v, err := mi.Next()
+			if err != nil {
+				return nil, err
+			}
+			kVal, err := k.AsString()
```
mvdan, line 32: perhaps leave a comment explaining that, for now, we assume all keys are strings.

it's an okay assumption to make right now, but we should also make it clear.

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
...
+			if err != nil {
+				return nil, err
+			}
+			vVal, err := ToInterfaceValue(v)
+			if err != nil {
+				return nil, err
+			}
+			outMap[kVal] = vVal
+
+			if mi.Done() {
```
mvdan, line 42: isn't this check duplicated?

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
...
+			if err != nil {
+				return nil, err
+			}
+			vVal, err := ToInterfaceValue(v)
+			if err != nil {
+				return nil, err
+			}
+			outList = append(outList, vVal)
+
+			if li.Done() {
```
mvdan, line 60: ditto

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,67 @@
...
+				break
+			}
+		}
+		return outList, nil
+	}
+	panic("unhandled case in switch")
```
mvdan, line 66: nit: move this under `default:`, that way you ensure you don't end up in this panic in any other of the `case`s

fluent/toInterfaceValue_test.go
```diff
@@ -0,0 +1,55 @@
+package fluent_test
+
+import (
+	"testing"
+
+	"github.com/ipld/go-ipld-prime/fluent"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+)
+
+func TestListValue(t *testing.T) {
```
mvdan, line 10: I'd encourage you to use table-driven tests, to avoid duplication. This way you can also have code coverage for all edge cases by just adding a few more entries to the table.

For instance: https://github.com/ipld/go-ipld-prime/blob/892d2f92ba8bdecef98d8863d310dbbd002b726c/node/bindnode/infer_test.go#L41

fluent/toInterfaceValue_test.go
```diff
@@ -0,0 +1,55 @@
...
+	outM := out.(map[string]interface{})
+
+	if len(a) != len(outM) {
+		t.Errorf("Mismatch in size")
+	}
+
+	for k, v := range outM {
+		if v != a[k] {
+			t.Errorf("expected %v, got %v at key %v", a[k], v, k)
+		}
```
mvdan, line 52: You'd save on keystrokes, and have better errors, if you use the qt testing library :) Like @masih did in PRs like https://github.com/ipld/go-ipld-prime/pull/291.

It's not a must to refactor all the code right now, but we might as well use it in new code.

mvdan (commented):
Looks good! The comments are mostly minor.

fluent/toInterfaceValue.go
MarcoPolo: My understanding is that IPLD maps always have string keys. It says this in the docs:
https://ipld.io/docs/intro/primer/#finer-points
> Map keys are strings.

mvdan: Ah, you're right, the docs even say that!

fluent/toInterfaceValue_test.go
```diff
@@ -0,0 +1,51 @@
...
+		})
+	}
+}
+
+func TestLink(t *testing.T) {
+	c := qt.New(t)
+	var someCid, err = cid.Parse("bafybeihrqe2hmfauph5yfbd6ucv7njqpiy4tvbewlvhzjl4bhnyiu6h7pm")
+	c.Assert(err, qt.IsNil)
+	var link = cidlink.Link{Cid: someCid}
+	var node = basicnode.NewLink(link)
```
mvdan, line 46: for consistency, use `foo := bar` rather than `var foo = bar`; the latter is generally only used when you want to explicitly declare the type - `var foo Type = bar` - or when you're declaring globals, where you can't just use `:=` statements :)

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,65 @@
+package fluent
+
+import (
+	"errors"
+
+	"github.com/ipld/go-ipld-prime/datamodel"
+)
+
+var ErrInvalidKind = errors.New("invalid kind")
+var ErrUnknownKind = errors.New("unknown kind")
```
mvdan, line 10: nit: I would avoid exposing these errors in the API for now, because once they are exposed, they are part of the API we must support. These errors should be very rare and almost always due to a bug that needs to be fixed in someone's code, so it also doesn't seem like errors that the end user should be checking against.

fluent/toInterfaceValue.go
```diff
@@ -0,0 +1,65 @@
...
+
+func ToInterface(node datamodel.Node) (interface{}, error) {
```
mvdan, line 12: Needs docs :) for instance:

```
// ToInterface converts an IPLD node to its simplest Go value equivalent.
//
// Booleans, integers, floats, strings, bytes, and links are returned as themselves,
// as per the node's AsT method. Note that nulls are returned as untyped nils.
//
// Lists and maps are returned as []interface{} and map[string]interface{}, respectively.
```

fluent/toInterfaceValue_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	qt "github.com/frankban/quicktest"
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime/fluent"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+)
+
+var roundTripTestCases = []struct {
+	desc  string
```
mvdan, line 15: nit: usually you'd name this `name`, since it's the sub-test name, not really its description

fluent/toInterfaceValue_test.go
```diff
@@ -0,0 +1,51 @@
...
+	value interface{}
+}{
+	{desc: "Number", value: 100},
+	{desc: "String", value: "hi"},
+	{desc: "Bool", value: "hi"},
```
mvdan, line 20: did you mean to make this a boolean? :)

if you use `go test -coverprofile=cover.out && go tool cover -html=cover.out`, you'll be able to switch to your file and see what lines aren't being tested at all.

fluent/toInterfaceValue_test.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, testCase := range roundTripTestCases {
+		t.Run(testCase.desc, func(t *testing.T) {
+			c := qt.New(t)
+			n, err := fluent.Reflect(basicnode.Prototype.Any, testCase.value)
+			c.Assert(err, qt.IsNil)
+			out, err := fluent.ToInterface(n)
+			c.Assert(err, qt.IsNil)
+			outJson, err := json.Marshal(out)
+			qt.Check(t, err, qt.IsNil)
+			qt.Check(t, outJson, qt.JSONEquals, testCase.value)
```
mvdan, line 36: nit: you should be able to use `qt.DeepEquals`, I think. the only other change you'd need is for the Array test case to use `[]interface{}` as its type.

mvdan (approved):
Looks good! Just some minor nits, we can merge when they're addressed.

fluent/toInterfaceValue_test.go
MarcoPolo: 👍 I was wondering if that would work after I made this change. `qt` is pretty neat.

fluent/toInterfaceValue_test.go
MarcoPolo: nice! I didn't know that

MarcoPolo:
Thanks for the great comments @mvdan

mvdan (approved):
Nice work :)

PR #309, opened 2021-12-07
Allow parsing padding in dag-json bytes fields

willscott:
While we specify that the bytes field in dag-json is canonically the un-padded variant of the base64 encoding, there are existing records / dags created on versions of this library which did use padding. We should be able to parse these nodes.


warpfork (approved):
I'm not hyped about this, and would prefer in general if our codecs supported fewer baroque variations on data... but Postel's Principle does seem to apply in this case.

(Lesson for when we work on new codecs in the future: make a much, much bigger point of making sure we are driving with fixtures and strictness from the start.)

PR #310, opened 2021-12-09
schema: add support for struct tuple reprs

mvdan:
(see commit message)

### DIFF
--- a/schema/dmt/compile.go
+++ b/schema/dmt/compile.go
@@ -211,6 +211,13 @@ func spawnType(ts *schema.TypeSystem, name schema.TypeName, defn TypeDefn) (sche
 
 			}
 			repr = schema.SpawnStructRepresentationMap2(renames, implicits)
+		case typ.Representation.StructRepresentation_Tuple != nil:
+			rp := typ.Representation.StructRepresentation_Tuple
+			if rp.FieldOrder == nil {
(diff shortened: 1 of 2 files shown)


mvdan:
Depends on https://github.com/ipld/ipld/pull/158.

schema/dmt/compile.go
```diff
@@ -211,6 +211,13 @@ func spawnType(ts *schema.TypeSystem, name schema.TypeName, defn TypeDefn) (sche
 
 			}
 			repr = schema.SpawnStructRepresentationMap2(renames, implicits)
+		case typ.Representation.StructRepresentation_Tuple != nil:
+			rp := typ.Representation.StructRepresentation_Tuple
+			if rp.FieldOrder == nil {
+				repr = schema.SpawnStructRepresentationTuple()
+				break
+			}
+			return nil, fmt.Errorf("TODO: support for tuples with  field orders in the schema package")
```
rvagg, line 220:
```suggestion
			return nil, fmt.Errorf("TODO: support for tuples with field orders in the schema package")
```

mvdan:
> can I assume that with .ipld update that it's automatically picking up the tests in specs/schemas/test/struct-tuple.yml and therefore exercising this path?

correct, it globs over all the `*.yml` test fixtures in that submodule, which is why the old ones are `*.yml-old`.

PR #311, opened 2021-12-09
node/bindnode: allow binding cidlink.Link to links

mvdan:
(see commit message)


mvdan:
FYI @adlrocha, could you test if this fixes the error for you? If not, could you please give a link to a commit/branch that reproduces the error?

rvagg (approved):

PR #312, opened 2021-12-10
Update bsadapter to context datastores

arajasek:


mvdan (approved):
Thanks! Merging right away as it seems like a no-brainer, and this blocks go-car and lotus from being compatible.

PR #317, opened 2021-12-17
schema/...: add support for enum int representations

mvdan:
(see commit message)

### DIFF
--- a/schema/dsl/parse.go
+++ b/schema/dsl/parse.go
@@ -424,7 +424,7 @@ func (p *parser) typeStruct() (*dmt.TypeDefnStruct, error) {
 		return defn, nil
 		// TODO: support custom fieldorder
 	default:
-		return nil, p.errf("unknown enum repr: %q", reprName)
+		return nil, p.errf("unknown struct repr: %q", reprName)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


schema/dsl/parse.go
```diff
@@ -639,6 +639,26 @@ func (p *parser) typeEnum() (*dmt.TypeDefnEnum, error) {
...
 		}
 		defn.Representation.EnumRepresentation_String = repr
+	case "int":
+		repr := &dmt.EnumRepresentation_Int{}
+		for i, key := range reprKeys {
+			if key == "" {
+				continue // no key; defaults to the name
+			}
+			if key[0] != '"' {
+				return nil, p.errf("enum string representation used with non-string key: %s", key)
```
willscott, line 649:
```suggestion
				return nil, p.errf("enum int representation used with non-string key: %s", key)
```

willscott (approved):

schema/dsl/parse.go
mvdan: My copy-paste is showing :)

mvdan:
I also noticed that the keys aren't optional in the `int` repr like they are in `string`, so I've removed that `continue`.

PR #324, opened 2022-01-07
schema,node/bindnode: add support for Any

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -3,6 +3,8 @@ package bindnode
 import (
 	"fmt"
 	"reflect"
+	"runtime"
+	"strings"
 
 	"github.com/ipfs/go-cid"
 	"github.com/ipld/go-ipld-prime/datamodel"
@@ -82,7 +84,38 @@ func (w *_node) Representation() datamodel.Node {
 }
 
 func (w *_node) Kind() datamodel.Kind {
-	return w.schemaType.TypeKind().ActsLike()
+	return actualKind(w.schemaType)
+}
+
+func compatibleKind(schemaType schema.Type, kind datamodel.Kind) error {
+	switch sch := schemaType.(type) {
+	case *schema.TypeAny:
+		return nil
+	default:
+		actual := actualKind(sch)
+		if actual == kind {
+			return nil
+		}
+		methodName := ""
+		if pc, _, _, ok := runtime.Caller(1); ok {
+			if fn := runtime.FuncForPC(pc); fn != nil {
+				methodName = fn.Name()
+				// Go from "pkg/path.Type.Method" to just "Method".
+				methodName = methodName[strings.LastIndexByte(methodName, '.')+1:]
+			}
+		}
+
+		return datamodel.ErrWrongKind{
+			TypeName:        schemaType.Name(),
+			MethodName:      methodName,
+			AppropriateKind: datamodel.KindSet{kind},
+			ActualKind:      actual,
+		}
+	}
+}
+
+func actualKind(schemaType schema.Type) datamodel.Kind {
+	return schemaType.TypeKind().ActsLike()
 }
 
 func (w *_node) LookupByString(key string) (datamodel.Node, error) {
@@ -110,6 +143,9 @@ func (w *_node) LookupByString(key string) (datamodel.Node, error) {
 				return datamodel.Null, nil
 			}
 			fval = fval.Elem()
+		}
+		if _, ok := field.Type().(*schema.TypeAny); ok {
+			return fval.Interface().(datamodel.Node), nil
 		}
 		node := &_node{
 			schemaType: field.Type(),
@@ -145,6 +181,9 @@ func (w *_node) LookupByString(key string) (datamodel.Node, error) {
 				return datamodel.Null, nil
 			}
 			fval = fval.Elem()
+		}
+		if _, ok := typ.ValueType().(*schema.TypeAny); ok {
+			return fval.Interface().(datamodel.Node), nil
 		}
 		node := &_node{
 			schemaType: typ.ValueType(),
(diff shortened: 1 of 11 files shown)


mvdan:
The DSL is also tested in parallel via https://github.com/ipld/ipld/pull/169.

Would like thoughts on this strategy before I polish up the patch and write a good commit message :)

mvdan:
Also, FYI @hannahhoward, this implements the "bindnode-only" version of `Any`, which is filling `datamodel.Node` values on the Go side with `basicnode.Prototype.Any`. Deferred decoding isn't part of this yet. @rvagg thinks perhaps this would be enough, as a way to not require even more features in ipld-prime to unblock graphsync, and I tend to agree.

node/bindnode/node.go
```diff
@@ -82,7 +84,38 @@ func (w *_node) Representation() datamodel.Node {
...
+			return nil
+		}
+		methodName := ""
+		if pc, _, _, ok := runtime.Caller(1); ok {
+			if fn := runtime.FuncForPC(pc); fn != nil {
+				methodName = fn.Name()
+				// Go from "pkg/path.Type.Method" to just "Method".
+				methodName = methodName[strings.LastIndexByte(methodName, '.')+1:]
+			}
+		}
```
warpfork, line 106: This is probably really nice and we probably should've done it ages ago.

But just to talk it through outloud: is this at all significantly expensive in execution time?  I don't have a good intuition for how much those `runtime.*` calls cost.

warpfork: On second thought: Nevermind.  I retract my interest to only academic.  Because this is only used in error paths, and specifically, an error path that should always be avoidable by code that bothers to check the `.Kind()` (which is about the cheapest operation one can make): it is not necessary to care about the possibility of a slow error reporting path.

node/bindnode/node.go
```diff
@@ -571,15 +643,19 @@ func (w *_assembler) AssignInt(i int64) error {
...
-			ActualKind:      w.kind(),
+	if err := compatibleKind(w.schemaType, datamodel.Kind_Float); err != nil {
+		return err
+	}
+	if _, ok := w.schemaType.(*schema.TypeAny); ok {
+		basicBuilder := basicnode.Prototype.Any.NewBuilder()
+		if err := basicBuilder.AssignFloat(f); err != nil {
+			return err
 		}
+		basicNode := basicBuilder.Build()
```
warpfork, line 654: You could use `basicnode.NewFloat` in this and similar methods for some slight improvement to terseness.  (Also it will do less allocating because it skips the builder.)

node/bindnode/node.go
```diff
@@ -625,22 +709,31 @@ func (w *_assembler) AssignBytes(p []byte) error {
...
+				// Unbox a cidlink.Link to assign to a go-cid.Cid value.
+				newVal = newVal.FieldByName("Cid")
+			} else {
+				// The target value cannot be assigned a datamodel.Link or go-cid.Cid.
+				return datamodel.ErrWrongKind{
+					TypeName:        w.schemaType.Name(),
+					MethodName:      "AssignLink",
+					AppropriateKind: datamodel.KindSet_JustLink,
+					ActualKind:      actualKind(w.schemaType),
+				}
```
warpfork, line 732: This error message doesn't seem right.  I think a custom message might be needed here.  In this branch, we're dealing: EITHER the field on the value we're reflecting on isn't something that bindnode understands as a link type, OR, bindnode might understand it as a link type, but the parameter we were given isn't go-cid.Cid.  Is that understanding correct?  The former case sorta fits ErrWrongKind; the latter doesn't at all.

(It may have been equally dubious before this diff too; I'm just catching it now.)

warpfork: Kinda suspect this whole area might be easier to reason about if it was rewritten to use earlier returns and have less leftward drift.  E.g.,

```
// If both the field and the parameter are `datamodel.Link`, easy: assign directly.
if newVal.Type().AssignableTo(w.val.Type()) {
    break
}
// ... more checks below, each breaking out as soon as possible ...
```

schema/typeMethods.go
```diff
@@ -66,6 +67,9 @@ func (t TypeEnum) RepresentationBehavior() datamodel.Kind {
 	// TODO: this should have a representation strategy switch too; sometimes that will indicate int representation behavior.
 	return datamodel.Kind_String
 }
+func (t TypeAny) RepresentationBehavior() datamodel.Kind {
+	return datamodel.Kind_Invalid // TODO: what can we possibly do here?
```
warpfork, line 71: Indeed.

Unions have a similar funny corner:

```
func (t TypeUnion) RepresentationBehavior() datamodel.Kind {
	switch t.representation.(type) {
	case UnionRepresentation_Kinded:
		return datamodel.Kind_Invalid // you can't know with this one, until you see the value (and thus can its inhabitant's behavior)!
```

I guess that if we were to dial types up to 11, this method should actually return a type that's `datamodel.Kind` plus another enum value for describing this case of "it depends".

Or we can keep (ab)using `Kind_Invalid` like this.

I don't know if we have any user complaints that provide a strong impetus one way or the other.

rvagg: can we have a `Kind_Unknown` for cases where we don't (yet) know?

rvagg:
lgtm

node/bindnode/node.go
mvdan: Correct, this does not slow down the common path at all. The cost of inspecting the call stack is noticeable; I seem to recall it's in the order of microseconds. Not something you'd want to do on every method call, but entirely reasonable when constructing an error.

node/bindnode/node.go
mvdan: Excellent callout, thank you.

node/bindnode/node.go
mvdan: Added a TODO for now, as this code was already here - this is just an indentation change. I'd rather skip this for now as I want to get graphsync working before blocking Hannah and Rod for another two weeks.

schema/typeMethods.go
mvdan: I think discussing this in a PR review isn't the best idea, and this isn't particuarly making things worse, so I've raised https://github.com/ipld/go-ipld-prime/issues/326. Marking as resolved for now :)

warpfork (approved):

PR #328, opened 2022-01-11
node/bindnode: allow mapping int-repr enums with Go integers

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -353,18 +353,34 @@ func (w *_nodeRepr) AsInt() (int64, error) {
 	case schema.UnionRepresentation_Kinded:
 		return w.asKinded(stg, datamodel.Kind_Int).AsInt()
 	case schema.EnumRepresentation_Int:
-		s, err := (*_node)(w).AsString()
-		if err != nil {
-			return 0, err
-		}
-		mapped, ok := stg[s]
-		if !ok {
+		switch w.val.Kind() {
+		case reflect.String:
+			s, err := (*_node)(w).AsString()
+			if err != nil {
+				return 0, err
+			}
+			mapped, ok := stg[s]
+			if !ok {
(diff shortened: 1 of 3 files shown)


node/bindnode/repr.go
```diff
@@ -605,13 +618,25 @@ func (w *_assemblerRepr) AssignInt(i int64) error {
...
+		for member, reprInt := range stg {
+			if int64(reprInt) != i {
+				continue
+			}
+			val := (*_assembler)(w).nonPtrVal()
+			if val.Kind() == reflect.String {
 				return (*_assembler)(w).AssignString(member)
+			} else {
+				// Short-cut to storing the repr int directly, akin to node.go's AssignInt.
+				val.SetInt(i)
```
warpfork, line 630: Are we okay with using an else here?  Maybe we should prefer to use a case switch?

As written with an "else" like this, if one has a golang `bool` field, we'll still go down this branch, and get a panic from SetInt and it won't tell the user anything useful about their schema-vs-struct mismatch.

node/bindnode/repr.go
```diff
@@ -353,18 +353,31 @@ func (w *_nodeRepr) AsInt() (int64, error) {
...
+			}
+			mapped, ok := stg[s]
+			if !ok {
+				// We assume that the schema strategy is correct,
+				// so we can only fail if the stored string isn't a valid member.
+				return 0, fmt.Errorf("AsInt: %q is not a valid member of enum %s", s, w.schemaType.Name())
+			}
+			// TODO: the strategy type should probably use int64 rather than int
+			return int64(mapped), nil
+		} else {
```
warpfork, line 369: Same question as https://github.com/ipld/go-ipld-prime/pull/328#discussion_r782329546 .

mvdan: fixed

warpfork (approved):

Issue #329, opened 2022-01-11
node/bindnode: improve UX around struct field names


mvdan:
Right now, bindnode with an IPLD schema and a Go type works for structs as long as:

* The fields on both sides are in the same order
* The names are equivalent, e.g. `fooBar` on IPLD and `FooBar` on Go (we just uppercase or lowercase the first letter)

That works fine for simple use cases, but it also has some noticeable issues:

1) It doesn't allow using custom names, such as `ID` instead of `Id` or `ADL` instead of `Adl`
2) It doesn't allow using different field orders, which could help for readability, debuggability, or even performance thanks to struct field padding

I propose that we instead do something akin to encoding/json and cbor-gen: the order of fields on the Go side shouldn't matter, and field tags could be used to override the default mapping of names. If a field tag isn't present, we would do a case-insensitive match on the IPLD Schema side, expecting exactly one match.

Another side effect of this change is that, since we don't care about order, we can start ignoring unexported fields on the Go side. This, again, is akin to how encoding/json works.

For example:

```
type T struct {
    FooBar string // will map to "foobar", "fooBar", "FooBar", or any other such case-insensitive-equal field
    metadata map[string]string // will be ignored now
    ID int64 `ipld:"requestID"` // perhaps ID is unambiguous and a better fit for the Go code
}
```

I'm not set on the field tag being `ipld`. It could also be something more explicit, like `ipldbind`. It needs to include the string `ipld` though, as something like `bindnode` would be too generic.

cc @warpfork

PR #331, opened 2022-01-11
node/bindnode: more minor improvements for graphsync

mvdan:

### DIFF
--- a/schema/typeMethods.go
+++ b/schema/typeMethods.go
@@ -175,11 +175,7 @@ func (r UnionRepresentation_Kinded) GetMember(k datamodel.Kind) TypeName {
 
 // Fields returns a slice of descriptions of the object's fields.
 func (t TypeStruct) Fields() []StructField {
-	a := make([]StructField, len(t.fields))
-	for i := range t.fields {
-		a[i] = t.fields[i]
-	}
-	return a
+	return t.fields
 }
 
 // Field looks up a StructField by name, or returns nil if no such field.
@@ -268,11 +264,7 @@ func (r StructRepresentation_Stringjoin) GetDelim() string {
 
 // Members returns a slice the strings which are valid inhabitants of this enum.
 func (t TypeEnum) Members() []string {
-	a := make([]string, len(t.members))
-	for i := range t.members {
-		a[i] = t.members[i]
-	}
-	return a
+	return t.members
 }
 
 func (t TypeEnum) RepresentationStrategy() EnumRepresentation {


mvdan:
Note to self: revisit bindnode's Next key being valid only until the next call.

node/basicnode/string.go
```diff
@@ -13,8 +13,12 @@ var (
 )
 
 func NewString(value string) datamodel.Node {
-	v := plainString(value)
-	return &v
+	return plainString(value)
+}
+
+// NewStringAddr is like NewString, but it does not allocate a new string.
```
warpfork, line 19: A short comment here about how this lets you make a mutable node, and warning that that might not be what you or others expect, would make me happier.

node/bindnode/repr.go
```diff
@@ -1026,7 +1026,8 @@ func (w *_structIteratorRepr) Next() (key, value datamodel.Node, _ error) {
 		keyStr, _ := key.AsString()
 		mappedKey := outboundMappedKey(stg, keyStr)
 		if mappedKey != keyStr {
-			key = basicnode.NewString(mappedKey)
+			w.keyStringAddr = mappedKey
+			key = basicnode.NewStringAddr(&w.keyStringAddr)
```
warpfork, line 1030: Similarly, a short comment on this iterator (or somewhere that's discoverable?) about how it's *not safe* to use this node elsewhere and assume that it's immutable seems very important.  The norm is to assume that a node is immutable and you can stitch it together in larger COW structures without second thoughts.

warpfork (commented):
Most of these make sense.

I concede on the `TypeStruct.Fields` accessor and `TypeEnum.Members` accessor for sure.  (I keep wanting Golang to have immutablity.  It doesn't.  I should stop wishing I can have nice things; and paying extra for them isn't worth it.)

I'm fairly nervous about the iterators returning something as a Node that's mutable.  Maybe with enough documentation, it's going to be survivable.  Still nerve-wracking and a fairly large divergence from the usual contract.

I wish we could make the iterators for structs and unions (e.g. what's wishing for these string wrappers) just grab pointers to "constant" strings that the `schema.Type` values already contain.  That's hard at present because _that_ package also spent effort on trying to be immutable in a language that's inimically hostile to it... but maybe it's worth drilling there instead.

node/bindnode/infer.go
```diff
@@ -285,7 +285,13 @@ func inferGoType(typ schema.Type) reflect.Type {
 }
 
 // from IPLD Schema field names like "foo" to Go field names like "Foo".
-func fieldNameFromSchema(name string) string {
+func inferFieldNameFromSchema(name string) string {
+	// Hard-code a few common names to be inferred better.
+	// TODO: is there a good and small library collecting such names?
+	switch name {
+	case "id":
+		return "ID"
```
rvagg, line 293: I'm _meh_ about this one, feels like a stretch to be doing this. `Id` isn't the worst thing to end up with is it? If you go down this path you're looking at having an entry for every known acronym out there.

warpfork: +1 tbh.  I don't relish anything that gives me flashbacks to Ruby/Rail's "Inflector".

mvdan: I definitely don't want to maintain this long-term, but without it, graphsync's schema doesn't quite work. Unless we rename its Go `ID` to `Id`, which feels unnecessary.

We can definitely do better on our end, and that's https://github.com/ipld/go-ipld-prime/issues/329. I don't want to block us for another two weeks, hence the temporary internal hack.

mvdan: On second thought, it sounds like on graphsync I'll be using separate types for bindnode, so I don't need to worry about `Id` vs `ID` anymore. I'll undo this.

mvdan: the commit is now entirely removed.

mvdan:
After talking to @warpfork about it yesterday, I ended up backing out my `NewStringAddr` change; it breaks the `Next` semantics from the MapIterator interface, and also makes us return `ipld.Node`s which get modified under the hood, which is likely going to cause bugs.

All that doesn't seem worth it for a minor speed-up right now. I might investigate other avenues which don't lose us safety in the near future.

mvdan:
> I concede on the `TypeStruct.Fields` accessor and `TypeEnum.Members` accessor for sure. (I keep wanting Golang to have immutablity. It doesn't. I should stop wishing I can have nice things; and paying extra for them isn't worth it.)

Taking this to mean that you're OK with that one commit, and given that the other two were removed, I'm going ahead and merging this :)

Issue #332, opened 2022-01-14
bindnode doesn't calculate map `Length()` properly when properties are absent


rvagg:
Can be seen @ https://github.com/ipfs/go-graphsync/pull/332

Relevant bit of schema looks like:

```ipldsch
type GraphSyncRequest struct {
  id                GraphSyncRequestID  (rename "ID")   # unique id set on the requester side
  root     optional Link                (rename "Root") # a CID for the root node in the query
  selector optional Any                 (rename "Sel")  # see https://github.com/ipld/specs/blob/master/selectors/selectors.md
  extensions        GraphSyncExtensions (rename "Ext")  # side channel information
  priority          GraphSyncPriority   (rename "Pri")  # the priority (normalized). default to 1
  cancel            Bool                (rename "Canc") # whether this cancels a request
  update            Bool                (rename "Updt") # whether this is an update to an in progress request
} representation map
```

So, using pointers for the 2 `optional` fields. But then when encoding one of these, with `nil` values, to dag-cbor I get:

```
a364426c6b7380645265717381a76249445020109ec634284a19a31e56de508bd0f063457874a063507269006443616e63f56455706474f4645273707380
```

Which, looks like this:

```
A3                                      # map(3)
   64                                   # text(4)
      426C6B73                          # "Blks"
   80                                   # array(0)
   64                                   # text(4)
      52657173                          # "Reqs"
   81                                   # array(1)
      A7                                # map(7)
         62                             # text(2)
            4944                        # "ID"
         50                             # bytes(16)
            2E039BD0020B4AE59E8983A201E5A777 # ".\x03\x9B\xD0\x02\vJ\xE5\x9E\x89\x83\xA2\x01\xE5\xA7w"
         63                             # text(3)
            457874                      # "Ext"
         A0                             # map(0)
         63                             # text(3)
            507269                      # "Pri"
         00                             # unsigned(0)
         64                             # text(4)
            43616E63                    # "Canc"
         F5                             # primitive(21)
         64                             # text(4)
            55706474                    # "Updt"
         F4                             # primitive(20)
         64                             # text(4)
            52737073                    # "Rsps"
         80                             # array(0)
```

Which is invalid, thanks to `map(7)` - i.e., here's a map with 7 fields, but I'm only going to give you 5 of them. Hence the empty `Rsps` array pulled up into the request map as if it's a field, but it's still missing one more field even then.

[...]

PR #333, opened 2022-01-14
fix(codecs): error when encoding maps whose lengths don't match entry count

rvagg:
Further to #332 I figured it'd be good to insert this sanity check so we can't produce this kind of malformed data at the codec level. I was wondering how I might add a test for this since it requires a dodgy type, but it turns out bindnode fails with AbsentOptionals when this is added, so that at least tells me this is correct.

Do you think this warrants adding a test? I think it would mean writing a full, dodgy, map type that reports the wrong `Length()`.

### DIFF
--- a/codec/dagcbor/marshal.go
+++ b/codec/dagcbor/marshal.go
@@ -162,7 +162,8 @@ func marshal(n datamodel.Node, tk *tok.Token, sink shared.TokenSink, options Enc
 func marshalMap(n datamodel.Node, tk *tok.Token, sink shared.TokenSink, options EncodeOptions) error {
 	// Emit start of map.
 	tk.Type = tok.TMapOpen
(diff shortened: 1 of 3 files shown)


mvdan (commented):
We could write a unit test, but I also think it's not super important given how a bindnode schema test is already failing on this :) The logic also seems pretty straightforward.

I'll push the bindnode fix on this branch, too, so we can merge with green checks.

mvdan:
I pushed the bindnode fix to this branch, PTAL :) cc @warpfork 

Also, your codec changes LGTM!

rvagg:
Additions lgtm

mvdan (approved):
Let's merge then; this is definitely a good change :) I have more bindnode fixes coming, so a merge will help with conflicts. @warpfork let me know if any of this doesn't look right and we can do a follow-up PR.

PR #334, opened 2022-01-19
selectors: fix for edge case around recursion clauses with an immediate edge.

warpfork:
See also the diff in the ipld/ipld repo for the new fixture,
which includes some explanation.

This would be a somewhat silly selector to write, but,
it doesn't seem to be something we should reject at compile time, either;
so, we must handle it gracefully.

Along the way, I documented some code that was sparse in comments.

I also added a shortcut to traversals so that a selector that
explicitly states it's not interested in anything (by returning a
non-nil but empty slice for its interests) is treated differently
than a nil return for interests (which means "I don't know; hit me
with everything you've got and lemme see").  This means that our
edge case here with edges (...heh) doesn't cause the traversal
to create an iterator that it doesn't really need, etc.

We turn out to need both that shortcut, *and* less panicky methods on
ExploreRecursiveEdge, *and* the edge case branch in ExploreRecursive...
because traversals are pre-order.  Decide is called first, then
Interests, and then Explore.  Therefore informing the Explore method
alone about this situation is not sufficient.

Previously, a recursion clause with an edge as its immediate and only
child would get a passing grade by the compiler (same as now),
but would panic when actually used (oh dear).

### DIFF
--- a/traversal/walk.go
+++ b/traversal/walk.go
@@ -203,6 +203,9 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
 	if attn == nil {
 		return prog.walkAdv_iterateAll(n, s, fn)
 	}
+	if len(attn) == 0 {
+		return nil
+	}
 	return prog.walkAdv_iterateSelective(n, attn, s, fn)
 
 }
--- a/traversal/selector/exploreRecursive.go
+++ b/traversal/selector/exploreRecursive.go
@@ -87,6 +87,7 @@ func (s ExploreRecursive) Interests() []datamodel.PathSegment {
 
 // Explore returns the node's selector for all fields
 func (s ExploreRecursive) Explore(n datamodel.Node, p datamodel.PathSegment) (Selector, error) {
+	// Check any stopAt conditions right away.
 	if s.stopAt != nil {
 		target, err := n.LookupBySegment(p)
 		if err != nil {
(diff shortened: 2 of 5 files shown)


warpfork:
Seealso matching specs PR: https://github.com/ipld/ipld/pull/172 .

Should fix https://github.com/filecoin-project/index-provider/issues/144 .

traversal/walk.go
```diff
@@ -203,6 +203,9 @@ func (prog Progress) walkAdv(n datamodel.Node, s selector.Selector, fn AdvVisitF
 	if attn == nil {
 		return prog.walkAdv_iterateAll(n, s, fn)
 	}
+	if len(attn) == 0 {
+		return nil
+	}
```
mvdan, line 208: personally, I don't think treating a nil slice different from an empty slice is a good idea. when consuming those values, they should be treated as identical, just like with nil vs empty maps.

the standard library does this, too; I don't think there's a single API that treats a nil slice differently from an empty slice. it will even sometimes turn one into the other, because they are considered identical. see for instance https://github.com/golang/go/issues/46415#issuecomment-849953262.

if you want the `Interests` method to signal something here, then we need another value, I think.

mvdan: > I also added a shortcut to traversals so that a selector that explicitly states it's not interested in anything (by returning a non-nil but empty slice for its interests) is treated differently than a nil return for interests (which means "I don't know; hit me with everything you've got and lemme see").

The commit message is turning my mild concern into a somewhat severe one, given that the two values that should be treated identically now mean very different things as well :)

mvdan (commented):
Changes SGTM, except the bit about nil vs empty - that seems like a red flag to me.

traversal/walk.go
warpfork: I sort of agree, but how else would we like to do this?

Refactor to `Interests() (bool, []datamodel.PathSegment)`?

I guess that's certainly within reach; we do control all the implementations, and I don't believe this is a functionally public surface area, even if it technically is (I don't think any downstreams touch this).

mvdan: If you're fairly certain this is a closed interface that downstreams shouldn't use, then I agree we can just break it. We're a v0, after all, so we might as well make use of its advantages when necessary.

An alternative is to deprecate `Interests` in favor of another method with the new signature, and type-assert whether the interface has the new method or not. Uglier for sure, but more backwards compatible.

warpfork: It's definitely not used by downstreams.  We could change the signature.

I think I'm going to skip that change for now, though.  On re-read, the documentation on the `Interests` method already phrased this fairly carefully.  Changing the code further feels like increasing the size of the diff without much real impact.

I've added another sentence of clarifying doc, though.

PR #335, opened 2022-01-19
node/bindnode: improve support for pointer types

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/infer.go
+++ b/node/bindnode/infer.go
@@ -39,6 +39,11 @@ type seenEntry struct {
 }
 
 func verifyCompatibility(seen map[seenEntry]bool, goType reflect.Type, schemaType schema.Type) {
+	// TODO(mvdan): support **T as well?
+	if goType.Kind() == reflect.Ptr {
+		goType = goType.Elem()
+	}
+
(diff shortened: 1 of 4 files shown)


mvdan:
This is essentially the same fix that I sent @rvagg nearly two days ago, but now with extensive tests.

rvagg (approved):
Supporting all those those link types are getting out of hand eh?

I'm not suggesting adding more testing but am wondering if the List and Map cases would provide enough cover to exercise the additional type kinds, like struct? Can I have a field that's a pointer to struct and have that nullable or optional (or neither, since I see `""` is a supported modifier here)?

mvdan:
> Supporting all those those link types are getting out of hand eh?

Just three types, meaning two extra lines - no big deal :)

> enough cover to exercise the additional type kinds, like struct?

Not yet; I've left a TODO about testing with more schema types. We're missing struct, enum, union, etc. I just thought this PR was already big enough as it is, and it does unblock you, as far as I recall.

The repr strategies will also need to be tested, as some of them inspect the Go values too. I'll send follow-up PRs to extend the tests and add more fixes.

Issue #342, opened 2022-01-25
Problematic panics in bindnode Builder functions


rvagg:
Currently it's easy to trigger a panic in bindnode if you feed it data that doesn't match the expected representation.

```go
package main

import (
	"github.com/ipld/go-ipld-prime"
	"github.com/ipld/go-ipld-prime/node/bindnode"
)

type Foo struct{ F int }

func main() {
	ts, _ := ipld.LoadSchemaBytes([]byte(`
		type Foo struct {
			I Int
		}
	`))
	FooProto := bindnode.Prototype((*Foo)(nil), ts.TypeByName("Foo"))
	builder := FooProto.Representation().NewBuilder()
	builder.AssignInt(100)
}
```

```
panic: TODO: schema.StructRepresentation_Map

goroutine 1 [running]:
github.com/ipld/go-ipld-prime/node/bindnode.(*_assemblerRepr).AssignInt(0xc0000c2c08, 0x64)
        /home/rvagg/go/pkg/mod/github.com/ipld/go-ipld-prime@v0.14.5-0.20220121142026-257b06219831/node/bindnode/repr.go:662 +0x2ac
main.main()
        /.../schema.go:30 +0xc5
exit status 2
```

I can see panics in:

* `AssignInt`
* `AssignFloat`
* `AssignString` (also a length mismatch panic)
* `AssignBytes`
* `AssignLink`
* `AssignNode`

For each of struct, map, list, union there's panics in the various assemblers, but I'm not yet sure how they could be triggered. There's more beyond that, but they're even less obvious.

The main concern here is that for a messaging system where the user can supply data, I want the builder to return a `datamodel.ErrWrongKind` like the typed nodes do (using the pieces mixin package for nicely consistent errors). I can deal with an error from the builder, so where there are panics could we just replace them an error return?

BigLep:
@mvdan : do we have a list of the known remaining spots?

mvdan:
@BigLep All the remaining spots that graphsync is likely to run into are now fixed as part of https://github.com/ipld/go-ipld-prime/pull/361; merging that PR will close this issue, and then I'll move onto fuzzing.

PR #344, opened 2022-01-27
a couple of panic fixes

mvdan:

Updates #342.


rvagg (approved):
very nice, more to test and do here but the builders and assemblers look fixed and tested and those are my top-line concerns since they are a vector for panic via codecs; the rest seem more likely to arise from programmer errors I think so the prioritisation here seems right. 👌

mvdan:
Thanks for the review! And yup, that's how I prioritized.

mehdi2021mp:
Bacap via wallet

mehdi2021mp:
F

PR #347, opened 2022-02-01
feat: add Garbage() Node generator

rvagg:
Bespoke garbage on a platter for all your testing and benchmarking needs. Modelled off https://github.com/rvagg/js-ipld-garbage, in fact it should produce roughly the same style of data.

It's not perfect, and it's not comprehensive, in fact it's garbage! But it's great for those tests where you need to throw random IPLD data at an API to see if it performs as you expect.

The API may need some critique, it feels comfortable to me but maybe it's too noisy with all the options and defaults.

A sampling of 10 pieces of garbage produced by this API with defaults:

```json
-0.815284567464555
[{"/":"baguqeerail77zjrvn52jkyv4wngs2x4sersmyf5xv3utayeme3x3tkn3ofpa"}]
{"-a`=D":{"/":{"bytes":"b0z9D15daxnyhWkKMwJHalyZYgBcZkApyJ+a2ofJNNVGrw"}},"LN_":{"-e@":{"[5B":{"/":"bafkreig6zjjdnwui43lfaugakyevsws7y3xv5xsa7vwe52m5mh63yrr4im"}},"?":null,"B{$t|":true,"hb💩P\ncl(_^wQ":2173616961668878594},"p`>=☺CR@s":-229867775964419454}
"3X39t6<X!v8yVX]<*QK;,zt@N[AKk+:4}{ijx/a;d9w0](]KzhoQQh3@G_vii/AXr02+a):Ir_ZVCT-U?kAP"
[{"CQ3Y":true,"U{8Y":601134118977126642,"X>zt":-0.03877292780261173,"[{\n<5P":"g\ts","h":true},{"'jS":{"/":{"bytes":"oxqlXpJCxL1uS8okyEMuBWoAu8IVKBiyTP1j+/8VOHQENsR3uNyHC9tHug"}},";h":false},[{"/":{"bytes":"ZmmdNbZzbO8YtgU"}},true,[{"UmH>Y|\tA#baANL)☺V\\l \"XY&.IwZf\"5Q~":7659331536677889261}]]]
[null,"j3M+@5oQ@#&=`x&[2X#~pEtocvN!70Y",false,{"PBj":{"/":"bafkrkmblwiedbdwui7woqf3y7zr366ywjclrcakryxuchvabgpe7huigskipbssohvpbwbhecb2ui5omz34q"},"Qk+\tG0Z;__$q💩(y&>kNJ@tqbxX_{M-;+EX7a%wb~jx>jW~s|":-1.2811529440381588}]
{"Zu":{":P":[{"/":{"bytes":"NWGx0k7eeiSb5pgn10HD1ZFnilnoCNyMFUPZQ3yQi/ZtVl1A9w"}},[{"/":{"bytes":"cLft/x96vA"}},null,4119241249922534669,"AnD0W-💩J",null,false,{"/":"bafk2bzacecjdasurd3g3fpapupw5nog3kev7mvvolp47ignmpb654n6mwanjg"}]]},"]{#!v":-10.969517253366307,"c\n?p ☺☺":-2.069906313212865}
{"?sd5|B":true,"☺0|y.mW_X":"7 dOoM;`/G<f~S%f","☺fy ":"DxW☺6tVQ9kpXrh2D25q#JxG$^☺,~#Fos]hN%3m<_w"}
{"fE}":"u5\nW~AiOu`6quwpFE6T!pf`0yj,{ad/<-iu,HA=o.2[klwn<g☺\\!Bp9v\t\nh#~IZc-)3\"N^E2*:sl0ML\nUTikQ\n3☺","s+x31":true}
[true,-0.46664879704136397,[5117978491893449015]]
```

When I was doing some work with Filecoin chain data I used this with maps and floats turned off.

### DIFF
--- a/testutil/garbage/garbage.go
+++ b/testutil/garbage/garbage.go
@@ -36,7 +36,7 @@ var (
 	generators map[datamodel.Kind]generator
 )
 
-// Garbage produces random Nodes which can be useful for testing and benchmarking. By default, the
+// Generate produces random Nodes which can be useful for testing and benchmarking. By default, the
 // Nodes produced are relatively small, averaging near the 1024 byte range when encoded
 // (very roughly, with a wide spread).
 //
@@ -47,7 +47,7 @@ var (
 // the randomness is stable across test runs, or a seed is captured in such a way that a failure
 // can be reproduced (e.g. by printing it to stdout during the test run so it can be captured in
 // CI for a failure).
-func Garbage(rand *mathrand.Rand, opts ...Option) datamodel.Node {
+func Generate(rand *mathrand.Rand, opts ...Option) datamodel.Node {
 	options := applyOptions(opts...)
 	_, n := generate(rand, options.blockSize, options)
 	return n
@@ -289,7 +289,7 @@ func Weights(weights map[datamodel.Kind]int) Option {
 
 // TargetBlockSize sets a very rough bias in number of bytes that the resulting
 // Node may consume when encoded (i.e. the block size). This is a very
-// approximate measure, but over enough repeated Garbage() calls, the resulting
+// approximate measure, but over enough repeated Generate() calls, the resulting
 // Nodes, once encoded, should have a median that is somewhere in this vicinity.
 //
 // The default target block size is 1024. This should be tuned in accordance with
--- a/testutil/garbage/garbage.go
+++ b/testutil/garbage/garbage.go
@@ -1,0 +1,304 @@
+package garbage
+
+import (
+	"math"
+	"math/rand"
+	"strings"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
+	"github.com/ipld/go-ipld-prime/must"
+	basicnode "github.com/ipld/go-ipld-prime/node/basic"
+	"github.com/multiformats/go-multihash"
+)
+
+type Options struct {
+	initialWeights map[datamodel.Kind]int
+	weights        map[datamodel.Kind]int
+	blockSize      uint64
+}
+
+type generator func(count uint64, opts Options) (uint64, datamodel.Node)
+
+type hasher struct {
+	code   uint64
+	length int
+}
+
+const charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`~!@#$%^&*()-_=+[]{}|\\:;'\",.<>?/ \t\n☺💩"
+
+var codecs []uint64
+var hashes []hasher
+var kinds datamodel.KindSet
+var generators map[datamodel.Kind]generator
+var runes []rune
+
+// Garbage produces random Nodes which can be useful for testing and benchmarking. By default, the
+// Nodes produced are relatively small, averaging near the 100 byte range when encoded
+// (very roughly, with a wide spread).
+//
+// Options can be used to adjust the average size and weights of occurances of different kinds
+// within the complete Node graph.
+func Garbage(opts ...Option) datamodel.Node {
+	options := applyOptions(opts...)
+	rand.Seed(time.Now().UnixNano())
+	_, n := generate(options.blockSize, options)
+	return n
+}
+
+func generate(count uint64, opts Options) (uint64, datamodel.Node) {
+	weights := opts.weights
+	if opts.initialWeights != nil {
+		weights = opts.initialWeights
+		opts = Options{weights: opts.weights}
+	}
+	totWeight := 0
+	for _, kind := range kinds {
+		totWeight += weights[kind]
+	}
+	r := rand.Float64() * float64(totWeight)
+	var wacc int
+	for _, kind := range kinds {
+		wacc += weights[kind]
+		if float64(wacc) >= r {
+			return generators[kind](count, opts)
+		}
+	}
+	panic("bad options")
+}
+
+func rndSize(bias uint64) uint64 {
+	// @rv: OK, I'll admit that I've completely forgotten how I arrived at this, but it
+	// generates numbers that bias toward `bias`, but not strictly, they may vary wildly
+	// but be drawn toward the lower end of the range and tend to be suitable for lengths
+	// of arrays and such
+	m := math.Abs(math.Floor(5/(1-(math.Pow(rand.Float64(), 2))) - 5 + rand.Float64()*float64(bias)))
+	return uint64(m)
+}
+
+func rndRune() rune {
+	return runes[rand.Intn(len(runes))]
+}
+
+func listGenerator(count uint64, opts Options) (uint64, datamodel.Node) {
+	len := rndSize(5)
+	lb := basicnode.Prototype.List.NewBuilder()
+	la, err := lb.BeginList(int64(len))
+	if err != nil {
+		panic(err)
+	}
+	size := uint64(0)
+	for i := uint64(0); i < len && size < count; i++ {
+		c, n := generate(count-size, opts)
+		la.AssembleValue().AssignNode(n)
+		size += c
+	}
+	err = la.Finish()
+	if err != nil {
+		panic(err)
+	}
+	return size, lb.Build()
+}
+
+func mapGenerator(count uint64, opts Options) (uint64, datamodel.Node) {
+	length := rndSize(5)
+	mb := basicnode.Prototype.Map.NewBuilder()
+	ma, err := mb.BeginMap(int64(length))
+	if err != nil {
+		panic(err)
+	}
+	size := uint64(0)
+	keys := make(map[string]struct{})
+	for i := uint64(0); i < length && size < count; i++ {
+		var key string
+		for {
+			c, k := stringGenerator(5, opts)
+			key = must.String(k)
+			if _, ok := keys[key]; !ok && len(key) > 0 {
+				keys[key] = struct{}{}
+				size += c
+				break
+			}
+		}
+		sz := count - size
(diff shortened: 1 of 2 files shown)


warpfork:
Seems pretty cool!

I'd be more interested in @mvdan's review than my own, because I think he's more familiar with golang fuzzing tech.  I think there might be things worth aligning with there, and I don't know them off the top of my head.

It seems like this is probably going to provide some noticable advantage over blind fuzzing, because for example, I can't imagine how a blind fuzzer would ever produce a correct link.

(I think it doesn't ever produce a diamond in data even when it produces links, though?  Yet?  Certainly not a blocker to merge, I'm just thinking that could be wicked Fun to see. :D)

If we do go with this: I'd want to see a randomness object passed down through the system, so we can do deterministic randomness.  If something fails, we should be able to reproduce it from a seed.

testutil/garbage/garbage.go
```diff
@@ -0,0 +1,304 @@
...
+	length int
+}
+
+const charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789`~!@#$%^&*()-_=+[]{}|\\:;'\",.<>?/ \t\n☺💩"
+
+var codecs []uint64
+var hashes []hasher
+var kinds datamodel.KindSet
+var generators map[datamodel.Kind]generator
+var runes []rune
```
mvdan, line 36: nit: you can group these with `var ( [many lines] )`, which will also give you nice alignment

testutil/garbage/garbage.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	panic("bad options")
+}
+
+func rndSize(bias uint64) uint64 {
+	// @rv: OK, I'll admit that I've completely forgotten how I arrived at this, but it
+	// generates numbers that bias toward `bias`, but not strictly, they may vary wildly
+	// but be drawn toward the lower end of the range and tend to be suitable for lengths
+	// of arrays and such
+	m := math.Abs(math.Floor(5/(1-(math.Pow(rand.Float64(), 2))) - 5 + rand.Float64()*float64(bias)))
```
mvdan, line 77: How about https://pkg.go.dev/math/rand#NormFloat64 instead?

It gives you a normal distribution, and the docs show how you can combine that with your desired mean and standard deviation. Plus, you can easily hook it up to a user-provided random source, via rand.Rand.

testutil/garbage/garbage.go
```diff
@@ -0,0 +1,304 @@
...
+	len := rndSize(5)
+	lb := basicnode.Prototype.List.NewBuilder()
+	la, err := lb.BeginList(int64(len))
+	if err != nil {
+		panic(err)
+	}
+	size := uint64(0)
+	for i := uint64(0); i < len && size < count; i++ {
+		c, n := generate(count-size, opts)
+		la.AssembleValue().AssignNode(n)
```
mvdan, line 95: nit: beware that you're missing error checks in some places like this one

testutil/garbage/garbage.go
```diff
@@ -0,0 +1,304 @@
...
+	for i := uint64(0); i < len; i++ {
+		sb.WriteRune(rndRune())
+	}
+	return len, basicnode.NewString(sb.String())
+}
+
+func bytesGenerator(count uint64, opts Options) (uint64, datamodel.Node) {
+	len := rndSize(count)
+	ba := make([]byte, len)
+	rand.Read(ba)
```
mvdan, line 159: nit: you should check for an error here, too. even math/rand can fail, for instance if a custom source is provided.

mvdan (approved):
Sounds good to me. This reminds me of https://pkg.go.dev/testing/quick#Generator, but specific to IPLD nodes.

I don't think this really overlaps with fuzzing; they're both about generating random-like inputs, but in very different ways. Fuzzing has the potential to be more generic and powerful, but this generator is domain-specific and will likely produce edge cases much quicker, akin to `testing/quick` versus Go's own native fuzzing.

I should note that fuzzing can still quickly find complex inputs like IPLD encoded links, as long as we provide it with a seed corpus that gives it a head start. If one seed input is an encoded `&Any` and another seed input is an encoded `{String:Any}`, it will not take a lot of effort to figure out it can stick the link as a map value by slicing some bytes together.

mvdan:
> I'd want to see a randomness object passed down through the system, so we can do deterministic randomness. If something fails, we should be able to reproduce it from a seed.

Yes please. This is exactly what `testing/quick` does, too. We should also strongly recommend that any use in tests should pick a static random seed, because we really want to avoid tests with random behavior by design. Those can make hunting down bugs or flakes really hard.

testutil/garbage/garbage.go
```diff
@@ -0,0 +1,304 @@
...
+//
+// The default target block size is 100. This should be tuned in accordance with
+// the anticipated average block size of the system under test.
+func TargetBlockSize(blockSize uint64) Option {
+	return func(o *Options) {
+		o.blockSize = blockSize
+	}
+}
+
+func init() {
```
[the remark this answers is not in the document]
rvagg: OK, to be honest, I didn't realise you could do a static `var foo = thing` (probably because I'm used to `var` for uninitialised/default values and `:=` for assigned), I tried a bunch of things but figured `const` was the best I could do to initialise things..

I've fixed these now but get an import cycle if I try `generators`:

```
# github.com/ipld/go-ipld-prime/testutil/garbage [github.com/ipld/go-ipld-prime/testutil/garbage.test]
./garbage.go:35:5: initialization loop:
        /home/rvagg/go/src/github.com/ipld/go-ipld-prime/testutil/garbage/garbage.go:35:5: generators refers to
        /home/rvagg/go/src/github.com/ipld/go-ipld-prime/testutil/garbage/garbage.go:93:6: listGenerator refers to
        /home/rvagg/go/src/github.com/ipld/go-ipld-prime/testutil/garbage/garbage.go:59:6: generate refers to
        /home/rvagg/go/src/github.com/ipld/go-ipld-prime/testutil/garbage/garbage.go:35:5: generators
FAIL    github.com/ipld/go-ipld-prime/testutil/garbage [build failed]
```

which is fair enough; got any suggestions for doing this outside of an `init()`?

testutil/garbage/garbage.go
rvagg: yeah, that's a neat API, have used it instead

testutil/garbage/garbage.go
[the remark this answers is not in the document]
mvdan: Ah, I had not noticed this. Seems fine to leave it in the init func, then - perhaps comment that there's an init cycle due to the generate func.

rvagg:
* Added a `rand` option.
* Tuned some internals, including using `NormFloat64()`
* Increased target block size to 1024
* Removed _most_ of the `init()` stuff (see note inline about the remaining bit)
* Added doc notes about deterministic testing — I don't quite agree with the assertion that tests should use the same randomness source, that's just going to mean that you may avoid flaky tests but hide problems that enough dice rolls will reveal, so as a compromise I've suggested that if you do use a source of randomness in testing that capturing a seed along with any test runs would be advisable so you can reproduce failures. I've patterned this behaviour in the tests here which use time as a seed.

testutil/garbage/garbage.go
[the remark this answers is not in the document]
rvagg: added comment

testutil/garbage/garbage.go
```diff
@@ -0,0 +1,316 @@
+package garbage
+
+import (
+	"math"
+	"math/rand"
```
mvdan, line 5: I would suggest to use `mathrand "math/rand"` for two reasons:

1) wherever it is used, it's clear it's math/rand, not crypto/rand
2) you can then use `rand` as a parameter name without shadowing, which can avoid bugs like unintentionally using the _global_ `rand.Func` rather than the `rand.Method` on the parameter of type `rand.Rand`.

testutil/garbage/garbage_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/codec/dagcbor"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+)
+
+func TestGarbageProducesAllKinds(t *testing.T) {
+	kindCount := make(map[datamodel.Kind]int)
+	seed := time.Now().Unix()
+	fmt.Printf("randomness seed: %v\n", seed)
```
mvdan, line 20: nit: use `t.Logf` for all of these, so the output is attached to the test result

mvdan (approved):
SGTM! I think your random seed recommendation docs are fair. I didn't mean to say that tests should always be deterministic, but that failures should be reproducible easily :)

mvdan:
To be clear, these are two last nits - happy for you to merge when you're happy with them.

testutil/garbage/garbage.go
rvagg: :thumbsup: good suggstion

testutil/garbage/garbage_test.go
rvagg: 👍

rvagg:
OK, so this is good to go, _but_ the remaining question - is `testutil/garbage` the right place for this? `testutil` only has internally useful things but this API is intended for external consumption. A top level `garbage` didn't seem appropriate though.

mvdan:
I think as a package it's fine. You're exposing multiple bits of API, so they'd all have to be like `GarbageOptions` etc if you put it in testutil. Which would also be fine - I think either is reasonable.

Something I just noticed, though - if you keep the package as-is, rename the `Garbage` API to `Generate`, so that it's then `garbage.Generate` instead of the stuttery `garbage.Garbage`.

PR #350, opened 2022-02-04
patch: first draft.

warpfork:
(Work in progress.)

The JSON Patch specification is pretty decent; this roughly follows it.

Most of the heavy lifting is done by shelling out to the existing
traversal.FocusedTransform functions.  We're just gluing that to a more
declarative API, which is more ready to be used via a serial API.

The performance of this system in the face of multiple patch operations
on the same target is probably abysmal.  This implementation is an MVP.
It's likely that a smarter implementation could combine more than one
operation application onto the same data build process, and thus
induce a *lot* fewer copies.  This also gets much trickier to implement
(maybe you need some patricia trees for ordering operations?  or more
mutable nodes for the duration of the computation, then ideally a way
to freeze them again at the end so we don't break other consistency?),
so it's "future work".  The goal of this changeset is probably going to
be limited in scope to getting the declarative API part hammered out.

As of this first commit, there's still several TODOs, regarding validation of
the patch instructions during their parse, generalizing the API to
support more codecs, and testing around upsets and moves and etc.

### DIFF
--- a/traversal/patch/patch_serial.go
+++ b/traversal/patch/patch_serial.go
@@ -1,0 +1,73 @@
+package patch
+
+import (
+	"io"
+	"strings"
+
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+
+	"github.com/ipld/go-ipld-prime/codec/json"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/schema"
+	schemadmt "github.com/ipld/go-ipld-prime/schema/dmt"
+	schemadsl "github.com/ipld/go-ipld-prime/schema/dsl"
+)
+
+var ts = func() schema.TypeSystem {
+	sch, err := schemadsl.Parse("", strings.NewReader(
+		// This could be more accurately modelled as an inline union,
+		// but that seems like work, given how high the overlap is.
+		`
+		type Operation struct {
+			op String
+			path String
+			value optional Any
+			from optional String
+		}
+		type OperationList [Operation]
+	`))
+	if err != nil {
+		panic(err)
+	}
+	var ts schema.TypeSystem
+	ts.Init()
+	if err := schemadmt.Compile(&ts, sch); err != nil {
+		panic(err)
+	}
+	return ts
+}()
+
+// FIXME this should surely accept a codec.Decoder parameter
+func LoadPatch(r io.Reader) ([]Operation, error) {
+	npt := bindnode.Prototype((*[]operationRaw)(nil), ts.TypeByName("OperationList"))
+	nb := npt.Representation().NewBuilder()
+	if err := json.Decode(nb, r); err != nil {
+		return nil, err
+	}
+	opsRaw := bindnode.Unwrap(nb.Build()).(*[]operationRaw)
+	var ops []Operation
+	for _, opRaw := range *opsRaw {
+		// TODO check the Op string
+		op := Operation{
+			Op:   Op(opRaw.Op),
+			Path: datamodel.ParsePath(opRaw.Path),
+		}
+		if opRaw.Value != nil {
+			op.Value = *opRaw.Value
+		}
+		if opRaw.From != nil {
+			op.From = datamodel.ParsePath(*opRaw.From)
+		}
+		ops = append(ops, op)
+	}
+	return ops, nil
+}
+
+// operationRaw is roughly the same structure as Operation, but more amenable to serialization
+// (it doesn't use high level library types that don't have a data model equivalent).
+type operationRaw struct {
+	Op    string
+	Path  string
+	Value *datamodel.Node
+	From  *string
+}
--- a/datamodel/path.go
+++ b/datamodel/path.go
@@ -213,6 +213,14 @@ func (p Path) Last() PathSegment {
 	return p.segments[len(p.segments)-1]
 }
 
+// Pop returns a path with all segments except the last.
+func (p Path) Pop() Path {
+	if len(p.segments) < 1 {
+		return Path{}
+	}
+	return Path{p.segments[0 : len(p.segments)-1]}
+}
+
 // Shift returns the first segment of the path together with the remaining path after that first segment.
 // If applied to a zero-length path, it returns an empty segment and the same zero-length path.
 func (p Path) Shift() (PathSegment, Path) {
--- a/traversal/focus.go
+++ b/traversal/focus.go
@@ -171,6 +171,8 @@ func (prog *Progress) get(n datamodel.Node, p datamodel.Path, trackProgress bool
 // a copy-on-write fashion -- and the FocusedTransform function as a whole will
 // return a new Node containing identical children except for those replaced.
 //
+// Returning nil from the TransformFn as the replacement node means "remove this".
+//
 // FocusedTransform can be used again inside the applied function!
 // This kind of composition can be useful for doing batches of updates.
 // E.g. if have a large Node graph which contains a 100-element list, and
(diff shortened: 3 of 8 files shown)


traversal/patch/patch_serial.go
```diff
@@ -0,0 +1,73 @@
...
+	schemadsl "github.com/ipld/go-ipld-prime/schema/dsl"
+)
+
+var ts = func() schema.TypeSystem {
+	sch, err := schemadsl.Parse("", strings.NewReader(
+		// This could be more accurately modelled as an inline union,
+		// but that seems like work, given how high the overlap is.
+		`
+		type Operation struct {
+			op String
```
rvagg, line 22: String representation enum would be nice here, also an opportunity to consider whether all those operations really are that useful; I'm instinctively dubious about basically everything except `add` and `remove`, the rest feel like complexity for the sake of theoretical efficiency. Maybe there's a case but they really turn the whole logic set (and therefore test suite for this) into a complex tangle.

warpfork: Yeah, this might get schema'd harder; I whipped this off the current state of this quite quickly.  I should consider putting [bindnode.ProduceGoTypes](https://pkg.go.dev/github.com/ipld/go-ipld-prime/node/bindnode#ProduceGoTypes) to work.

The sheer number of operations: I admit I blinked at the JSON Patch spec at first too, but think they're all justified.  'replace' is actually needed, since 'add' is supposed to error rather than upsert if the target path exists.  'move' and 'copy' are a little more arguable, but can't really be replaced if the situation is that you aren't already holding the document contents.  (Consider using 'move' to shift content in a HAMT from under one map key to another.  You wouldn't want to download the document just to give it a different key.)

I think we may end with slightly more operations, eventually.  E.g. I think I'd like an 'add2' that will create parent paths without a fuss, and upsert by default (perhaps an optional bool field to disable upsert).  We may also need to add some features to describe the distinction between "update this document that is the root of a block that is linked to" vs "replace this link node with something else entirely".  But sticking with the JSON Patch RFC for a baseline still seems like an easy choice.

warpfork:
This now appears to work with all six operations: add, replace, remove, copy, move, and test.  Fixtures exist in the ipld/ipld repo for each.

warpfork:
Probable future work:
- Even more fixtures!  Especially those that probe error paths.
  - ... it might be nice to adopt the [serum errors](https://github.com/serum-errors/serum-spec) conventions here, and then as a bonus even get to use the [go serum static analyzer](https://github.com/serum-errors/go-serum-analyzer) for the developer assistance and enforced high code quality.
- I'd like to standardize the error codes and how they're serialized.
  - ... it might be nice to do something about https://github.com/ipld/go-ipld-prime/issues/360 first, because that's gnarly to try to paint over at a higher level.
- Tests still needed for patches that apply across links.  (I'm fairly sure `traversal.FocusedTransform`, which I built this feature on, already supports this -- but it's not tested at this level.)
  - ... it might be nice to build some more helper features around go-testmark to help manage fixtures that are stored by content addressed hunk names (currently you can of course do this, but manually it would be a bit ishy to write such fixtures).
- There's still a variety of TODOs scattered in the implementation about better validation and error handling paths in general.

I'm not sure if these are worth doing immediately, or if this PR is worth landing as-is and iterating more in future PRs.  (Leaning towards the latter, especially because having long-running branches in the ipld/ipld specs&meta repo becomes a bit frictional.)

warpfork:
Status of this: I think I call this "alpha".  The current behavior appears to be correct so far.  The tests we have cover key features.  I'd still consider it relatively experimental and the odds of bugs relatively high -- "use at your own risk".  (I want *many* more test fixtures probing corner cases, in the long run, before promising complete stability.)

I think we may be able to merge this, with those caveats.

BigLep:
What are the next steps here?  Can this be merged after feedback is incorporated?

RangerMauve:
TODO post merge: Update `ipld.io/docs/intro/hello-world.md`

traversal/patch/patch_serial.go
RangerMauve: Added the enum for the Op.

Should I be copying the schema into `ipld/ipld` somewhere as it gets updated?

RangerMauve:
Two questions:

- I'm not seeing the errors for the operations when I try to run the tests locally on my machine. What can I do to debug this? Is it a golang version? I'm on `go version go1.17.9 linux/amd64`.

- What can I do about the following errors I'm getting when running `go test ./...`?

```
--- FAIL: TestRoundtripSchemaSchema (0.00s)
    roundtrip_test.go:23: 
        error:
          got non-nil error
        got:
          e"open ../../.ipld/specs/schemas/schema-schema.ipldsch.json: no such file or directory"
        stack:
          /home/mauve/programming/go-ipld-prime/schema/dmt/roundtrip_test.go:23
            qt.Assert(t, err, qt.IsNil)
        
FAIL
FAIL	github.com/ipld/go-ipld-prime/schema/dmt	0.002s
--- FAIL: TestParseSchemaSchema (0.00s)
    parse_test.go:30: 
        error:
          got non-nil error
        got:
          e"open ../../.ipld/specs/schemas/schema-schema.ipldsch: no such file or directory"
        stack:
          /home/mauve/programming/go-ipld-prime/schema/dsl/parse_test.go:30
            qt.Assert(t, err, qt.IsNil)
        
--- FAIL: TestParse (0.00s)
    parse_test.go:63: 
        error:
          unexpected success
        len(got):
          int(0)
        got:
          []string(nil)
        want length:
          <same as "len(got)">
        stack:
          /home/mauve/programming/go-ipld-prime/schema/dsl/parse_test.go:63
            qt.Assert(t, matches, qt.Not(qt.HasLen), 0)
        
FAIL
FAIL	github.com/ipld/go-ipld-prime/schema/dsl	0.003s
```

rvagg:
@RangerMauve we have a git submodule sync problem here; both ipld/ipld and `master` have moved on from the last commit here.

If you run `git submodule update --init --recursive` you should reset back to the submodule commit this branch has and you should get the errors CI is showing here.

I think what you probably should do is rebase both https://github.com/ipld/ipld/pull/187 and then here against current master versions and get them both in sync.

Then, if you replace your enum in the schema with:

```
		type Op enum {
			| add
			| remove
			| replace
			| move
			| copy
			| test
		}
```

it should work OK, bindnode is using the strings `"Op_Add"` etc. and they're just getting converted straight to `Op`s as `Op("Op_Add")` which is failing in the `switch` in eval.go.

Also as an aside, now I look at this conversion code in parse.go, I reckon I can make `operationRaw` go away with https://github.com/ipld/go-ipld-prime/pull/414, since the main problem seems to be the conversion of `string` to `datamodel.Path`. So it's nice to see another use-case for doing that to avoid intermediate types.

rvagg:
Also, the `embed` trick might be worthwhile here, see: https://github.com/ipfs/go-graphsync/tree/main/message/ipldbind

import `_ "embed"` then you can `//go:embed schema.ipldsch` a file into `[]byte` so you can keep the schema file separate and it gets loaded automatically.

rvagg:
I've cleaned this up a bit:

* rebased to master
* fixed the enum
* extracted the schema to a separate file
* cleaned up bindnode use a bit
* added an "EXPERIMENTAL" notice at the top of the package
* updated https://github.com/ipld/ipld/pull/187 (rebased to ipld/ipld master and updated submodule commit here); we have a green.


I reckon this might be mergeable along with https://github.com/ipld/ipld/pull/187, what do you reckon @RangerMauve?

RangerMauve:
Thanks @rvagg! I was away last week so I'm just getting into this now.

I was thinking some link traversal tests would be useful for patch before folks start building on it, but I guess that could be done in a separate PR so we can get this out the door. 🤔

traversal/patch/patch_serial.go
RangerMauve: The `_ "embed"` trick is great for this!

rvagg:
> Also, I'm seeing a bunch of warnings saying that there's lines not covered by tests. Is that something we should address or is it expected?

The codecov is new, comes with the rebase to master where we now have the unified-ci workflow! So we get to feel that little bit more guilty about poor test coverage.

The uncovered bits are all for error cases, which are pretty annoying to cover in their entirety; and for a first pass at an experimental feature I don't think it's worth chasing those down here - at least that's not been the past practice in this repo where coverage hasn't been a super high priority. The one case where it's not for errors is the `"-"` path segment case where there's two paths where that would invoke if it were tested. It's in the comments:

> There's also a special case for "-", which means "append to the end of the list".

But no tests for it. We maybe should add one for that since we have two codepaths that are relevant.

RangerMauve:
Tests pass, merging.

PR #354, opened 2022-02-07
Add a TraversingLinksystem supporting traversal resumption

willscott:
This is a proposal for how we would extend the current Progress / traversal structure to optionally support efficient resumption of traversals.

The trade-off is that more state is retained in order to know which links have been seen by which points in the traversal. This proposal handles that decision by allowing the user to wrap their `Progress` via `WithTraversingLinksystem`. This method wraps the linksystem block loader in order to track and appropriately skip forward when a subsequent `resumer` is used.

I'd be happy if we could arrive at a more intuitive design, but this seems like the path of least resistance for maintaining current API compatibility.

### DIFF
--- a/traversal/traversing_linksystem.go
+++ b/traversal/traversing_linksystem.go
@@ -1,0 +1,146 @@
+package traversal
+
+import (
+	"io"
+
+	"github.com/ipld/go-ipld-prime"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/linking"
+)
+
+type pathNode struct {
+	link     datamodel.Link
+	children map[datamodel.PathSegment]*pathNode
+}
+
+func newPath(link datamodel.Link) *pathNode {
+	return &pathNode{
+		link:     link,
+		children: make(map[datamodel.PathSegment]*pathNode),
+	}
+}
+
+func (pn pathNode) addPath(p []datamodel.PathSegment, link datamodel.Link) {
+	if len(p) == 0 {
+		return
+	}
+	if _, ok := pn.children[p[0]]; !ok {
+		child := newPath(link)
+		pn.children[p[0]] = child
+	}
+	pn.children[p[0]].addPath(p[1:], link)
+}
+
+func (pn pathNode) allLinks() []datamodel.Link {
+	if len(pn.children) == 0 {
+		return []datamodel.Link{pn.link}
+	}
+	links := make([]datamodel.Link, 0)
+	if pn.link != nil {
+		links = append(links, pn.link)
+	}
+	for _, v := range pn.children {
+		links = append(links, v.allLinks()...)
+	}
+	return links
+}
+
+// getPaths returns reconstructed paths in the tree rooted at 'root'
+func (pn pathNode) getLinks(root datamodel.Path) []datamodel.Link {
+	segs := root.Segments()
+	if len(segs) == 0 {
+		if pn.link != nil {
+			return []datamodel.Link{pn.link}
+		}
+		return []datamodel.Link{}
+	} else if len(segs) == 1 {
+		// base case 1: get all paths below this child.
+		next := segs[0]
+		if child, ok := pn.children[next]; ok {
+			return child.allLinks()
+		} else if !ok {
+			return []datamodel.Link{}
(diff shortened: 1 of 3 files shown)


traversal/traversing_linksystem.go
```diff
@@ -0,0 +1,146 @@
...
+// WithTraversingLinksystem extends a progress for traversal such that it can
+// subsequently resume and perform subsets of the walk efficiently from
+// an arbitrary position within the selector traversal.
+func WithTraversingLinksystem(p *Progress) (TraverseResumer, error) {
+	ts := &traversalState{
+		underlyingOpener: p.Cfg.LinkSystem.StorageReadOpener,
+		position:         0,
+		pathOrder:        make(map[int]datamodel.Path),
+		pathTree:         newPath(nil),
+		target:           nil,
```
mvdan, line 141: nit: you can omit zero values from composite literals

traversal/traversing_linksystem.go
```diff
@@ -0,0 +1,146 @@
...
+
+// getPaths returns reconstructed paths in the tree rooted at 'root'
+func (pn pathNode) getLinks(root datamodel.Path) []datamodel.Link {
+	segs := root.Segments()
+	if len(segs) == 0 {
+		if pn.link != nil {
+			return []datamodel.Link{pn.link}
+		}
+		return []datamodel.Link{}
+	} else if len(segs) == 1 {
```
mvdan, line 56: nit: use a switch

traversal/traversing_linksystem.go
```diff
@@ -0,0 +1,146 @@
...
+		// base case 1: get all paths below this child.
+		next := segs[0]
+		if child, ok := pn.children[next]; ok {
+			return child.allLinks()
+		} else if !ok {
```
mvdan, line 61: nit: this can just be an else

mvdan (commented):
What is the cost of resuming a traversal? As a user, that would be my main worry. For example, if I have a deeply nested graph, and I want to resume at a deep point, will the resumption need to do work to get back there? One can imagine a similar situation with very wide nodes, such as a very long list where I want to resume a traversal towards the end of the list.

willscott:
> What is the cost of resuming a traversal

Instead of keep track of all previous links that have been followed, which is already done in a progress, you are now keeping track of the ipld.PathSegment map as well.
If you have many small / highly-pathed objects relative to the number of links this could be larger than the existing state, but it shouldn't be that bad overall.

hannahhoward (commented):
Fascinating. 

I think the code works on its own but I believe the comment I left would make it slightly more efficent.

warpfork:
I think my main concern/ask for either this or https://github.com/ipld/go-ipld-prime/pull/358/ isn't even so much the code at this point -- it's: can we get a serial fixture to appear in [ipld/ipld](https://github.com/ipld/ipld/) with it?

I realize some of this may feel like it's golang specific APIs, but I doubt it's going to stay that way -- other languages will need to replicate the behavior at some point, and we might as well be ready for them with fixtures.  Also, personally, I'm finding serial fixtures a lot easier to review for this stuff.

There's a `spec_test.go` file in the same directory that should provide an example, although it may need some extending to test a new feature like this.

willscott:
I would like to leave this as a golang specific API at this point, and not attempt to define a specification for how resumption is exposed that is cross-language standardized until we understand what is appropriate in other language. It feels generally like we are not proposing a change in spec or standard at this point, but rather providing a language-specific convenience method.

mvdan:
@willscott regarding the cost - sounds good. Perhaps it would be good to summarize this in the docs; as someone reading the API docs and not knowing how traversals and selectors are implemented, it would be very hard to estimate how expensive a resumption could be.

willscott:
This is fairly separable, and I plan to close this PR and move this code with the go-car PR that makes use of it. It's probably not a generic interface we want to have to keep supporting forever at this library level, but it was useful for a couple related PRs that have been set up already using the same structure.

BigLep:
@willscott : this came up during 2022-04-05 triage.  Are you able to close this now?

willscott:
Closing as this logic has now been fully transferred to https://github.com/ipld/go-car/pull/291

PR #355, opened 2022-02-09
add top-level "go value with schema" example

mvdan:
(see commit message)


mvdan:
@willscott this is the same example I mentioned to you before the holidays, just recovered it from the local git stash :)

warpfork (approved):
:100: :100: :+1:  :+1:  :+1: :ship: :shipit: :ship: :shipit: :ship: :shipit:

warpfork:
... I think this got lost in my event firehose.  Looks great and is easy to say yes to.

I'm going to give this a [non-author +2](https://github.com/ipld/go-ipld-prime/issues/370#issuecomment-1055408886) and merge it immediately.

PR #358, opened 2022-02-12
Implement option to start traversals at a path

hannahhoward:
# Goals

This is an alternative way to "resume selector traversals"  - similar to https://github.com/ipld/go-ipld-prime/pull/354 - but tailored to different needs. 

https://github.com/ipld/go-ipld-prime/pull/354 is ideal for someone writing a CAR file who needs to stop on a block boundary and wants to retain information about blocks traversed so they can resume writing at a point.

This is ideal for someone who wants to maintain as little state information as possible and simply execute a selector as if it started at a specific path -- even the visits will not get called before the path. It's deal for a scenario like graphsync resumption -- where transmitting a list of "do not send these blocks" has proved quite problematic in the past (cause of the network overhead on large DAGs)

# Implementation

It's really simple: you can now set a StartAtPath on the TraversalConfig

The progress keeps an additional piece of state of whether you're "past the startup path" at which point you can just not worry about checking

The code is a wee bit imperative and tricky, but it's tested and works.

# For Discussion

Why not do it in a selector itself? Selectors are stateless and I don't believe you can execute this in a selector as a result.

### DIFF
--- a/traversal/fns.go
+++ b/traversal/fns.go
@@ -37,7 +37,7 @@ type Progress struct {
 		Path datamodel.Path
 		Link datamodel.Link
 	}
-	PastStartAtPath bool
+	PastStartAtPath bool                        // Indicates whether the traversal has progressed passed the StartAtPath in the config -- use to avoid path checks when inside a sub portion of a DAG that is entirely inside the "not-skipped" portion of a traversal
 	Budget          *Budget                     // If present, tracks "budgets" for how many more steps we're willing to take before we should halt.
 	SeenLinks       map[datamodel.Link]struct{} // Set used to remember which links have been visited before, if Cfg.LinkVisitOnlyOnce is true.
 }
--- a/traversal/fns.go
+++ b/traversal/fns.go
@@ -47,7 +47,7 @@ type Config struct {
 	LinkSystem                     linking.LinkSystem             // LinkSystem used for automatic link loading, and also any storing if mutation features (e.g. traversal.Transform) are used.
 	LinkTargetNodePrototypeChooser LinkTargetNodePrototypeChooser // Chooser for Node implementations to produce during automatic link traversal.
 	LinkVisitOnlyOnce              bool                           // By default, we visit across links wherever we see them again, even if we've visited them before, because the reason for visiting might be different than it was before since we got to it via a different path.  If set to true, track links we've seen before in Progress.SeenLinks and do not visit them again.  Note that sufficiently complex selectors may require valid revisiting of some links, so setting this to true can change behavior noticably and should be done with care.
-	StartAtPath                    datamodel.Path                 // allows a traversal to skip execution of everything it would traverse before a given path
+	StartAtPath                    datamodel.Path                 // If set, causes a traversal to skip forward until passing this path, and only then begins calling visit functions.  Block loads will also be skipped wherever possible.
 }
 
 type Budget struct {
--- a/traversal/fns.go
+++ b/traversal/fns.go
@@ -37,8 +37,9 @@ type Progress struct {
 		Path datamodel.Path
 		Link datamodel.Link
 	}
-	Budget    *Budget                     // If present, tracks "budgets" for how many more steps we're willing to take before we should halt.
-	SeenLinks map[datamodel.Link]struct{} // Set used to remember which links have been visited before, if Cfg.LinkVisitOnlyOnce is true.
+	PastStartAtPath bool
+	Budget          *Budget                     // If present, tracks "budgets" for how many more steps we're willing to take before we should halt.
+	SeenLinks       map[datamodel.Link]struct{} // Set used to remember which links have been visited before, if Cfg.LinkVisitOnlyOnce is true.
 }
 
 type Config struct {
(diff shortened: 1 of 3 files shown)


warpfork:
I think my main concern/ask for either this or https://github.com/ipld/go-ipld-prime/pull/354 isn't even so much the code at this point -- it's: can we get a serial fixture to appear in [ipld/ipld](https://github.com/ipld/ipld/) with it?

I realize some of this may feel like it's golang specific APIs, but I doubt it's going to stay that way -- other languages will need to replicate the behavior at some point, and we might as well be ready for them with fixtures.  Also, personally, I'm finding serial fixtures a lot easier to review for this stuff.

There's a `spec_test.go` file in the same directory that should provide an example, although it may need some extending to test a new feature like this.

traversal/fns.go
```diff
@@ -37,15 +37,17 @@ type Progress struct {
 		Path datamodel.Path
 		Link datamodel.Link
 	}
-	Budget    *Budget                     // If present, tracks "budgets" for how many more steps we're willing to take before we should halt.
-	SeenLinks map[datamodel.Link]struct{} // Set used to remember which links have been visited before, if Cfg.LinkVisitOnlyOnce is true.
+	PastStartAtPath bool
```
warpfork, line 40: Would like a comment here about what this state means.

traversal/fns.go
```diff
@@ -37,15 +37,17 @@ type Progress struct {
...
 	Ctx                            context.Context                // Context carried through a traversal.  Optional; use it if you need cancellation.
 	LinkSystem                     linking.LinkSystem             // LinkSystem used for automatic link loading, and also any storing if mutation features (e.g. traversal.Transform) are used.
 	LinkTargetNodePrototypeChooser LinkTargetNodePrototypeChooser // Chooser for Node implementations to produce during automatic link traversal.
 	LinkVisitOnlyOnce              bool                           // By default, we visit across links wherever we see them again, even if we've visited them before, because the reason for visiting might be different than it was before since we got to it via a different path.  If set to true, track links we've seen before in Progress.SeenLinks and do not visit them again.  Note that sufficiently complex selectors may require valid revisiting of some links, so setting this to true can change behavior noticably and should be done with care.
+	StartAtPath                    datamodel.Path                 // allows a traversal to skip execution of everything it would traverse before a given path
```
warpfork, line 50: Nit: had different capitalization than sibling comments.

Here's a slightly different take; does this describe the behaviour accurately?

```suggestion
	StartAtPath                    datamodel.Path                 // If set, causes a traversal to skip forward until passing this path, and only then begins calling visit functions.  Block loads will also be skipped wherever possible.
```

warpfork (approved):
I think I understood this, and it does what it does with very minimal added state, so I give it a tentative +1.

I'd still be much happier if we got purely serial fixtures for this, though.  If I look at these tests and I'm not a go programmer, I don't think I could easily reverse engineer what they mean.

traversal/fns.go
warpfork: ... I wonder if some comment is also needed for how this interplays with the Transform functions?  (e.g. it doesn't -- i think.)

hannahhoward:
re: serial fixtures

I think I could certainly implement a serial fixture test BUT I'm hesitant to do so.

I read over our selector spec and it actually do NOT outline our assumptions about execution and ordering: name serial, depth first traversal. This strikes me as a more significant issue if we want to put any kind of "start at path" into the spec -- how can I know what skipping to a path means when I haven't defined the order of the nodes visited? 

The current fixtures are technically valid with the lack of ordering for execution in the spect cause the header simply reads "expected visit events" (while the actual go test looks at ordering, the language-ing of "expected visit events" does not explicitly mean 'in this order').

Anyway, this seems like spec work that should be done -- especially since GraphSync's DDOS protections rely on predictable ordering. But it feels outside the auspices of this PR to do it.

hannahhoward:
I believe this is ready to merge pending discussion of spec tests.

warpfork:
Alright, SGTM.

FWIW, we do have several selector fixtures which touch on ordering.  But it may well be true that it's not highlighted as much as it should be in other parts of the spec.

I'll quit pushing on spec improvements for here for now.

warpfork:
To clarify: this is effectively mergable.

If we use this as a dry run of [the recently proposed contribution and merge policy](https://github.com/ipld/go-ipld-prime/issues/370#issuecomment-1055408886):

- I already +1'd this.  (I'm declining to give it a +2 because I reviewed it only briefly, and because I'm not thrilled about the lack of language-agnostic fixtures that are easy to review.)
- You can self+1 this, which gives a net +2, and that's good enough to merge it.
- Or you can poke someone else (perhaps @willscott ; I know he's already looked at it) to give it a +1, gives a net +2, and that's good enough to merge it.

I'm not clicking merge, because at the current count, it's got a net +1 (which is from me; the author can do a self+1, but it's not the default).  But you have two routes for getting it to +2, and when it does so, you have the merge button available to click.

willscott (approved):

willscott:
(have been bugged. looks reasonable to me.)

Issue #357, opened 2022-02-11
fuzz node/bindnode and schema/dmt via codec/dagcbor


mvdan:
**done criteria**: can run fuzzer for ~5 minutes and not find any panics.

A Go 1.18 fuzz func can have two `[]byte` inputs, both decoded as dag-cbor, where the first is an IPLD schema-schema in DMT form, and the second is an IPLD node to decode.

This will help us uncover panics, hangs, and other forms of fatal bugs in:

* The schema/dmt package, such as its Compile API
* The node/bindnode package, especially all of its edge cases using reflection
* The dagcbor package, for its decoding and perhaps encoding too, as a bonus

The fuzzer could be a bit expensive to run and find new interesting inputs, but if we supply the two sides with a good quantity of seed corpus entries, we should be fine.

PR #361, opened 2022-02-16
node/bindnode: cover even more edge case panics

### LINKED ISSUE #342: Problematic panics in bindnode Builder functions

mvdan:
(see commit message)

of the fuzzing effort in #357.
Fixes #342.

### DIFF
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -164,6 +164,9 @@ func (w *_nodeRepr) LookupByIndex(idx int64) (datamodel.Node, error) {
 		return w.asKinded(stg, datamodel.Kind_List).LookupByIndex(idx)
 	case schema.StructRepresentation_Tuple:
 		fields := w.schemaType.(*schema.TypeStruct).Fields()
+		if idx < 0 || int(idx) >= len(fields) {
+			return nil, datamodel.ErrNotExists{Segment: datamodel.PathSegmentOfInt(idx)}
+		}
 		field := fields[idx]
 		v, err := (*_node)(w).LookupByString(field.Name())
 		if err != nil {
@@ -259,7 +262,9 @@ func (w *_mapIteratorRepr) Done() bool {
 }
 
 func (w *_nodeRepr) ListIterator() datamodel.ListIterator {
-	// TODO: support kinded unions?
+	if stg, ok := reprStrategy(w.schemaType).(schema.UnionRepresentation_Kinded); ok {
+		w = w.asKinded(stg, datamodel.Kind_List)
+	}
 	switch reprStrategy(w.schemaType).(type) {
 	case schema.StructRepresentation_Tuple:
 		typ := w.schemaType.(*schema.TypeStruct)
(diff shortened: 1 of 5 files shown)


node/bindnode/repr.go
```diff
@@ -537,15 +542,24 @@ type _assemblerRepr struct {
...
-func assemblerRepr(am datamodel.NodeAssembler) *_assemblerRepr {
-	return (*_assemblerRepr)(am.(*_assembler))
+func assemblerRepr(am datamodel.NodeAssembler) datamodel.NodeAssembler {
+	switch am := am.(type) {
+	case *_assembler:
+		return (*_assemblerRepr)(am)
+	case _errorAssembler:
+		return am
+	default:
+		panic(fmt.Sprintf("unexpected NodeAssembler type: %T", am))
```
rvagg, line 552: I assume that this means that you don't see a valid path to reach here in the current code? it's got to be either an `*_assembler`, or an `_errorAssembler` or someone's done something wrong and messed with the internals somehow?

rvagg (approved):
this is looking pretty good I think; a lot of it is deeper in this code than I currently understand but it certainly seems to be in the right direction and I don't see an obvious problem

node/bindnode/repr.go
mvdan: Correct. And if we get there, I have basically no other option but to fatally fail.

I guess I could use an errorAssembler, but I also want these unexpected paths caused by internal bugs to loudly fail, because they need to be fixed. The way to ensure that a panic doesn't happen isn't by removing the panic, but rather by testing and fuzzing well :)

mvdan:
Thanks for the review! I need to do other bindnode changes for graphsync, so merging for now. More reviews always welcome even after the merge.

PR #366, opened 2022-02-22
node/bindnode: make AssignNode work at the repr level

### LINKED ISSUE #365: AssignNode() has trouble with renamed fields (via bindnode)
Reproduced with bindnode, error says its bindnode, but maybe it's TypedNode?

```go
type Foo struct{ I int }

func TestRenameAssignNode(t *testing.T) {
	ts, _ := ipld.LoadSchemaBytes([]byte(`
type Foo struct {
	I Int (rename "J")
}
`))
	FooProto := bindnode.Prototype((*Foo)(nil), ts.TypeByName("Foo"))

	// Decode straight into bindnode typed builder
	nb := FooProto.Representation().NewBuilder()
	if err := dagjson.Decode(nb, bytes.NewReader([]byte(`{"J":100}`))); err != nil {
		panic(err)
	}
	node := nb.Build()
	fmt.Printf("decode into: ")
	dagjson.Encode(node, os.Stdout)
	fmt.Println()

	// decode into basicnode builder
	nb = basicnode.Prototype.Any.NewBuilder()
	if err := dagjson.Decode(nb, bytes.NewReader([]byte(`{"J":100}`))); err != nil {
		panic(err)
	}
	node = nb.Build()

	// AssignNode from the basicnode form
	nb = FooProto.Representation().NewBuilder()
	if err := nb.AssignNode(node); err != nil {
		panic(err) // fails here
	}
	node = nb.Build()
	dagjson.Encode(node, os.Stdout)
	fmt.Println()
}
```

```
panic: bindnode TODO: invalid key: "J" is not a field in type Foo [recovered]
        panic: bindnode TODO: invalid key: "J" is not a field in type Foo
```

mvdan:
(see commit message)

Fixes #365.


rvagg (approved):
fix by removing gobs of code, the best sort 👌

rvagg:
Working on a fork makes it harder to `go get` a branch like this, unless there's a simple Go trick I'm missing?

Merging because I'd like to use this.

warpfork:
Posthoc review:

Wish: it looks like both the typelevel and the repr level are missing a fastpath branch for if they detect they're working on the same golang type and the same schema?  Mild bummer.  But I see the comment about trying to make that work.  And it's only a fastpath; not critical to correctness.

Critical sanitycheck: is it working on the correct level at the type level and the repr level: and it looks like yes, each is applying `datamodel.Copy` on a golang type that's working at the correct level, so all is good.  :heavy_check_mark: 

So, LGTM (even if belatedly) :) and indeed, yay for removing gobs of code :ok_hand:

Issue #367, opened 2022-02-23
AssignNode from a representation form of a TypedNode including a tuple panics with default values


rvagg:
This one's a little subtle and I'm not entirely clear on the bounds, but I believe it only relates to tuple representation, not map. But it could also relate to other forms where the typed form doesn't match the representation form.

If you take a node that's constructed using a TypedNode (I've only done this via bindnode so it could be bindnode specific) with unassigned values within a tuple representation struct, then use `AssignNode()` from the representation of that typenode, it fails, with "missing fields". But if I do it from a basic form (or just from a codec) with what should be the default values, then it's fine.

```go
func TestRenameAssignNode(t *testing.T) {
	type Foo struct {
		I string
		J string
		K int
	}
	ts, _ := ipld.LoadSchemaBytes([]byte(`
type Foo struct {
	I String
	J String
	K Int
} representation tuple
`))
	FooProto := bindnode.Prototype((*Foo)(nil), ts.TypeByName("Foo"))

	// decode an "empty" object into Foo, these are all default values
	nb := basicnode.Prototype.Any.NewBuilder()
	if err := dagjson.Decode(nb, bytes.NewReader([]byte(`["","",0]`))); err != nil {
		panic(err)
	}
	node := nb.Build()

	// AssignNode from the basicnode form
	nb = FooProto.Representation().NewBuilder()
	if err := nb.AssignNode(node); err != nil {
		panic(err)
	}
	node = nb.Build()
	fmt.Printf("assign into: ")
	dagjson.Encode(node, os.Stdout)
	fmt.Println()

	// make an "empty" form, although none of the fields are optional so we should end up with defaults
	nb = FooProto.Representation().NewBuilder()
	empty := nb.Build()
	fmt.Printf("empty build: ")
	dagjson.Encode(empty, os.Stdout) // prints the same as the previous
	fmt.Println()

	// AssignNode from the representation of the "empty" form, which should pass through default values
	nb = FooProto.Representation().NewBuilder()
	if err := nb.AssignNode(empty.(schema.TypedNode).Representation()); err != nil {
		panic(err) // fails here: panic: missing required fields: I,J,K
	}
	node = nb.Build()
	fmt.Printf("empty assign into: ")
	dagjson.Encode(node, os.Stdout)
	fmt.Println()
}
```

rvagg:
I believe https://github.com/ipld/go-ipld-prime/pull/368 fixes this

PR #368, opened 2022-02-23
fix(bindnode): tuple struct iterator should handle absent fields properly

rvagg:
Fixes: https://github.com/ipld/go-ipld-prime/issues/367

### DIFF
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -268,7 +268,9 @@ func (w *_nodeRepr) ListIterator() datamodel.ListIterator {
 	switch reprStrategy(w.schemaType).(type) {
 	case schema.StructRepresentation_Tuple:
 		typ := w.schemaType.(*schema.TypeStruct)
-		return &_tupleIteratorRepr{schemaType: typ, fields: typ.Fields(), val: w.val}
+		iter := _tupleIteratorRepr{schemaType: typ, fields: typ.Fields(), val: w.val}
+		iter.reprEnd = int(w.lengthMinusTrailingAbsents())
+		return &iter
 	default:
 		iter, _ := (*_node)(w).ListIterator().(*_listIterator)
 		if iter == nil {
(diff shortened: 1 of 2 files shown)


node/bindnode/repr.go
```diff
@@ -320,7 +322,7 @@ _skipAbsent:
 	if err != nil {
 		return 0, nil, err
 	}
-	if value.IsAbsent() {
+	if w.nextIndex > w.reprEnd {
```
rvagg, line 325: Is it possible to be in a situation where `value.IsAbsent()` and we're not at the end amongst trailing absents? e.g. `[absent,present,present,absent,absent]`? Is there a use of this iterator where that can happen, perhaps by constructing this from the typed form and then list iterating on the repr form?

rvagg:
btw, I could go either way on the tuple optionals, I've always thought that it was reasonable to allow trailing optionals but it does complicate matters somewhat for marginal gain and increased ambiguity. Maybe we should rip that support out (for now, or for good?)

node/bindnode/repr.go
mvdan: Absents can only be at the end, as per the undocumented spec :) See https://github.com/ipld/go-ipld-prime/blob/e39d20bf18dcc21a50ef2005bae6de11e5c311e0/schema/gen/go/genStructReprTuple.go#L13.

Arguably this needs to be properly documented in the specs, which is why I'll raise an issue elsewhere. But in short: we should support trailingn optionals in bindnode, to at least match codegen in terms of features.

mvdan (approved):

mvdan:
TL;DR fine to merge as is, looks great!

node/bindnode/repr.go
```diff
@@ -348,11 +350,7 @@ func (w *_nodeRepr) Length() int64 {
 	case schema.StructRepresentation_Map:
 		return w.lengthMinusAbsents()
 	case schema.StructRepresentation_Tuple:
-		// TODO: presumably, tuples should disallow optional fields.
-		// But the following tests disagrees:
-		// https://github.com/ipld/go-ipld-prime/blob/297518c911326ef7f44800dc32a923e7b084fccc/node/tests/schemaStructReprTuple.go#L24-L32
-		// return int64(len(w.schemaType.(*schema.TypeStruct).Fields()))
```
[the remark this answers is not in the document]
mvdan: Now https://github.com/ipld/go-ipld-prime/issues/369.

PR #371, opened 2022-03-01
node/bindnode: support pointers to datamodel.Node to bind with Any

mvdan:
(see commit message)


rvagg (approved):
👍 but this is working fine without any fixes, just needing to be careful about the pointer type (https://github.com/filecoin-project/go-data-transfer/pull/305)

Issue #319, opened 2021-12-27
need API for handling large bytes nodes


warpfork:
The `Node` interface needs to provide some way to handle "big" bytes -- bigger than would readily fit into a single `[]byte`.  Some sort of `io.Reader` implementation; and probably, an `io.ReadSeeker`.

This should make it possible to work even with data that wouldn't fit entirely in memory.  For example, we should be able to use this for ADLs that store huge ranges of bytes, across multiple blocks of actual storage, which may happen via something like the [FBL ADL](https://ipld.io/specs/advanced-data-layouts/fbl/).

A `GetBytesReadSeeker() (io.ReadSeeker, error)` method may be sufficient?  Are there any more operations than that which we expect to need?

The current `AsBytes() ([]byte, error)` method should also remain (to avoid breaking existing code), but newly written generic code would be encouraged to prefer to use `GetBytesReadSeeker` so that it works regardless of data size.

Correspondingly, `NodeAssembler` should be able to create a new node of bytes by copying in from an `io.Reader`, or returning an `io.WriteCloser`.

We may be able to provide both these features with feature detection rather than making them a required new part of `Node` and `NodeBuilder`, but if so, should make sure we provide good helper functions that provide the streaming style interfaces regardless of whether the optional features are present, so that this doesn't become something user code has to be bothered about.

Stretch goal: think about having variants of these features which state that byte slices used for intermediate handling may be retained by the implementation, so that it can attempt zero-copy handling of data.  (E.g. I can imagine that one might want to read some data off the filesystem; "chunk" it; and then it might be nice to pass off the chunks in a way that eventually ends up in a `putv([][]byte,...)`-style syscall.)

mvdan:
Another idea, to keep the API more consistent long-term, would be to deprecate the existing `AsBytes` and add a method that returns an `io.ReadSeeker`. Any implementations which can cheaply supply a `[]byte` could do so by returning an implementation which has a `Bytes() []byte` method, such as a `*bytes.Buffer`.

I don't think we can do this sanely without breaking downstreams, but I thought it would be worth mentioning as an option.

[...]

PR #372, opened 2022-03-01
Allow feature detection for `io.ReaderAt` on Bytes nodes

willscott:
We can handle large bytes by allowing Bytes kinded nodes to optionally support
the `ReadAt` method thereby implementing the ReaderAt interface.

This is proposed over a distinct the reader / seeker implementation because the ReaderAt interface is stateless - it does not need the addition of a position cursor to be tracked in the node between calls, while still allowing for partial reads of the node contents.

### DIFF
--- a/datamodel/node.go
+++ b/datamodel/node.go
@@ -168,15 +168,18 @@ type Node interface {
 }
 
 // LargeBytesNode is an optional interface extending a Bytes node that allows its
-// contents to be accessed through an io.Reader instead of a []byte slice. Use of
+// contents to be accessed through an io.ReadSeeker instead of a []byte slice. Use of
 // an io.Reader is encouraged, as it allows for streaming large byte slices
 // without allocating a large slice in memory.
 type LargeBytesNode interface {
 	Node
-	// AsLargeBytes returns an io.Reader that can be used to read the contents of the node.
+	// AsLargeBytes returns an io.ReadSeeker that can be used to read the contents of the node.
 	// Note that the presence of this method / interface does not imply that the node
 	// can always return a valid io.ReadSeeker, and the error value must also be checked
 	// for support.
+	// It is not guaranteed that all implementations will implement the full semantics of
+	// Seek, in particular, they may refuse to seek to the end of a large bytes node if
+	// it is not possible to do so efficiently.
 	AsLargeBytes() (io.ReadSeeker, error)
 }
 
--- a/datamodel/node.go
+++ b/datamodel/node.go
@@ -1,5 +1,7 @@
 package datamodel
 
+import "io"
+
 // Node represents a value in IPLD.  Any point in a tree of data is a node:
 // scalar values (like int64, string, etc) are nodes, and
 // so are recursive values (like map and list).
@@ -163,6 +165,19 @@ type Node interface {
 	//
 	// Calling this method should not cause an allocation.
 	Prototype() NodePrototype
+}
+
+// LargeBytesNode is an optional interface extending a Bytes node that allows its
+// contents to be accessed through an io.Reader instead of a []byte slice. Use of
+// an io.Reader is encouraged, as it allows for streaming large byte slices
+// without allocating a large slice in memory.
+type LargeBytesNode interface {
+	Node
+	// AsLargeBytes returns an io.Reader that can be used to read the contents of the node.
+	// Note that the presence of this method / interface does not imply that the node
+	// can always return a valid io.ReadSeeker, and the error value must also be checked
+	// for support.
+	AsLargeBytes() (io.ReadSeeker, error)
 }
 
 // NodePrototype describes a node implementation (all Node have a NodePrototype),
--- a/node/basicnode/bytes.go
+++ b/node/basicnode/bytes.go
@@ -73,6 +73,10 @@ func (plainBytes) AsLink() (datamodel.Link, error) {
 func (plainBytes) Prototype() datamodel.NodePrototype {
 	return Prototype__Bytes{}
 }
+func (pb plainBytes) ReadAt(p []byte, off int64) (n int, err error) {
+	n = copy(p, pb[off:])
+	return
+}
 
 // -- NodePrototype -->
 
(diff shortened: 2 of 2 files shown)


willscott:
Work i see in this repo for fully implementing this proposal if it gets general approval:

* [ ] Add comment about optional support for the interface at the main Node `AsBytes` method
* [ ] Add `ReadAt` to the bind node implementation
* [ ] Make use of the interface when available in codec Encoding

mvdan:
Before: https://github.com/ipld/go-ipld-prime/issues/319

I lean towards `ReadSeeker` over `ReaderAt` because `io.Reader` is fairly common, and because `ReaderAt` doesn't give any information about the total number of bytes - with seeking, you can seek all the way to the end to find out. Note that, currently, the `Length` method isn't valid for Bytes nodes.

On the flip side, `ReaderAt` is slightly easier to implement, and one could always implement an `io.Reader` atop an `io.ReaderAt`. I think I see your point about altering state when reading/seeking as a potential problem, but beyond "IPLD nodes are immutable", is that really a problem? It wouldn't affect what the node would contain or how it would be encoded; one could always seek back to the beginning.

For the sake of not keeping the read/seek offset state as part of the state itself, that's why in the other thread I suggested `AsLargeBytes() (io.ReadSeeker, error)`. The other advantage of this mechanism is that an implementation can support two concurrent readers.

There is one other thing to keep in mind about `ReaderAt`:

> Clients of ReadAt can execute parallel ReadAt calls on the same input source. 

This would mean that any implementation, including ADLs, would need to support concurrent calls.

mvdan:
I should also warn against having a node optionally implement an io interface directly, without a method to obtain that implementation instead. With `AsLargeBytes() (io.ReadSeeker, error)`, you can decide at run time whether or not you're really "large bytes" or not, whereas you must decide what the set of methods on a type is at compile time. This matters for `any` and kinded unions; `AsLargeBytes` will be defined for those, but will error if the kind isn't actually bytes. You could potentially emulate this by having `ReadAt` calls return a sentinel error, but then checking whether a node value or type implements `io.ReaderAt` is almost pointless, as it doesn't give you a good answer.

willscott:
* Having to calculate full size may also be expensive, it would be nice to be able to support offset reads without needing to sign up for that (maybe you don't support seek to the end, i suppose)
* The point on the clearly defined method makes sense. An `AsLargeBytes` seems reasonable.
* Do we not think supporting concurrent calls is desirable / causes undue burden versus read seeker?

mvdan:
> * (maybe you don't support seek to the end, i suppose)

Right, you could indeed error when asked to seek relative to the end.

> Do we not think supporting concurrent calls is desirable / causes undue burden versus read seeker?

With the understanding that we need an `AsT` method either way, the difference is that `ReaderAt` _requires_ concurrent calls to work, whereas `ReadSeeker` doesn't. Given that most users would presumably grab their own "reader" via the method, and not share it with other goroutines, then I imagine requiring concurrency would be unnecessary, and that requirement could make implementing a "big bytes ADL" that loads on-demand harder - one suddenly needs to worry about data races :)

All this said, it's still a "leaning towards `ReadSeeker`" for me. I think my only ask is that, if we do go for `ReaderAt`, then we should leave space for implementers to provide the total length, such as by extending the existing `Length` method.

willscott:
I guess the main candidate here at the moment is the unixfsnode ADL. I think i see what a read-seeker would look like there, so i'll prototype that library as a candidate implementation next.

warpfork:
I have read this, and believe you two have more detailed understanding than I've got loaded anywhere near to my "hot cache", and I trust you :)

This sounds broadly directionally correct to me.  Between the two specific interfaces I don't have a strong opinion.

FWIW, having one alloc to return whatever-interface-it-is sounds probably acceptable to me.  Iterators already have that same cost threshhold in most implementations.  If we wanted to optimize heavily: for both iterators, and for this, we could embed one of the thing in the main structure, wrap it in some sync primitives, and do allocs iff more than one is requested -- at the cost of larger resident memory.  That hasn't seemed necessary yet, but seems plausible as future work.

mvdan:
That sounds reasonable, but I also think that one alloc to be able to consume a "big bytes" node will likely be negligible. At that point, you're going to be worrying about bigger factors like copying the bytes or reusing buffers correctly.

willscott:
it's going to be hard to add `AsLargeBytes` to the core ipld.node interface because it'll gate updates of ipld-prime to a release where all derived gen'd code has been re-built to include the additional method, right?

mvdan:
We could declare it as an optional interface either for some time, or forever. I think either of those will work as long as "imlements the interface" is not understood as "can be used as large bytes". But in any case, we can start with declaring the extra interface for now, and later decide if enough implementations have transitioned to it that we can add the method straight to Node. For example:

```
type LargeBytesNode interface {
    Node
    AsLargeBytes() (io.ReadSeeker, error)
}
```

willscott:
the downside is that there are two checks on each access - first a check of whether the interface is implemented, and then the error check when attempting to get the `io.Reader`.
That said, this seems simplest for an incremental rollout

mvdan:
Indeed, so I'm inclined to believe that we could move the method to `ipld.Node` sometime down the line - perhaps in six months or so.

willscott:
@mvdan  ready for review. Should we also think about a way to make a bytes node from an io.Reader as an extension to the builder side API?

I'm inclined to defer for now, but it seems useful to figure out for efficiency of copies at some point.

mvdan:
I am happy to defer on that too, for now. I think building "large bytes" nodes will fall into two categories:

1) The user doesn't care about scalability or efficiency. They can use `SetBytes`.
2) The user does care about scalability, so they will use some ADL depending on their needs, which will have its own efficient and tailored "build" API.

mvdan (approved):
nit: the docs talk about an io.Reader when it's actually ReadSeeker

mvdan:
We could also document that implementations may refuse to support certain kinds of seeks, like SeekEnd.

PR #375, opened 2022-03-03
Add partial-match traversal of large bytes

willscott:
* add a `streamBytes` parallel to `plainBytes` for byte nodes backed by `io.ReadSeekers` rather than `[]byte`
* Implement `Slice` for subset matches
* Extend the `Decide(node) bool` method for matching to `Match(node) (node, error)`, which allows a match to be done on only a subset of a node.
* Update `walk_adv` to make use of the `Slice` selector by way of the `Match` method.


traversal/selector/matcher.go
```diff
@@ -16,7 +18,61 @@ import (
...
+// The returned node will be limited based on slicing the specified range of the
+// node into a new node, or making use of the `AsLargeBytes` io.ReadSeeker to
+// restrict response with a SectionReader.
+type Slice struct {
+	From int64
+	To   int64
+}
+
+func (s Slice) Slice(n datamodel.Node) (datamodel.Node, error) {
+	if n.Kind() == datamodel.Kind_String {
```
mvdan, line 35: use a switch?

traversal/selector/matcher.go
```diff
@@ -16,7 +18,61 @@ import (
...
+		}
+		return basicnode.NewString(str[from:to]), nil
+	} else if n.Kind() == datamodel.Kind_Bytes {
+		if lbn, ok := n.(datamodel.LargeBytesNode); ok {
+			rdr, err := lbn.AsLargeBytes()
+			if err != nil {
+				return nil, err
+			}
+
+			sr := io.NewSectionReader(readerat{rdr}, int64(s.From), int64(s.To))
```
mvdan, line 56: I think these int64 conversions are redundant

traversal/selector/matcher.go
```diff
@@ -16,7 +18,61 @@ import (
...
+			return basicnode.NewBytesFromReader(sr), nil
+		}
+		bytes, err := n.AsBytes()
+		if err != nil {
+			return nil, err
+		}
+		to := s.To
```
mvdan, line 63: might as well declare from/to at the top of the func and share them everywhere

traversal/selector/matcher_util.go
```diff
@@ -0,0 +1,15 @@
+package selector
+
+import "io"
+
+type readerat struct {
+	io.ReadSeeker
+}
+
+func (r readerat) ReadAt(p []byte, off int64) (n int, err error) {
+	_, err = r.Seek(off, 0)
```
mvdan, line 10: nit: as an optimization, if you're already at the right offset, you could skip the seek. fine to leave for later, but I'd still keep a TODO, because this will matter for e.g. SectionReader - one vs two calls per read.

traversal/selector/matcher_util.go
```diff
@@ -0,0 +1,15 @@
+package selector
+
+import "io"
+
+type readerat struct {
+	io.ReadSeeker
```
mvdan, line 6: note that ReaderAt is safe for concurrent use and ReadSeeker isn't, so you should probably grab a sync.Mutex or some other form of semaphore on every ReadAt call

mvdan (approved):
Code SGTM with the review, so I think this is fine to merge as long as the spec changes are also approved and merged

traversal/selector/matcher_util.go
willscott: leaving as todo, because i think there's a tradeoff between either calling seek to get current offset each time, or keeping extra state to track.

traversal/selector/matcher_util.go
willscott: Since i only use this within the readseeker that is non-concurrent again, is it okay to note that it isn't actually concurrent safe?

traversal/selector/matcher_util.go
mvdan: Keeping extra state seems easy enough, given that you need a mutex anyway :) but fine to leave as a TODO for sure

traversal/selector/matcher_util.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// ReadAt provides the io.ReadAt method over a ReadSeeker.
+// This implementation does not support concurrent calls to `ReadAt`,
```
hannahhoward, line 10: So my one worry here is: what are the odds the consumer of this code reads this comment, and not the comment on golang's ReaderAt interface -- is there a chance for misuse through the chain of readerAt -> SectionReader -> NewBytesFromReader? SectionReader certainly seems to use ReadAt to implement seek+read for example. 

A seperate question: should AsLargeBytes return a ReadSeeker that is also a ReaderAt? os.File satisfies both. new section reader satisfies both.

Non blocking, but we should look at figuring this out when we can.

traversal/selector/selector.go
```diff
@@ -106,6 +106,12 @@ type Selector interface {
 	// Only "Matcher" clauses actually implement this in a way that ever returns "true".
 	// See the Selector specs for discussion on "matched" vs "reached"/"visited" nodes.
 	Decide(node datamodel.Node) bool
```
hannahhoward, line 108: Are we using decide any more? Should we just delete it? No one uses compile selectors outside of go-ipld-prime I think

hannahhoward (approved):
non-blocking comments but otherwise LGTM

hannahhoward:
Also I rebased this for you since I was the one who created your merge conflict :)

traversal/selector/selector.go
willscott: i don't see any uses of `Decide` in the go-ipld-prime directly before this PR either. I think the thought is to transition to Match as the signature that we actually need for these walks.

My preference would be to mark this as deprecated rather than turn this into a backwards-incompatible change.

traversal/selector/matcher_util.go
willscott: @mvdan argued that `ReadSeeker` is a more common and lower effort to ask implementations of node to implement. This wrapper is internal (since the struct type is not exported) and so is only for use by the Slice matcher to wrap a `SectionReader` around a node's `ReadSeeker`. As such, I believe there's minimal potential for misuse.

warpfork:
Crosslinking (since I just had to go re-find it): spec happened in: https://github.com/ipld/ipld/pull/184

PR #376, opened 2022-03-04
Implement WalkTransforming traversal

willscott:
* there's probably a better algorithm for the skip/don't skip subset of interests that are to be recursively traversed versus what's set up here.
* there might be a way to combine the list and map recursion methods
* should think about a reasonable set of test cases / fixtures


warpfork (approved):
+1 from me.  (Not a detailed review; but it looks about right.)

I'm... probably _not_ going to harp on test fixtures for this too much, for once (surprising though that might be).  I think this function may be a golang-specific cul-de-sac a bit, and therefore I care a lot less.

willscott:
I considered if this needs to add the partial skip logic added in #358. I believe in the same way that it is not used for `WalkLocal` it isn't directly needed for `WalkTransforming` (or can be added later if it becomes needed)

PR #383, opened 2022-03-08
node/bindnode: start fuzzing with schema/dmt and codec/dagcbor

mvdan:
(see commit message)

Updates #357.

### DIFF
--- a/node/bindnode/fuzz_test.go
+++ b/node/bindnode/fuzz_test.go
@@ -1,0 +1,238 @@
+//go:build go1.18
+// +build go1.18
+
+package bindnode_test
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/ipld/go-ipld-prime/codec/dagcbor"
+	"github.com/ipld/go-ipld-prime/codec/dagjson"
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+	"github.com/ipld/go-ipld-prime/schema"
+	schemadmt "github.com/ipld/go-ipld-prime/schema/dmt"
+	schemadsl "github.com/ipld/go-ipld-prime/schema/dsl"
+)
+
+var fuzzInputs = []struct {
+	schemaDSL, nodeDagJSON string
+}{
+	{
+		schemaDSL:   `type Root bool`,
+		nodeDagJSON: `true`,
+	},
+	{
+		schemaDSL:   `type Root int`,
+		nodeDagJSON: `123`,
+	},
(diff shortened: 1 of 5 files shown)


mvdan:
This is the first batch of fuzzing work. Plenty of crashers remain, but I'm sending this for now to get reviews on the fuzz code itself.

A sample of how it fails:

```
--- FAIL: FuzzBindnodeViaDagCBOR (0.00s)
    --- FAIL: FuzzBindnodeViaDagCBOR/821c248529299bb6b68e443e4b00cc11c6605f766f25e619d85e6d6b40a33dac (0.00s)
        fuzz_test.go:148: node in dag-cbor: "a000"
        fuzz_test.go:149: schema in dag-json: {"types":{"Root":{"enum":{"members":["000","000"],"representation":{"int":{"000":-17,"000":-17}}}}}}
        fuzz_test.go:158: node in dag-json: "0"
        fuzz_test.go:198: decode and encode roundtrip with dag-cbor repr=false
        fuzz_test.go:217: decode successful: "0"
        fuzz_test.go:224: re-encode successful: "a0"
        fuzz_test.go:198: decode and encode roundtrip with dag-cbor repr=true
        fuzz_test.go:217: decode successful: "0"
        fuzz_test.go:218: AsInt: "0" is not a valid member of enum Root
FAIL
```

If you look at the logs carefully, you can see that we successfully decoded the string `"0"` at the repr level for an int-repr enum, which should never happen. That was the second bug we've fixed.

mvdan:
I've also added initial fuzz inputs driven by the IPLD Schema features used in graphsync: keyed unions, enums, structs, etc. I've omitted the others for now.

node/bindnode/fuzz_test.go
```diff
@@ -0,0 +1,238 @@
...
+				node = node.(schema.TypedNode).Representation()
+			}
+			// Unwrap returns a pointer, and %#v prints pointers as hex,
+			// so to get useful output, use reflect to dereference them.
+			t.Logf("decode successful: %#v", reflect.ValueOf(bindnode.Unwrap(node)).Elem().Interface())
+			reenc := marshalDagCBOR(t, node)
+			if !bytes.Equal(reenc, nodeDagCBOR) {
+				// TODO: reenable this once the dagcbor codec rejects non-canonical
+				// inputs like "00" and "a000", "0" and "a0" respectively.
+				// t.Errorf("node reencoded as %q rather than %q", reenc, nodeDagCBOR)
```
mvdan, line 233: I'm not sure if this matters for graphsync. The dagcbor codec will, as of today, happily accept non-canonical inputs like `a000`; note that it's the string `"0"`, but with two trailing bytes `"00"`. If it were a three-byte string, you'd have `c000`:

https://cbor.me/?diag=%220%22
https://cbor.me/?diag=%22000%22

warpfork (approved):

PR #385, opened 2022-03-09
many more fuzz fixes

mvdan:

### DIFF
--- a/node/bindnode/infer.go
+++ b/node/bindnode/infer.go
@@ -206,7 +206,30 @@ func verifyCompatibility(seen map[seenEntry]bool, goType reflect.Type, schemaTyp
 	}
 }
 
-func inferGoType(typ schema.Type) reflect.Type {
+// If we recurse past a large number of levels, we're mostly stuck in a loop.
+// Prevent burning CPU or causing OOM crashes.
+// If a user really wrote an IPLD schema or Go type with such deep nesting,
+// it's likely they are trying to abuse the system as well.
+const maxRecursionLevel = 1 << 10
+
+type inferredStatus int
+
+const (
+	_ inferredStatus = iota
+	inferringInProcess
+	inferringDone
+)
+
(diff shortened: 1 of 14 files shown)


mvdan:
I briefly thought about adding the schema repr check in the correspoinding `schema.Spawn` APIs, but those have no room to return errors, and it's fairly likely we'll start hiding those APIs for internal use only once the schemadmt and schemadsl packages are mature enough. So I think doing the checking at dmt.Compile is good enough.

mvdan:
I've spent another couple of hours fuzzing, and found more crashers with fixes. Note that the majority found so far related to weird schema edge cases, so I don't think they will affect graphsync. One that certainly could is the second commit, where empty union values were accepted when decoding via bindnode.

cc @willscott

mvdan:
Also worth noting that the fuzzer isn't particularly complete in terms of finding bindnode bugs; see the multiple TODOs I've left around fuzz_test.go.

warpfork (approved):
Spectacular!  Yayy fuzzing

node/bindnode/infer.go
```diff
@@ -206,7 +206,30 @@ func verifyCompatibility(seen map[seenEntry]bool, goType reflect.Type, schemaTyp
 	}
 }
 
-func inferGoType(typ schema.Type) reflect.Type {
+// If we recurse past a large number of levels, we're mostly stuck in a loop.
+// Prevent burning CPU or causing OOM crashes.
+// If a user really wrote an IPLD schema or Go type with such deep nesting,
+// it's likely they are trying to abuse the system as well.
+const maxRecursionLevel = 1 << 10
```
rvagg, line 213: 1024 is still pretty large; find if this isn't a significant cost but you could also lower this and wait till someone squeals and asks for it to be increased

mvdan: Maybe - I admit I chose it arbitrarily. It's high enough that I don't expect anyone will actually run into it, but low enough that it should never cause an OOM on its own.

Issue #374, opened 2022-03-03
dagjson.Decode support for self-delimited reading


petar:
The current implementation of dagjson.Decode consumes whitespaces and EOF after a valid object.
This precludes repeated use on a single stream. Perhaps we could introduce a decoder option that does not consume any bytes after the terminal "}". This is generally the expected design for a decoder of any self-delimited format, in the sense that the self-delimitedness property is otherwise unusable in code. (E.g. json.Decode.)

BigLep:
2022-05-03 triage conversation: this isn't something we're going to build into the core.  This would need to be a special mode.  Contributions welcome.

PR #386, opened 2022-03-15
fix(dagcbor): don't accept trailing bytes

rvagg:
Just re-using an approach that's used by the dag-json codec rather than trying to hack refmt which won't properly support this kind of inspection since it only considers complete terminal sets rather than caring about where the bytes stop.

This does mean https://github.com/ipld/go-ipld-prime/issues/374 will apply to dag-cbor too, but that's appropriate - we should be erroring here and allowing the user to opt-in (somehow!) to allowing it to go only up to the end of the terminal set and no more.

---

To try out the fuzz thing I grabbed go 1.18 and ran `go test -fuzz=FuzzBindnodeViaDagCBOR` in `node/bindnode` to make it run and the `a000` case is fixed there, but it still fails:

```
fuzz: elapsed: 0s, gathering baseline coverage: 0/25 completed
fuzz: elapsed: 0s, gathering baseline coverage: 25/25 completed, now fuzzing with 16 workers
fuzz: minimizing 112-byte failing input file
fuzz: elapsed: 0s, minimizing
--- FAIL: FuzzBindnodeViaDagCBOR (0.37s)
    --- FAIL: FuzzBindnodeViaDagCBOR (0.00s)
        fuzz_test.go:165: schema in dag-cbor: a1657479706573a164526f6f74a1636d6170a2676b65795479706566537472696e676976616c75655479706563496e74
        fuzz_test.go:166: node in dag-cbor: a2615830613030
        fuzz_test.go:167: schema in dag-json: {"types":{"Root":{"map":{"keyType":"String","valueType":"Int"}}}}
        fuzz_test.go:176: node in dag-json: {"0":-17,"X":-17}
        fuzz_test.go:219: decode and encode roundtrip with dag-cbor repr=false
        fuzz_test.go:238: decode successful: struct { Keys []string; Values map[string]int }{Keys:[]string{"X", "0"}, Values:map[string]int{"0":-17, "X":-17}}
        fuzz_test.go:241: node reencoded as "\xa2a00aX0" rather than "\xa2aX0a00"
        fuzz_test.go:243: re-encode successful: "\xa2a00aX0"
        fuzz_test.go:219: decode and encode roundtrip with dag-cbor repr=true
        fuzz_test.go:238: decode successful: struct { Keys []string; Values map[string]int }{Keys:[]string{"X", "0"}, Values:map[string]int{"0":-17, "X":-17}}
        fuzz_test.go:241: node reencoded as "\xa2a00aX0" rather than "\xa2aX0a00"
        fuzz_test.go:243: re-encode successful: "\xa2a00aX0"
    
    Failing input written to testdata/fuzz/FuzzBindnodeViaDagCBOR/11124a93392709940afb28dd60fecc9841776627a436091b4b9d09da57a97905
    To re-run:
    go test -run=FuzzBindnodeViaDagCBOR/11124a93392709940afb28dd60fecc9841776627a436091b4b9d09da57a97905
FAIL
exit status 1
FAIL    github.com/ipld/go-ipld-prime/node/bindnode     0.557s
```

is that expected?

### DIFF
--- a/codec/dagcbor/unmarshal.go
+++ b/codec/dagcbor/unmarshal.go
@@ -59,14 +59,15 @@ func (cfg DecodeOptions) Decode(na datamodel.NodeAssembler, r io.Reader) error {
 	}
 
 	var buf [1]byte
-	count, err := r.Read(buf[:])
-	if count != 0 {
-		return ErrTrailingBytes
-	}
-	if err == io.EOF {
+	_, err = io.ReadFull(r, buf[:])
+	switch err {
+	case io.EOF:
 		return nil
+	case nil:
+		return ErrTrailingBytes
+	default:
+		return err
 	}
-	return err
 }
 
 // Future work: we would like to remove the Unmarshal function,
--- a/codec/dagcbor/unmarshal.go
+++ b/codec/dagcbor/unmarshal.go
@@ -18,6 +18,7 @@ import (
 var (
 	ErrInvalidMultibase         = errors.New("invalid multibase on IPLD link")
 	ErrAllocationBudgetExceeded = errors.New("message structure demanded too many resources to process")
+	ErrTrailingBytes            = errors.New("unexpected content after end of cbor object")
 )
 
 const (
@@ -49,9 +50,23 @@ func (cfg DecodeOptions) Decode(na datamodel.NodeAssembler, r io.Reader) error {
 		return na2.DecodeDagCbor(r)
 	}
 	// Okay, generic builder path.
-	return Unmarshal(na, cbor.NewDecoder(cbor.DecodeOptions{
+	err := Unmarshal(na, cbor.NewDecoder(cbor.DecodeOptions{
 		CoerceUndefToNull: true,
 	}, r), cfg)
+
+	if err != nil {
+		return err
+	}
+
+	var buf [1]byte
+	count, err := r.Read(buf[:])
+	if count != 0 {
+		return ErrTrailingBytes
+	}
+	if err == io.EOF {
+		return nil
+	}
+	return err
 }
 
 // Future work: we would like to remove the Unmarshal function,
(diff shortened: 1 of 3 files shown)


codec/dagcbor/unmarshal.go
```diff
@@ -49,9 +50,23 @@ func (cfg DecodeOptions) Decode(na datamodel.NodeAssembler, r io.Reader) error {
...
 		CoerceUndefToNull: true,
 	}, r), cfg)
+
+	if err != nil {
+		return err
+	}
+
+	var buf [1]byte
+	count, err := r.Read(buf[:])
+	if count != 0 {
```
mvdan, line 63: From the io.Reader docs:

> Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) == 0. Callers should treat a return of 0 and nil as indicating that nothing happened; in particular it does not indicate EOF. 

So you could end up with Read calls returning the following:

```
0, nil
0, nil
1, nil
0, io.EOF
```

If you want to properly skip those "noop" reads, use `io.ReadFull`:

> The error is EOF only if no bytes were read. If an EOF happens after reading some but not all the bytes, ReadFull returns ErrUnexpectedEOF. On return, n == len(buf) if and only if err == nil.

mvdan (approved):
Thank you for this! I tried calling Step on the tokenizer, which panicked, but it didn't occur to me to use Read directly. I assume this works because refmt doesn't read more than it needs to, e.g. due to buffered reads.

I think the next crasher you've found is not about trailing bytes, but about dagcbor accepting unmarshals where some of the data is out of order (presumably map elements). I think we can deal with that in a separate PR.

codec/dagcbor/unmarshal.go
rvagg: k, can you check the latest commit and comment on that approach?

mvdan (approved):
SGTM!

codec/dagcbor/unmarshal.go
mvdan: Done, approved the entire thing. Will let you merge as you please, e.g. squashing.

rvagg:
doh! I rebase-merged this, I really wish GitHub had proper native `fixup!` support for this.

@mvdan @warpfork I hope you don't mind if I force push to master while it's on top?

mvdan:
Go ahead!

PR #388, opened 2022-03-17
node/bindnode: minor fuzz improvements

mvdan:
(see commit message)

### DIFF
--- a/node/bindnode/fuzz_test.go
+++ b/node/bindnode/fuzz_test.go
@@ -85,7 +85,7 @@ var fuzzInputs = []struct {
 			| One ("1")
 			| Two ("2")
 		} representation int`,
-		nodeDagJSON: `"Two"`,
+		nodeDagJSON: `2`,
 	},
 	{
 		schemaDSL: `type Root union {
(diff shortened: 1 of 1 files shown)


node/bindnode/fuzz_test.go
```diff
@@ -132,6 +145,7 @@ func marshalDagJSON(tb testing.TB, node datamodel.Node) []byte {
 
 func FuzzBindnodeViaDagCBOR(f *testing.F) {
 	for _, input := range fuzzInputs {
+		// f.Logf("debug: %#v\n", input)
```
willscott, line 148: remove fully?

willscott (approved):

node/bindnode/fuzz_test.go
mvdan: It's been useful to me twice already, so I thought it was harmless to leave behind. I don't feel strongly either way, as this is test code anyway. But keeping it uncommented always is unfortunate, because fuzzing begins by spamming a bunch of generally unhelpful stuff.

willscott: I don't think we have a strong convention here. probably fine to leave as is.

warpfork: season to taste indeed, but I certainly don't object to leaving printfs like this behind in comments.

It's not always obvious what a useful printf incantation is going to be, so hints are cool imo.

PR #391, opened 2022-03-22
datamodel: document that repr prototypes produce type nodes

### LINKED ISSUE #362: better document that representation-level prototypes build type-level nodes
Let's say I have a node `n` which satisfies schema.TypedNode. How do I encode this to CBOR and decode it back to the same kind of schema.TypedNode

Let's say I do:

```golang
buf := new(bytes.Buffer)
dagcbor.Encode(n, buf)
```

I believe this will not encode `n` as expected -- it will encode according to the in memory node interface presented by `n`, rather than it's representation, which according to the spec describes how it maps to the IPLD data model and hence how its serialized.

Ok so instead I do:

```golang
buf := new(bytes.Buffer)
dagcbor.Encode(n.Representation(), buf)
```

I believe I now have the right data serialized.

But how to get get it back out and into another instance the same kind of node as n. I can get the node prototype by doing: 

```golang
np := n.Prototype()
```

But if I do:

```golang
nb := np.NewBuilder()
dagcbor.Decode(nb, buf)
```

Well this doesn't work as np.NewBuilder returns a builder for the in memory representation presumably, which also presumably will fail to deserialize should I use say `representation tuple` on a struct.

Ok, so I can get the representation prototype and decode with it as so:

```golang
tnp := np.(schema.TypedPrototype)
rp := tnp.Representation()
rnb := rp.NewBuilder()
dagcbor.Decode(rnb, buf)
```

But if I do:

```golang
n2 := rnb.Build()
```

I get a representation node and now I am stuck, cause I have no method to get back from the representation to schema.TypedNode

I wonder if need to make the following changes to the schema types:

```golang
type TypedNode interface {
	// schema.TypedNode acts just like a regular Node for almost all purposes;
	// which datamodel.Kind it acts as is determined by the TypeKind.
	// (Note that the representation strategy of the type does *not* affect
	// the Kind of schema.TypedNode -- rather, the representation strategy
	// affects the `.Representation().Kind()`.)
	//
	// For example: if the `.Type().TypeKind()` of this node is "struct",
	// it will act like Kind() == "map"
	// (even if Type().(Struct).ReprStrategy() is "tuple").
	datamodel.Node

	// Type returns a reference to the reified schema.Type value.
	Type() Type

	// Representation returns a datamodel.Node which sees the data in this node
	// in its representation form.
	//
	// For example: if the `.Type().TypeKind()` of this node is "struct",
	// `.Representation().TypeKind()` may vary based on its representation strategy:
	// if the representation strategy is "map", then it will be Kind=="map";
	// if the streatgy is "tuple", then it will be Kind=="list".
	Representation() RepresentationNode
}

type RepresentationNode interface {
      datamodel.Node
      
      // Typed returns a datamodel.Node which sees the data in this node
      // in its typed schema form
      Typed() TypedNode
}
// ....

// TypedPrototype is a superset of the datamodel.Nodeprototype interface, and has
// additional behaviors, much like TypedNode for datamodel.Node.
type TypedPrototype interface {
	datamodel.NodePrototype

	// Type returns a reference to the reified schema.Type value.
	Type() Type

	// Representation returns a datamodel.NodePrototype for the representation
	// form of the prototype.
	Representation() RepresentationPrototype
}

type RepresentationPrototype interface {
      datamodel.NodePrototype
      
      	// Representation returns a datamodel.NodePrototype for the typed
	// form of the prototype.
      Typed() TypedPrototype
```

mvdan:
(see commit message)

Fixes #362.

### DIFF
--- a/datamodel/nodeBuilder.go
+++ b/datamodel/nodeBuilder.go
@@ -149,6 +149,14 @@ type NodeBuilder interface {
 	// all responsibility for validating the assembled data and returning
 	// any errors from that process.
 	// (Correspondingly, there is no error return from this method.)
+	//
+	// Note that building via a representation-level NodePrototype or NodeBuilder
+	// returns a node at the type level which implements schema.TypedNode.
+	// To obtain the representation-level node, you can do:
+	//
+	//    // builder is at the representation level, so it returns typed nodes
+	//    node := builder.Build().(schema.TypedNode)
+	//    reprNode := node.Representation()
 	Build() Node
 
 	// Resets the builder.  It can hereafter be used again.


willscott (approved):
this edge does continue to be a source of pain

mvdan:
You mean the fact that it wasn't documented, or that you wish the code worked differently? :) The issue does talk about other options, like changing the behavior, but I do worry that that would break the interface - which is probably the worst kind of change we can release.

willscott:
I meant more generally the somewhat easy-to-forget need to call `.Representation()` in a variety of places. It's caught us up before at runtime, and probably will again.

mvdan:
Ah yes. I'd still like for the codec entrypoint APIs to call `.Representation()` on typed nodes by default, and I don't think that would break any code in practice, given that basically noone should _need_ to encode at the type level for production software.

PR #398, opened 2022-04-04
node/bindnode: add some more docs

mvdan:
(see commit message)


masih (approved):

rvagg (approved):
extra :thumbsup: because we got a comprehensive video explaining these too

PR #401, opened 2022-04-10
node/bindnode: allow nilable types for IPLD optional/nullable

### LINKED ISSUE #378: node/bindnode: allow nilable types instead of a pointer for optional/nullable schemas
For example, right now, to bind an IPLD `struct { Foo optional Any }`, one must use a pointer for the field type, like `struct { Foo *datamodel.Node }`.

However, in this case and some others, the Go type `datamodel.Node` already has a `nil` state to signal "absent", so we could reuse that instead of adding our own via a pointer. Not only is that type easier for the user to read and write, but it's also less prone to bugs, because a pointer to an interface has two nil states: nil pointer, and non-nil pointer to a nil interface.

This would also apply to IPLD schema's `nullable`, which similarly requires a pointer now.

This would have helped with https://github.com/ipld/go-ipld-prime/pull/371, and also with the new https://github.com/ipld/go-ipld-prime/issues/377.

What to do with `optional nullable` fields remains a bit of a TODO, because right now those use two chained pointers, which is a bit weird. We might want to rethink that with generics at some point, to end up with a wrapper like `OptionalNullable` that keeps two boolean fields, akin to https://github.com/ipld/go-ipld-prime/issues/340.

mvdan:
(see commit message)

Fixes #378.

### DIFF
--- a/node/bindnode/infer.go
+++ b/node/bindnode/infer.go
@@ -108,10 +108,11 @@ func verifyCompatibility(seen map[seenEntry]bool, goType reflect.Type, schemaTyp
 		}
 		goType = goType.Elem()
 		if schemaType.ValueIsNullable() {
-			if goType.Kind() != reflect.Ptr {
-				doPanic("nullable types must be pointers")
+			if ptr, nilable := ptrOrNilable(goType.Kind()); !nilable {
+				doPanic("nullable types must be nilable")
+			} else if ptr {
+				goType = goType.Elem()
 			}
-			goType = goType.Elem()
 		}
 		verifyCompatibility(seen, goType, schemaType.ValueType())
 	case *schema.TypeMap:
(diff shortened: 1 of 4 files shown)


mvdan:
FYI @rvagg @BigLep 

@masih @hannahhoward (and Rod) please test this in projects like graphsync and storetheindex, since I believe those use optional/nullable with bindnode and had run into the limitation with pointers. This should work, but it's best to double check there are no known regressions before it's merged.

node/bindnode/infer_test.go
```diff
@@ -377,33 +404,48 @@ func TestPrototypePointerCombinations(t *testing.T) {
...
+					}
+				default:
 					qt.Assert(t, err, qt.Not(qt.IsNil))
 				}
 			})
 		}
 	}
 }
 
+func nilable(kind reflect.Kind) bool {
```
willscott, line 440: do you need this, or can you look at the second return of `ptrOrNilable`?

willscott (approved):

node/bindnode/infer_test.go
mvdan: The funcs are unexported so they cannot be shared between the test and non-test packages - that's how I arrived at the smaller copy. Any other alternative is more lines of code AFAICT.

node/bindnode/infer.go
```diff
@@ -186,10 +201,11 @@ func verifyCompatibility(seen map[seenEntry]bool, goType reflect.Type, schemaTyp
 
 		for i, schemaType := range schemaMembers {
 			goType := goType.Field(i).Type
-			if goType.Kind() != reflect.Ptr {
-				doPanic("union members must be pointers")
+			if ptr, nilable := ptrOrNilable(goType.Kind()); !nilable {
+				doPanic("union members must be nilable")
```
rvagg, line 205: I don't think this is the ideal error here - it's talking schema language about a Go struct, maybe that's the best we have? but istm that this would be better stated as "must be pointers or interfaces".

Not blocking, just noting that it's weird encountering this error in the test when it's refering to the Go type.

rvagg (approved):
awesome
you're already missed @mvdan, such nice code to review

PR #404, opened 2022-04-26
schema: enable inline types through dsl parser & compiler

rvagg:
PR to ipld/ipld coming up after this for fixtures

This lets the schema parser parse things like:

```
  type LinkMap {String:&Boom}
  type MapMap {String:{String:Boom}}
  type ListMap {String:[Boom]}
```

enabled for links, maps, structs, keyed and kinded unions, for inline maps, lists and links (which have their own awkwardness)

### DIFF
--- a/schema/dmt/compile.go
+++ b/schema/dmt/compile.go
@@ -78,9 +78,18 @@ func todoAnonTypeName(nameOrDefn TypeNameOrInlineDefn) string {
 	case defn.TypeDefnList != nil:
 		defn := defn.TypeDefnList
 		return fmt.Sprintf("List__%s", todoAnonTypeName(defn.ValueType))
+	case defn.TypeDefnLink != nil:
+		return todoAnonLinkName(*defn.TypeDefnLink)
 	default:
 		panic(fmt.Errorf("%#v", defn))
 	}
+}
+
+func todoAnonLinkName(defn TypeDefnLink) string {
+	if defn.ExpectedType != nil {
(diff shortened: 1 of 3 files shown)


rvagg:
https://github.com/ipld/ipld/pull/211

schema/dmt/compile.go
```diff
@@ -78,11 +78,20 @@ func todoAnonTypeName(nameOrDefn TypeNameOrInlineDefn) string {
...
 		defn := defn.TypeDefnList
 		return fmt.Sprintf("List__%s", todoAnonTypeName(defn.ValueType))
+	case defn.TypeDefnLink != nil:
+		return todoAnonLinkName(*defn.TypeDefnLink)
 	default:
 		panic(fmt.Errorf("%#v", defn))
 	}
 }
 
+func todoAnonLinkName(defn TypeDefnLink) string {
```
willscott, line 88: does the `todo` prefix here mean this isn't done?

willscott (approved):

schema/dmt/compile.go
rvagg: I don't actually know .. there's a few of these in compile.go now and it's not entirely clear why the "todo" is in their names.. Do you think I should remove it from this one and `todoAnonTypeName()`? By adding links we should be making it complete, so maybe it's as simple as that.

willscott: it seems reasonable to remove the todo's at this point. they seem misleading

rvagg: yeah, done

PR #409, opened 2022-05-02
unified-ci

rvagg:
Manually add unified-ci configs; **and** adjust to make it work here by amending the `go test` command to exclude some tests that involve plugins which aren't compatible with `coverpkg` which breaks plugin usage.

Opening this for discussion because it needs upstream changes in https://github.com/protocol/.github

Ref: https://github.com/golang/go/issues/23410
Ref: https://github.com/ipld/go-ipld-prime/blob/master/schema/gen/go/HACKME_testing.md


rvagg:
This also adds a .github/workflows/go-test-prime.yml workflow to do a _normal_ `go test` without coverage and just on Linux, so we can invoke the behavioural tests from here--this means a minor double-up on testing but Linux testing has some shortcuts here and is pretty quick and not having the behavioural tests in CI would be very unfortunate.

Issue #352, opened 2022-02-04
node/bindnode: provide a mode that's compatible with cbor-gen


mvdan:
https://github.com/whyrusleeping/cbor-gen is used on multiple projects, and it works directly with Go types and the dag-cbor encoding.

bindnode, when combined with the dagcbor package, is actually pretty close to how cbor-gen works - if we squint a little to swap code generation for reflection.

We should support some way to use those two packages together as a drop-in replacement to cbor-gen, ideally without having to change the Go types at all. This way, we could encourage users of cbor-gen to switch to the new libraries and automatically get the rest of the advantages of go-ipld-prime and its node interface - being able to use selector traversals, being able to choose other codecs, being able to use IPLD schemas with different representation strategies, etc.

We should see whether we can support that compatibility with cbor-gen without adding an opt-in flag to either package. My guess is that we will need one flag in the bindnode package, as we've chosen our default behaviors without paying attention to cbor-gen. Our defaults are also what make the most sense for go-ipld-prime and IPLD schemas, so it's a different starting point.

If we do need an option, then we should figure out where to stick it - `Prototype` and `Wrap` should likely have the ability to consume the option.

mvdan:
cc @warpfork @hannahhoward @rvagg @whyrusleeping as I've mentioned this in passing to all of you at some point.

rvagg:
> If we do need an option, then we should figure out where to stick it

Are we imagining a single option for everything we find that's different? Like `CborGenCompatibility`; or an option for each little thing, like the optional+nullable double pointer thing in #396, maybe something to do with map key ordering (though that one's trickier since it's a codec concern)?

mvdan:
I would go for one option for all the cbor-gen quirks; I doubt someone would want some of the quirks but not others. Ideally we won't need an option, but in practice I think we will, because the default behaviors might not be compatible.

PR #414, opened 2022-05-13
feat(bindnode): allow custom type conversions with options

rvagg:
This arises out of the attempts to push bindnode down into go-fil-markets for go-data-transfer vouchers. There are some types that are difficult to handle without just replacing them and having a conversion layer.

(Skip to custom_test.go for the code TL;DR)

To give some idea of the challenges, here's the list of types in their hierarchies that become difficult:

- `retrievalmarket.DealResponse`
    - `go-state-types/abi/TokenAmount` - `go-state-types/big/Int` - `struct{*big.Int}`
- `retrievalmarket.DealProposal`
    - `retrievalmarket.Params`
      - `go-state-types/abi/TokenAmount` - `go-state-types/big/Int` - `struct{*big.Int}`
- `retrievalmarket.DealPayment`
    - `go-address/Address` - `struct{string}`
    - `specs-actors/actors/builtin/paych/SignedVoucher`
      - `go-address/Address` - `struct{string}`
      - `specs-actors/actors/builtin/paych/ModVerifyParams`
        - `go-address/Address` - `struct{string}`
      - `go-state-types/big/Int` - `struct{*big.Int}`
      - `go-state-types/crypto/crypto.Signature`
        - this is a `byteprefix` union, `SigType` is first byte, `Data` is remainder, it could be read as `Bytes` and converted internally rather than implementing `byteprefix`

All of these types currently have their own custom, hand-rolled, `UnmarshalCBOR()` and `MarshalCBOR()` methods for cbor-gen. They're also not in go-fil-markets, so attempting to do anything direct with them (like adding custom methods) means chasing through lots of repos.

The approach I'm taking here is to add some options when calling bindnode, with these you can register converters for your custom types. In all of the above instances it's `Bytes` that we're dealing with, I think it's going to be the most common so I've only implemented that for now but I think a complete version of this would do it for all scalar kinds. Your schema still has to say that the field is `Bytes` / `bytes` but when traversing the Go type, if it encounters that type you registered for and it has the right kind then it'll run through your converter, either way.

The current version requires you supply two functions, a `to` and a `from` and they are checked for types. A previous iteration which you can see in 12a0ca48 had a converter type interface which had two functions and it'd need one for each kind we're allowing to register. I switched to supplying functions and validating with reflection for a slightly cleaner API. Plus, if you have existing functions like `NewFromBytes()` (which `Address` has) that matches the converter signature then you can just pass that in.

### DIFF
--- a/node/bindnode/api.go
+++ b/node/bindnode/api.go
@@ -301,7 +301,7 @@ func Wrap(ptrVal interface{}, schemaType schema.Type, options ...Option) schema.
 		// inferred.
 		verifyCompatibility(cfg, make(map[seenEntry]bool), goVal.Type(), schemaType)
 	}
-	return &_node{cfg: cfg, val: goVal, schemaType: schemaType}
+	return newNode(cfg, schemaType, goVal)
 }
 
 // TODO: consider making our own Node interface, like:
--- a/node/bindnode/api.go
+++ b/node/bindnode/api.go
@@ -27,10 +27,12 @@ import (
 // from it, so its underlying value will typically be nil. For example:
 //
 //     proto := bindnode.Prototype((*goType)(nil), schemaType)
-func Prototype(ptrType interface{}, schemaType schema.Type) schema.TypedPrototype {
+func Prototype(ptrType interface{}, schemaType schema.Type, options ...Option) schema.TypedPrototype {
 	if ptrType == nil && schemaType == nil {
 		panic("bindnode: either ptrType or schemaType must not be nil")
 	}
+
+	cfg := applyOptions(options...)
 
 	// TODO: if both are supplied, verify that they are compatible
 
@@ -50,11 +52,60 @@ func Prototype(ptrType interface{}, schemaType schema.Type) schema.TypedPrototyp
 		if schemaType == nil {
 			schemaType = inferSchema(goType, 0)
 		} else {
-			verifyCompatibility(make(map[seenEntry]bool), goType, schemaType)
+			verifyCompatibility(cfg, make(map[seenEntry]bool), goType, schemaType)
 		}
 	}
 
-	return &_prototype{schemaType: schemaType, goType: goType}
+	return &_prototype{cfg: cfg, schemaType: schemaType, goType: goType}
+}
+
+// CustomTypeConverter is an empty interface intended as a parent for the
+// various type converters
+type CustomTypeConverter interface {
+}
+
+// CustomTypeBytesConverter is able to convert byte slices to and from a
+// specific type that can't otherwise be handled by bindnode. Such a type will
+// likely not be instantiated by plain reflection and therefore need custom
+// logic to decode and/or instantiate.
+type CustomTypeBytesConverter interface {
+	FromBytes([]byte) (interface{}, error)
+	ToBytes(interface{}) ([]byte, error)
+}
+
+type config struct {
+	customConverters map[reflect.Type]CustomTypeConverter
+}
+
+// Option is able to apply custom options to the bindnode API
+type Option func(*config)
+
+// AddCustomTypeBytesConverter adds a CustomTypeConverter for a particular
+// type as referenced by ptrType. The CustomTypeConverter must be able to
+// handle that specific type, and the data model kind it's converting must be
+// present at the schema location it's encountered.
+func AddCustomTypeBytesConverter(ptrType interface{}, converter CustomTypeBytesConverter) Option {
+	val := reflect.ValueOf(ptrType)
+	if val.Kind() == reflect.Ptr {
+		val = val.Elem()
+	}
+	name := val.Type().Name()
+	if name == "" {
+		panic("not a named type")
+	}
+	return func(cfg *config) {
+		cfg.customConverters[val.Type()] = converter
+	}
+}
+
+func applyOptions(opt ...Option) config {
+	cfg := config{
+		customConverters: make(map[reflect.Type]CustomTypeConverter),
+	}
+	for _, o := range opt {
+		o(&cfg)
+	}
+	return cfg
 }
 
 // Wrap implements a schema.TypedNode given a non-nil pointer to a Go value and an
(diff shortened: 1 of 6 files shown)


rvagg:
Something I'm pondering after this is to be able to do this same thing for but for `Any`. That's quite invasive unfortunately because it needs to be done at every `AssignX()` and every `Lookup*()`, and every iterator `Next()`.

The obvious use-case now is for selectors, which need to be `Any` anyway. But for cbor-gen they are `Deferred`s. So I'd like to have a wrapper class that can hold an `ipld.Node` and has the `Deferred` methods as well. So it can act as both a bindnode type and a cbor-gen type at the same time and migration is relatively straightforward.

But, it also ends up being a backdoor for all sorts of whacky hacks. You just stop the schema short with an `Any`, and then defer to some other method of converting your types. You could even mix it with codegen or some other means of pulling in a typed builder. It'd allow you to do novel validation and conditionals too .. potentially lots of behaviours that we may or may not really want to happen, but such is the nature of pluggable systems.

node/bindnode/api.go
```diff
@@ -47,11 +49,78 @@ func Prototype(ptrType interface{}, schemaType schema.Type) schema.TypedPrototyp
...
+	return &_prototype{cfg: cfg, schemaType: schemaType, goType: goType}
+}
+
+type converter struct {
+	kindType reflect.Type
+	fromFunc reflect.Value // func(Kind) (*Custom, error)
+	toFunc   reflect.Value // func(*Custom) (Kind, error)
+}
+
+type config struct {
```
willscott, line 65: do you expect additional config members, or could you simplify this to:
```
type config map[reflect.Type]converter
```

node/bindnode/node.go
```diff
@@ -432,6 +442,27 @@ func (w *_node) AsBytes() ([]byte, error) {
...
+	if ok {
+		res := customConverter.toFunc.Call([]reflect.Value{w.val.Addr()})
+		if !res[1].IsNil() {
+			err, ok := res[1].Interface().(error)
+			if !ok {
+				return nil, fmt.Errorf("converter function did not return expected error: %v", res[1].Interface())
+			}
+			return nil, err
+		}
+		byts, ok := res[0].Interface().([]byte)
```
willscott, line 460: having to do the call in the abstract reflection world is going to be slower than if you can use a concrete function - i wonder if generics could be enough for defining these functions rather than needing to have it as a `reflect.Value`

node/bindnode/api.go
rvagg: :thumbsup:

node/bindnode/node.go
rvagg: yeah, I wasn't aware of the size of that cost .. looks like more than 70 times slower than a plain call. Which is not awesome. It's not the worst thing in the world but if we're trying to compete against cbor-gen we're already starting off on a back foot.

So a couple of options:

1. Use `interface{}` where for the custom type, and pass in that type via other means (probably just a null pointer like the rest of bindnode uses). Internally it can cast through the various options to find the one that is being used (e.g. `FromBytes([]byte) (interface{}, error)` & `ToBytes(interface{}) ([]byte, error)`. Then the API just needs that extra argument to pass the type. This steps back a little to the previous version of this API (12a0ca48c); it's not terrible but more verbose is slightly annoying.
2. Generics - this is the nice use-case for generics, but I don't believe we can even use reflection to access the concrete type being used, so we still need to derive that from somewhere. So we end up very similar to option 1, of needing to add the type. I think also we might have trouble casting to find out which one is being used and may have to use reflection to figure that out (like the current impl) but I'm speaking out of a bit of ignorance on what Go generics are giving us.
3. Live with the cost and 🤞 that https://github.com/golang/go/issues/49340 gets implemented sometime soon.

I'll try option 1 out and see how it goes. I'm making this API as "EXPERIMENTAL" in its docs to give a bit of permission to tweak and change it as we figure out how well it works for what we need.

rvagg:
OK, given the comment above about reflect function call performance, I've gone back to one option per kind for converters, so there's now a `AddCustomTypeXConverter` option for each scalar kind other than Null.

This now gives concrete functions to call and strong type checking by avoiding reflection. It makes the API more verbose, and it does force the use of `interface{}` on the user as a stand-in for their custom type, which is a little annoying (especially when you have to type check & convert within your converter functions). Plus you need to supply your type up-front that you want to register; so each of these option functions takes 3 arguments: type, convertFromFn, convertToFn.

I've implemented them all and added a variation for each in the main test that covers this. There's a bunch of whacky types is in there that are represented as one thing in the data model but another thing as a Go type. With them going both ways:

* a Bool as an int enum with 2 variations
* an Int that's the integer in words ("two thousand")
* a Float that's a wrapped big.Float for the user
* a String that's really an array of `|` separated []bytes
* a Bytes that's really a struct wrapping a string (like cid.Cid and graphsync.RequestId)
* a Bytes that's a wrapped big.Int, like go-state-types/big/Int
* a Link that's a reversed hex form of the hash digest (Bitcoin ID style)

willscott (commented):
While it feels pretty verbose, this seems like it should be able to have much more reasonable performance on these conversions.

rvagg:
Considering shortening the names: from `AddCustomTypeBoolConverter()` to just `BoolConverter()`, `TypedBoolConverter()`, `BoolTypeConverter()`. Maybe there's a better term than "converter" too. It's all a bit too verbose now that I'm typing them out.

hannahhoward (approved):
While the code itself is quite verbose, the end result seems to do exactly what we want it to, and using it doesn't seem too bad. Especially when I look at the markets PR, I think our options list will be pretty standard across all of filecoin, so this should be relatively convenient to implement.

+1 approved

rvagg:
This is close to ready given the work in #415 that's now merged in. I've been putting it into use in https://github.com/filecoin-project/go-fil-markets/pull/713 and have now switched back to go-data-transfer where I'm redoing https://github.com/filecoin-project/go-data-transfer/pull/305 and I'll use this branch there too. That may expose bugs or other problems in the implementation.

rvagg:
I think I'll close #352 when this is merged; there may be further work to do but so far we're making it work fairly nicely for common Filecoin chain types with this PR.

rvagg:
I've pulled #429 into this branch to make it a larger bindnode updates branch that I can use in go-fil-markets and go-data-transfer, along with some other updates I'm about to propose.

PR #415, opened 2022-05-17
feat(bindnode): add AddCustomTypeAnyConverter() to handle `Any` fields

rvagg:
Continuing #414, this adds a `AddCustomTypeAnyConverter()` option. It can handle maps, lists and all scalar kinds (including Null, which is a weird one since we're talking about `nullable` fields). A receiving Go type that is registered as a CustomTypeAnyConverter will receive a `datamodel.Node` and is expected to return one when converting to and from that custom Go type.

Still need to consider whether these need doing and how best to do it & test for it:

* `_mapIterator#Next()`
* `_listIterator#Next()`
* `_node#LookupByString()`
* `_node#LookupByIndex()`

and unions.. not fully sure about how unions fit into any of this yet.

The tests in here might be worth a look, there's two variations of an `Any` converter - one expects a Map that's of a very particular form and encodes and decodes it manually to and from a struct. The other one receives any `datamodel.Node`, encodes it as dag-cbor and stores those bytes (so it's too not dissimilar to the way selectors work with cbg.Deferred in the fil markets stack), and in the reverse it decodes them and provides a `datamodel.Node`.

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -560,8 +560,9 @@ func (w *_assembler) Representation() datamodel.NodeAssembler {
 type basicMapAssembler struct {
 	datamodel.MapAssembler
 
-	builder datamodel.NodeBuilder
-	parent  *_assembler
+	builder   datamodel.NodeBuilder
+	parent    *_assembler
+	converter *converter
 }
 
 func (w *basicMapAssembler) Finish() error {
@@ -569,7 +570,15 @@ func (w *basicMapAssembler) Finish() error {
 		return err
 	}
 	basicNode := w.builder.Build()
-	w.parent.createNonPtrVal().Set(reflect.ValueOf(basicNode))
+	if w.converter != nil {
+		typ, err := w.converter.customFromAny(basicNode)
+		if err != nil {
+			return err
+		}
+		w.parent.createNonPtrVal().Set(matchSettable(typ, reflect.ValueOf(basicNode)))
+	} else {
+		w.parent.createNonPtrVal().Set(reflect.ValueOf(basicNode))
+	}
 	if w.parent.finish != nil {
 		if err := w.parent.finish(); err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


node/bindnode/node.go
```diff
@@ -560,16 +560,25 @@ func (w *_assembler) Representation() datamodel.NodeAssembler {
...
 	basicNode := w.builder.Build()
-	w.parent.createNonPtrVal().Set(reflect.ValueOf(basicNode))
+	if w.converter != nil {
+		typ, err := w.converter.customFromAny(basicNode)
+		if err != nil {
+			return err
+		}
+		w.parent.createNonPtrVal().Set(matchSettable(typ, reflect.ValueOf(basicNode)))
+	} else {
+		w.parent.createNonPtrVal().Set(reflect.ValueOf(basicNode))
```
willscott, line 580: is there a way to not need these `reflect.ValueOf` calls on each map operation? is it possible to learn once when the converter is set up / constructed what the type it's going to produce is?

the general goal in how i saw bind-node structure before was that reflection is used when setting up the type system, but once it's set up, the types are fixed, so the actual interaction with data is fast.

rvagg: sadly it's reflection (almost) all the way down; although Dan was able to demonstrate that the perf wasn't too bad because we're limiting mostly to setters and getters with a little bit of inspection (most of that is done at setup, though).

the calls here are taking the value being set, and instantiating it if it's a nil pointer, then setting it to the appropriate form of the value it needs, which requires some finesse - sometimes they're pointers and sometimes not. And of course the `reflect.ValueOf` calls are needed here because `Set()` is a reflection method which can only set reflection value types.

rvagg:
Use of these new features is demonstrated in https://github.com/filecoin-project/go-fil-markets/pull/713

rvagg:
I've changed the option names to `TypedXConverter()`, and done some minor perf and organisational improvements.

hannahhoward (approved):
I think I understand what this does, and I'm generally ok with it.

I would say this: for the future, bindnode is a lot of reflection mad science, especially with this and #414. I think it will be important in subsequent PRs to document in comments and seperate doc files what all its doing.

rvagg:
* finished out the main TODOs for this feature (made it work for Lists and Maps)
* added a bunch of docs around the converters as well as just general notes through bindnode (didn't get into repr.go though, ran out of steam)
* I removed the typed functions for the converters, they just weren't as helpful as when I originally added them and were getting a bit in the way, plus the API docs look far too verbose with them in there.

Looks a bit better now:

Screenshot 2022-05-19 at 8 49 37 pm

I'm merging this into #414 and we'll deal with anything outstanding and final sign-off there.

Issue #348, opened 2022-02-03
node/bindnode: plug the footgun that is using the repr-level enum strings in Go


mvdan:
A user ran into a bug in the form of:

```
type GraphSyncLinkAction enum {
   | Present             ("p")
   | Missing             ("m")
} representation string
```

then backed in Go by a string, like:

```
type LinkAction string
const (
	LinkActionPresent = LinkAction("p")
	LinkActionMissing = LinkAction("m")
)
```

This didn't work:

> AsString: "p" is not a valid member of enum GraphSyncLinkAction

The fix was to use the strings for the enum members at the type level, rather than the representation level, which is how bindnode works:

```
type LinkAction string
const (
	LinkActionPresent = LinkAction("Present")
	LinkActionMissing = LinkAction("Missing")
)
```

This is all working as intended, but the mistake is easy to make, and the error didn't really point the user in the right direction.

We should probably do something here like a better error message, such as:

> AsString: "p" is not a valid member of enum GraphSyncLinkAction (bindnode works at the type level; did you mean "Present"?)

rvagg:
Genuine footgun, also hit in this commit (hence failing tests): https://github.com/ipld/go-ipld-prime/pull/350/commits/dfa7373f15dfc81027ecd0ce9760bc4befda2214

PR #430, opened 2022-05-24
fix(bindnode): more helpful error message for enum value footgun

rvagg:
Fixes: https://github.com/ipld/go-ipld-prime/issues/348

### DIFF
--- a/node/bindnode/api_test.go
+++ b/node/bindnode/api_test.go
@@ -1,0 +1,38 @@
+package bindnode_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


willscott (approved):

rvagg:
Note that this doesn't actually fix the other way around - if you decode `"Present"` you'll get `Action("Present")` which isn't part of the enum, it's a new typed string. So it won't actually solve a problem like https://github.com/ipld/go-ipld-prime/pull/350#issuecomment-1127268338, there's not much we can do about that I think though.

Issue #246, opened 2021-09-07
Cid.Undef IPLD de-serialization error


adlrocha:
If we use `cid.Undef` to create a new `datamodel.Link` with `cidLink`, it is serialized as `"/": "b"`, which means that when trying to de-serialize it, we get `cid is too short`. I don't think we should error in this case. We should probably not serialize `cid.Undef`, or use a `Null` type, to allow de-serialization in the other end. I don't know if this the expected behavior we wanted for `cid.Undef` or it is open to discussion.

### Background
I was serializing an IPLD node map with a bunch of links, some of them were `cid.Undef`, and I was getting `cid is too short`. I was blaming the Links which were actually set, when in the end it was the `cid.Undef` the one to blame.

willscott:
Note that part of this duality is because of the quirk in how we represent `Cid`'s in Golang. Since they are strings, a primitive type, they cannot be 'nil', and so `Undef` gets commonly used for that 'not yet set' case that may represent either explicit Absentness or Null

rvagg:
https://github.com/ipld/go-ipld-prime/pull/196 possibly helps with this?

PR #433, opened 2022-05-24
fix(codecs): coerce cid.Undef to null in dag{json,cbor} output

rvagg:
Otherwise we end up with invalid output like this:

dag-cbor: `0xd82a4100`

```
  d8 2a                                           #   tag(42)
    41                                            #     bytes(1)
      00                                          #       "\x00"
```

dag-json: `{"/":"b"}`

Fixes: https://github.com/ipld/go-ipld-prime/issues/246

---

@aschmahmann @willscott I'd appreciate your 👍 on this as it's a coercion rather than an error or other funky behaviour. But this probably isn't that simple:

* Unlike string, or int, the zero-value of CID doesn't have a round-trippable form that we've come up with for any of our codecs, so we're essentially making one up here, `null`. The alternative may be to define a null CID for both of these codecs - we could just use the values above and make it work across our current codec implementations (currently in JS, both of these result in errors because it can't decode a varint where it expects, IIRC round-tripping in Go does essentially the same thing).
* For typed nodes, this coercion could have blow-back if you have a link field that's not `nullable` or `optional`, then a round-trip with a `cid.Undef` is going to result in a `null` in that position which might be a problem. But we could push that problem up the stack and make it a concern for bindnode and friends to figure out. They could even use `cid.Undef` for non-nullable non-optional fields that come out as `null`.
* We have the option of erroring on this if we want - make it a user problem, don't give us `cid.Undef`, it's not allowed.

_(Related discussion in https://github.com/ipld/go-ipld-prime/issues/191 and the currently rejected PR that would error on `Absent` fields: https://github.com/ipld/go-ipld-prime/pull/196)._

@Stebalien @vmx you might want to weigh in too because we're talking possible codec behaviour changes.

### DIFF
--- a/codec/dagcbor/marshal.go
+++ b/codec/dagcbor/marshal.go
@@ -144,6 +144,9 @@ func marshal(n datamodel.Node, tk *tok.Token, sink shared.TokenSink, options Enc
 		}
 		switch lnk := v.(type) {
 		case cidlink.Link:
+			if !lnk.Cid.Defined() {
+				return fmt.Errorf("encoding undefined CIDs are not supported by this codec")
+			}
 			tk.Type = tok.TBytes
 			tk.Bytes = append([]byte{0}, lnk.Bytes()...)
 			tk.Tagged = true
--- a/codec/dagcbor/marshal_test.go
+++ b/codec/dagcbor/marshal_test.go
@@ -7,7 +7,11 @@ import (
 	"time"
 
 	qt "github.com/frankban/quicktest"
+	"github.com/ipfs/go-cid"
+	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/fluent/qp"
+	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
 	basicnode "github.com/ipld/go-ipld-prime/node/basic"
 	"github.com/ipld/go-ipld-prime/testutil/garbage"
 )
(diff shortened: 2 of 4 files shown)


vmx:
I'm not sure I've grasp the full impact of this. But serializing a `cid.Undef` to `null` makes sense to me.

rvagg:
I'm thinking through what I might do in JS and I don't think I have the ability to round-trip a null CID, we don't have (and haven't needed) an equivalent of `Undef` in JS, `null` works just fine.

@vmx `Undef` is normally used as the "unset" / zero form of a CID when it's not a pointer, so can't be `nil`. The equivalent of `""` for the zero form of a `string` or `0` for an int. We currently have a gap in the implementation that it will try and serialise it but it's really just an empty byte array, which isn't a CID at all. So we have to do _something_.

rvagg:
Here's a case for just erroring—the old go-ipld-cbor will if it gets an `Undef`: https://github.com/ipfs/go-ipld-cbor/blob/3c41942dd2857bcf82bea83160439ef8f1cac5c9/node.go#L541

vmx:
@rvagg I'm not sure I'm missing something here. I don't understand the relation to codecs. It seems to be a Go specific issue. I wouldn't expect codecs to support/round-trip invalid CIDs and I also think we should try as hard as possible to not leak language specific things into codecs.

willscott:
I also have mixed feelings on this one. Having errors when nil/undef cids is a good indication that you should be checking and converting that case into the null case of a nullable field. That said, it is boilerplate that gets added in at all those places and does seem right that we can just automatically interpret that to the null case when possible.

rvagg:
OK, no appetite for changing codec specs to allow some form of zero-CID, makes this a bit easier.

Error vs coerce: I coming down on the side of erroring - it may lead to boilerplate, but if you end up with an undefined CID maybe it should have been a pointer and therefore be `optional` or `nullable`, in which case this would be taken care of without the need for boilerplate. Plus, is it likely that if you're allowing an undefined CID to get this low down in the stack that maybe you've missed adding some checking much much further up the stack?

Erroring does give us some good symmetry as well, there's no magic or surprises and ensures round-trip stability. Make your input the same as what you should get on the output, and we're not making undefined CIDs anywhere on the way out, it'll just be a null Node.

So latest commit makes it an error; reviews or further discussion?

PR #435, opened 2022-05-26
add option to not parse beyond end of structure

petar:
Resolves https://github.com/ipld/go-ipld-prime/issues/374.

### DIFF
--- a/codec/dagjson/unmarshal.go
+++ b/codec/dagjson/unmarshal.go
@@ -32,6 +32,10 @@ type DecodeOptions struct {
 	// If true, parse DAG-JSON `{"/":{"bytes":"base64 bytes..."}}` as a Bytes kind
 	// node rather than nested plain maps
 	ParseBytes bool
+
+	// If true, the decoder stops reading from the stream at the end of the JSON structure.
+	// I.e. it does not slurp remaining whitespaces and EOF.
+	DontParseBeyondEnd bool
 }
 
 // Decode deserializes data from the given io.Reader and feeds it into the given datamodel.NodeAssembler.
@@ -42,6 +46,9 @@ func (cfg DecodeOptions) Decode(na datamodel.NodeAssembler, r io.Reader) error {
 	err := Unmarshal(na, json.NewDecoder(r), cfg)
 	if err != nil {
 		return err
+	}
+	if cfg.DontParseBeyondEnd {
+		return nil
 	}
 	// Slurp any remaining whitespace.
 	//  This behavior may be due for review.
--- a/codec/dagjson/unmarshal.go
+++ b/codec/dagjson/unmarshal.go
@@ -34,7 +34,10 @@ type DecodeOptions struct {
 	ParseBytes bool
 
 	// If true, the decoder stops reading from the stream at the end of the JSON structure.
-	// I.e. it does not slurp remaining whitespaces and EOF.
+	// i.e. it does not slurp remaining whitespaces and EOF.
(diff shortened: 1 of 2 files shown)


codec/dagjson/unmarshal.go
```diff
@@ -32,6 +32,10 @@ type DecodeOptions struct {
 	// If true, parse DAG-JSON `{"/":{"bytes":"base64 bytes..."}}` as a Bytes kind
 	// node rather than nested plain maps
 	ParseBytes bool
+
+	// If true, the decoder stops reading from the stream at the end of the JSON structure.
+	// I.e. it does not slurp remaining whitespaces and EOF.
```
rvagg, line 37:
```suggestion
	// i.e. it does not slurp remaining whitespaces and EOF.
	// As per standard IPLD behavior, the parser considers the entire block to be
	// part of the JSON structure and will error if there is extraneous
	// non-whitespace data.
```

Something like that? I just want to make it clear that this is non-standard for IPLD, that it's appropriate to error if there's extraneous data and that this mode is for when you're using the parser for some other purpose than decoding discrete IPLD blocks.

rvagg:
this is fine, just needs a test or two to exercise is

codec/dagjson/unmarshal.go
```diff
@@ -43,6 +47,9 @@ func (cfg DecodeOptions) Decode(na datamodel.NodeAssembler, r io.Reader) error {
 	if err != nil {
 		return err
 	}
+	if cfg.DontParseBeyondEnd {
+		return nil
+	}
```
aschmahmann, line 52: @petar IIUC this is perhaps a nice config option, but only slightly helps fix the problem you're likely currently worrying about (i.e. streaming decodes of multiple dag-json objects for Reframe using Edelweiss).

Since the specific application format (in this case Reframe's HTTP+dag-json transport) has its own way of dealing with concatenating dag-json blobs (in this case appending `\n`) you're going to need some custom code anyhow to parse the  `\n`.

At that point is it so different from just using `Unmarshal(na, json.NewDecoder(r), cfg)` directly and then trying to slurp up one more `\n` before continuing instead of using `Decode`?

petar: The issue with this proposal is that `json.NewDecoder` won't decode bytes and other IPLD-specific objects. Right?

The solution to Edelweiss' problem could be composed of two steps:
- this PR, which makes sure nothing is slurped after the JSON object, and
- a manual reader.ReadChar, following every invocation of ipld.StreamingUnmarshal(dagjson.Decoder, ...) in the Edelweiss-generated code for reading multiple `\n`-separated results.

aschmahmann: I just wanted to flag more concretely that this PR just allows for using:

`DecodeOptions{ ParseLinks : true, ParseBytes: true, DontParseBeyondEnd : true }.Decode(na, r)`

rather than:

`Unmarshal(na, json.NewDecoder(r), DecodeOptions{ ParseLinks : true, ParseBytes: true })`.

Is this config mostly about helping discoverability so people know how to do this (i.e. is the Unmarshal code path not obvious enough)?

petar:
@rvagg should be good now.

codec/dagjson/unmarshal.go
petar: It turns out that Unmarshal is deprecated.

rvagg (approved):

PR #438, opened 2022-06-13
trigger a release to capture the new options in dagjson.Decode

petar:
The new dagjson.Decode option is utilized by Edelweiss and Delegated Routing (Reframe), and transitively by Hydra and StoreTheIndex.


rvagg (changes requested):
Yeah ok, I was going to wait till I got #414 in, but there has been quite a bit of work since 0.16.0 that we probably should get out!

I need to do a CHANGELOG.md update for this though, and it's going to have to be an 0.17.0 so I'm going to block this PR and will work on an update this week that gets you a release.

rvagg:
Will take care of a release in https://github.com/ipld/go-ipld-prime/pull/440

petar:
Sounds good. Thank you.

Issue #387, opened 2022-03-15
Create a release checklist/SOP


BigLep:
It's awesome a release SOP has been started here: https://github.com/ipld/go-ipld-prime/issues/384#issuecomment-1061964620 🎉 

A few things.  Can we:
1. Make it discoverable? - I think this is mostly just extracting it out of a specific release issue.  Whether it become an issue that is cloned from, an issue template, or a markdown file that copy/paste from, it doesn't really matter.
2. Add a checklist component? - I like how the SOP provides explanatory information and rationale.  The further improvement would be to make it easy for the person releasing to know where they are and show their status.

For reference, here are some other release templates:
- https://github.com/ipfs/go-ipfs/blob/master/docs/RELEASE_ISSUE_TEMPLATE.md
- https://github.com/libp2p/go-libp2p/blob/master/.github/ISSUE_TEMPLATE/release.md

PR #440, opened 2022-06-14
Prepare v0.17.0

rvagg:
(Ref: #438)

Lots of stuff built up in here that needs to get out. There's a couple of concerning changes that warrant a semver-minor bump (semver-major in an ideal world).

I have #414 and possibly #437 around the corner but they can go in a v0.17.1 as soon as they're in and we can tag a bit more frequently from then on.

----

* [x] Add new h3 to `CHANGELOG.md` under **Released Changes** with curated list of changes and links to relevant PRs
  * [x] Highlight any potentially breaking or disruptive changes
  * [x] Add special-thanks call-outs to contributors making significant contributions
* [x] Update **Planned/Upcoming Changes** section to remove completed items and add newly upcoming, but incomplete items
* [x] Bump version number appropriately in `version.json`
* [x] Propose release via pull request, merge after enough time for async global feedback
* [ ] Create GitHub [release](https://github.com/ipld/go-ipld-prime/releases) with the new tag, copying the new `CHANGELOG.md` contents
* [ ] Announce on relevant Discord/Matrix/Slack channel(s)
* [ ] (Optional) Announce to shipped@protocol.ai

Issue #384, opened 2022-03-08
Release v0.16.0


BigLep:
Note: as part of doing this release, we'll develop a release issue SOP.

Things we'll document:
* Even vs Odd releases: https://gist.github.com/warpfork/98d2f4060c68a565e8ad18ea4814c25f

warpfork:
Done: https://github.com/ipld/go-ipld-prime/releases/tag/v0.16.0

PR #442, opened 2022-06-15
feat: add release checklist

rvagg:
Closes: https://github.com/ipld/go-ipld-prime/issues/387
Ref: https://github.com/ipld/go-ipld-prime/issues/384#issuecomment-1061964620

Putting into practice @ https://github.com/ipld/go-ipld-prime/pull/440 for v0.17.0

### DIFF
--- a/HACKME_releases.md
+++ b/HACKME_releases.md
@@ -27,8 +27,8 @@ git log --all --graph --date-order --abbrev-commit --decorate --oneline
 The CHANGELOG should be informative for developers wanting to know what changes may pose a risk (highlight these!) and what changes introduce features they may be interested in using.
 
 1. Group commits to subsystem to create a two-level list. Subsections can include "Data Model", "Schemas", "Bindnode", "Selectors", "Codecs", and the meta-category of "Build" to describe changes local to the repository and not necessarily relevant to API consumers.
-2. If there are breaking, or potentially breaking changes, list them under a `### 🛠 Breaking Changes` section.
-3. Otherwise, prune the list of commits down to the set of changes relevant to users, and list them under a `### 🔦 Highlights` section.
+2. If there are breaking, or potentially breaking changes, list them under a `#### 🛠 Breaking Changes` section.
+3. Otherwise, prune the list of commits down to the set of changes relevant to users, and list them under a `#### 🔦 Highlights` section.
 
 Note that there is also a **Planned/Upcoming Changes** section near the top of the CHANGELOG.md. Update this to remove _done_ items and add other items that may be nearing completion but not yet released.
 
--- a/HACKME_releases.md
+++ b/HACKME_releases.md
@@ -1,0 +1,69 @@
+# Making go-ipld-prime Releases
+
+## Versioning strategy
+
+go-ipld-prime follows **[WarpVer](https://gist.github.com/warpfork/98d2f4060c68a565e8ad18ea4814c25f)**, a form of SemVer that never bumps the major version number and uses minor version numbers to indicate degree of *changeness*: **even numbers should be easy upgrades; odd numbers may change things**. The patch version number is rarely used in this scheme.
+
+## CHANGELOG.md
+
+There is a CHANGELOG.md, it should be relevant and updated. Notable items in the commit history since the last release should be included. Where possible and practical, links to relevant pull requests or other issues with discussions on the items should be included.
+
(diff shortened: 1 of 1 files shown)


HACKME_releases.md
```diff
@@ -0,0 +1,69 @@
...
+
+Use the GitHub UI to make a [release](https://github.com/ipld/go-ipld-prime/releases), copying in the contents of the CHANGELOG.md for that release.
+
+Drop in a note to the appropriate Matrix/Discord/Slack channel(s) for IPLD about the release.
+
+Optional: Protocol Labs staff can send an email to shipped@protocol.ai to describe the release, these are typically well-read and appreciated.
+
+## Checklist
+
+A release proposal pull request may use the following markdown checklist to help ensure the requisite steps are taken:
```
BigLep, line 57: I see this checklist is added to a PR.  It may be better to do it as an issue so:
1. Can create in advance and store extra metadata for the community like anticipated ship date or link other release PRs to
2. It stays open even as specific releases PRs close.  The issue can be closed when the release is done-done-done.

BigLep (approved):

HACKME_releases.md
rvagg:
```suggestion
Prior to opening a release proposal pull request, create an issue with the following markdown checklist to help ensure the requisite steps are taken. The issue can also be used to alert subscribed developers to the timeframe and the approximate scope of changes in the release.
```

rvagg:
@BigLep:

> It seems like we have a decent amount of manual steps here in creating the changelog

That's why I'm recommending using [changelog-maker](https://github.com/nodejs/changelog-maker) to help craft it. I wrote that tool in 2015 to help better communicate changes in Node.js releases and it's since become a standard part of the process. It's helped by a formalised commit message scheme that can be pulled apart into sub-sections automatically, and some metadata that's now mandatory in each merged commit. I've done some updating of it (@ https://github.com/nodejs/changelog-maker/pull/130) so it doesn't need the metadata on commits to match PRs so it can be used on any repo. With descriptive enough commits I think it should be fairly minimal work to group things together, remove extraneous ones and do a bit of renaming an highlighting to make a workable changelog entry.

I've reduced the manual steps even more from the ones that Eric has previously taken and documented in https://github.com/ipld/go-ipld-prime/issues/384#issuecomment-1061964620, I think it's more manageable now. With better commit conventions it could be even better but one thing at a time, people get precious about their git conventions.

> some more headers/emojis

Oh, you're stretching me there with emojis ... But OK, I'll do the headers thing in #440 and come back and update this.

rvagg:
Added 🤮 emojis and grouped the changes into two sections to highlight "breaking" changes in #440; updated the docs here to suggest that.

HACKME_releases.md
```diff
@@ -0,0 +1,69 @@
...
+
+Use the GitHub UI to make a [release](https://github.com/ipld/go-ipld-prime/releases), copying in the contents of the CHANGELOG.md for that release.
+
+Drop in a note to the appropriate Matrix/Discord/Slack channel(s) for IPLD about the release.
+
+Optional: Protocol Labs staff can send an email to shipped@protocol.ai to describe the release, these are typically well-read and appreciated.
+
+## Checklist
+
+Prior to opening a release proposal pull request, create an issue with the following markdown checklist to help ensure the requisite steps are taken. The issue can also be used to alert subscribed developers to the timeframe and the approximate scope of changes in the release.
```
RangerMauve, line 57: Would it make sense to create a GH issue template for this?

RangerMauve (approved):

HACKME_releases.md
rvagg: I don't really want to do this because we currently don't have any issue templates, so it introduces an extra step in the New Issue flow for everyone, just to support releases for one person.

If/when we start adding templates for other types of issues then we could add this one, otherwise it's going to be an extra "do you want to file an issue or make a release" in-between step for everyone.

rvagg:
Going to punt on the issue template [thread](https://github.com/ipld/go-ipld-prime/pull/442#discussion_r899391534) for further discussion and _maybe_ implementation later; just not in this current version.

HACKME_releases.md
RangerMauve: Yeah totally fair. Revisiting once there are other templates seems like a good idea. 😁

Issue #320, opened 2021-12-27
APIs for creating incrementally modified nodes


warpfork:
We need APIs for creating new nodes which include content from a previous node, along with some additions or subtractions.

The first step is figuring out what pattern we use when placing these APIs.  Critical check: the API should possible to implement while using COW (copy-on-write) semantics internally.

There might be more than one form of API matching this general description (depending on if it's mostly additions; mostly subtractions; if it's insertions-with-opinions-on-ordering, etc).

I suspect that, like has been a recurring theme lately, we should both introduce the feature itself via feature detection, and then probably accompany it by helper functions which expose the desired API to end users, and then tries to detect and use the feature, or does fallbacks quietly.  (This might turn out a little tricky, though, considering actual objects must be returned for assembly to continue working on, and we don't want that to have to wrap them and end up with more allocations.)

It should be possible to map an "IPLD Patch" operation (e.g., something very comparable to [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902)) onto whatever APIs we come up with.  (The library API can focus on single elements, rather than any of the pathing parts of the Patch declarations -- but any of the semantics about e.g. order of insertion should be alignable.)

smrz2001:
Thanks for the discussions, @warpfork!

I've captured some initial thoughts [here](https://gist.github.com/smrz2001/bbeaf4a6ae03996d643bf37308ae4f2c).

smrz2001:
I did some more thinking and prototyping, and realized that @warpfork's [patch PR](https://github.com/ipld/go-ipld-prime/pull/350) provides a much better base to start from than what I'd been thinking.

In hindsight, perhaps that should have been the logical sequence of steps anyway: first a patch spec, then an update API to optimize patch application.

The IPLD `traversal` package seems to already have a lot of the bookkeeping I had in my [earlier notes](https://gist.github.com/smrz2001/bbeaf4a6ae03996d643bf37308ae4f2c).

Incremental updates would then need to focus on optimizing some of the `traversal` logic/bookkeeping and adding some companion copy-on-write optimizations in a particular `Node` implementation.

[Here](https://github.com/smrz2001/go-ipld-prime/pull/1/files) are some tweaks off of the patch PR and [here's](https://github.com/smrz2001/go-ipld-adl-hamt/pull/1) a very crude implementation in `go-ipld-adl-hamt`.

[...]

Issue #443, opened 2022-06-15
Node deserialization of Graphsync Extensions & data transfer vouchers


hannahhoward:
I've been thinking through optimization of decoding of Graphsync Extensions & data transfer vouchers & IPLD's type system, as well as how they relate to bindnode & #437

The basic structure of both of Graphsync Extensions and Data Transfer Vouchers is as follows:
- a key that identifies the kind of data
- a value that contains the data, whose type is determined by the key

The primary difference between Graphsync Extensions & data transfer vouchers is a single graphsync request or response can contain multiple extensions where a data transfer message contains only a single voucher

The key complication for both of these systems is:
- the set of known key -> value type conversions are set at run time (by higher level libraries using go-data-transfer or go-graphsync)
- there data is transmitted over the network, implying that we may receive types we simply don't know about

The current schema used for Graphsync extensions is:

```ipldsch
# Extensions as a name:data map where the data is any arbitrary, valid Node
type GraphSyncExtensions { String : nullable Any }
```

The abbrievated current schema used for a data transfer message is:

```ipldsch
type TypeIdentifier string

# ...

type TransferRequest struct {
# ... other fields ...
	VoucherPtr            nullable Any            (rename "Vouch")
	VoucherTypeIdentifier          TypeIdentifier (rename "VTyp")
}

```

First, let's look at compatibility between IPLD Schemas & Graphsync Extensions. Let's say we had a list of extension names to known extension types. A better schema for graphsync extensions would be something like

```
type GraphSyncExtensions struct {
   KnownKind1 optional nullable KnownKind1 (rename "KnownExtensionName1")
   KnownKind2 optional nullable KnownKind2 (rename "KnownExtensionName2")
} representation map
```

Assuming the schema builders provide APIs for building this at run time (not sure if they do), we could theoretically assemble such a type once we had a stable list of extensions known. However, we still have the problem of data from the network, and what to do values we don't recognize. 

The optional allows us to figure out what to do with missing fields. It doesn't really say what to do with surplus fields in a representation. the current bindnode code will error -- https://github.com/ipld/go-ipld-prime/blob/master/node/bindnode/node.go#L809

Now let's look at data transfer vouchers. . Let's say we had a list of voucher type identifiers to known voucher types. A better schema for data transfer vouchers would be (cause there is only one present in a message):

```

```ipldsch
type Voucher union {
  | KnownVoucherKind1 "KnownVoucherTypeName1"
[...]

PR #445, opened 2022-06-20
feat: ipld amend

smrz2001:
This PR implements a possible solution for [incrementally modifying IPLD nodes](https://github.com/ipld/go-ipld-prime/issues/320).

It uses the same interface as @warpfork's patch [implementation](https://github.com/ipld/go-ipld-prime/pull/350) but accumulates updates and internal state to provide a more optimized `Node` "lens" for copy-on-write during `Encode`.

The [draft PR](https://github.com/smrz2001/go-ipld-prime/pull/1) included some additional discussions and [benchmarking results](https://github.com/smrz2001/go-ipld-prime/pull/1#issuecomment-1152992683) comparing this PR's implementation with the current `patch` implementation.

cc @rvagg @RangerMauve @warpfork @BigLep


smrz2001:
Also, @rvagg, how would you suggest I resolve this error?
https://github.com/ipld/go-ipld-prime/runs/6974693099?check_suite_focus=true

RangerMauve:
> I think a good subsequent PR might be one that integrates amend into walk and focus so that the latter can take advantage of amend's performance enhancements while continuing to provide link traversal, budgets, etc., and then switching patch/eval back to using the updated traversal/focus.

Could you elaborate a bit on what integrating amend into walk and focus would entail?

> and to your specific point, @RangerMauve, I can update patch/eval to use traversal/amend instead of traversal/focus

Question, would it make sense to put all the `amend` stuff under the `patch` namespace and merge the two concepts? Or are they so different that they need to have different names? One reason I'm thinking about this direction is that the names seem very similar and it could be confusing to API consumers to know what the difference is. It'd be nice if we could say "Patch has a high level API for patch sets (current /patch), and a lower level API for modifications (current /amend)".

smrz2001:
> Could you elaborate a bit on what integrating amend into walk and focus would entail?
Yes, `walk` and `focus` involve similar operations as I've implemented in `amend`, (hence my thought of integrating them) - traversing a node graph, trying to reach a particular child node, modifying such a child, etc. In addition, they also traverse links as they encounter them and manage a budget for such link traversal, etc.

To integrate `amend`, I would replace the code that `walk` and `focus` currently use for traversal, iteration, etc. with that implemented in `amend`, which minimizes the operations and memory required. Won't be straightforward but I feel that it's doable.

I'm also debating whether to just add missing features to `amend` and let it grow organically based on user requirements instead of causing a ton of code churn while trying to fit it into `walk` and `focus`. If the IPLD team can provide a bit of guidance around which features are the most important for amend to have as it evolves, I can start adding them in order of importance. I feel like link traversal should definitely be added next. 

> Question, would it make sense to put all the `amend` stuff under the `patch` namespace and merge the two concepts? Or are they so different that they need to have different names? One reason I'm thinking about this direction is that the names seem very similar and it could be confusing to API consumers to know what the difference is. It'd be nice if we could say "Patch has a high level API for patch sets (current /patch), and a lower level API for modifications (current /amend)".
Yes, that makes sense. The main reason I used different terminology and separate packages was to be able to refer to the two implementations individually while comparing them, but, at this point, merging them seems to be appropriate.

If I were to merge them, my thinking was to replace the body of `patch/eval` with what is currently `amend/eval`, ie. an `Amender` is now used to apply patches instead of `focus`. That would have the effect of not (currently) allowing link traversal but, as I mentioned previously, I can add support for that soon. 

Does that seem reasonable?

smrz2001:
Actually, I think I'm going to experiment a bit with the amend/walk/focus integration this week and let you know how it goes. Now that I'm looking at it again, I think it might be less work than I was initially thinking and could potentially fit in the same PR.

Integrating it there would allow amend's efficiencies to be available to two different APIs - patch and traversal.

@RangerMauve

rvagg:
@smrz2001 can you give me your thoughts on the performance gains from this impl over FocusedTransform? They're essentially having to perform the same operations by the time you get to final node realisation aren't they? What am I not seeing here? Are the transform operations having to double (and triple etc.) up on parent node assembling as they deal with batches of operations?

And yeah, now I look at this code, I can see the logic of going back and seeing if you can adapt the walk and focus transforms to use it. That would also help prove the concept.

I think that making sure this works across an ADL would be a good step in the process too. The HAMT ADL is actively getting work, Masih has been pushing new changes to it as they're using it for the Filecoin Indexer service I believe, so it's getting closer to feature-complete. Alternatively, the most complete ADL we have is go-unixfsnode. I'm not sure we want those pulled into the repo to test here but if you can figure it out then wiring up a transformation over a working ADL would be really interesting and confirm that the layering is working as expected. Although, it's possible that the pieces around finalisation are a bit too rough to make it all work, so don't spend too much time spinning wheels on that.

Also, is amending a non-list or map root node intended to work? Looking at the Any amender, I can't see how that would work. Maybe it doesn't make sense, but shouldn't I be able to amend using a `/` path and replace whatever node it finds with something new?

smrz2001:
> @smrz2001 can you give me your thoughts on the performance gains from this impl over FocusedTransform? They're essentially having to perform the same operations by the time you get to final node realisation aren't they? What am I not seeing here? Are the transform operations having to double (and triple etc.) up on parent node assembling as they deal with batches of operations?

That's exactly right, @rvagg - the transform operations must assemble and build a new `Node` for each update operation. This doesn't matter a ton for a small set of updates, but `amend`'s amortization is more noticeable as the number of updates increases.

In the case of `amend`, `Build()` and `Encode()` use the `Node` **_view_** of the `Amender`, which presents the cumulative result of the updates without needing to perform _any_ assembly at all.

> And yeah, now I look at this code, I can see the logic of going back and seeing if you can adapt the walk and focus transforms to use it. That would also help prove the concept.

Ok, excellent. I already started looking at this and will make updates in the current PR. I did run into some blockers and might need guidance - will let you know.

> I think that making sure this works across an ADL would be a good step in the process too. The HAMT ADL is actively getting work, Masih has been pushing new changes to it as they're using it for the Filecoin Indexer service I believe, so it's getting closer to feature-complete. Alternatively, the most complete ADL we have is go-unixfsnode. I'm not sure we want those pulled into the repo to test here but if you can figure it out then wiring up a transformation over a working ADL would be really interesting and confirm that the layering is working as expected. Although, it's possible that the pieces around finalisation are a bit too rough to make it all work, so don't spend too much time spinning wheels on that.

Great idea!

Coincidentally, being able to efficiently update a HAMT (plus simplifying some of the ugly logic I had to add in `dag-jose`) was my original impetus for this project 😄 I would love to see updating HAMTs through once it's in a good place to test. I'm sure `amend` will need some enhancements at that point as well.

I'll start playing around with `go-unixfsnode` and see if I can set up a good test.

> Also, is amending a non-list or map root node intended to work? Looking at the Any amender, I can't see how that would work. Maybe it doesn't make sense, but shouldn't I be able to amend using a `/` path and replace whatever node it finds with something new?

Good question. I had the same question during my implementation and was of the opinion that really only recursive `Nodes` need updates. Any primitive `Nodes` can just be recreated.

`anyAmender` acts as a wrapper for primitive leaf `Nodes` (and possibly for other (recursive) `Amenders`) and simplifies some of the iteration and processing logic without really providing any intelligence of its own.

I do see what you're saying though. I could update its `Replace()` method to check whether `path` is `/` and apply the update. I'll go ahead and do that.

As a side thought, I'm wondering if I should store a pointer to the base `Node` inside `anyAmender`. Recursive `Nodes` are already pointers but not primitive `Nodes`. For most primitive `Nodes` this probably doesn't matter but `String` or `Bytes` types could be large and creating an `anyAmender` wrapper around them currently results in a copy. Using pointers would reduce the memory consumption even further and probably justify `anyAmenders` existence a bit more.

smrz2001:
@rvagg, after rereading your question regarding perf, I reworded some of my explanation. Please let me know if that makes sense.

To summarize, the main reason behind the new implementation's better performance is that it completely bypasses `Node` (re)assembly.

An `Amender` masquerades as a fully updated `Node` by letting lookup and iteration work as they would on a `Node` with updates already applied, but without actually doing any reassembly, not even during `Encode()`.

smrz2001:
@rvagg @RangerMauve, I've completed the integration with `focus`. All missing features, such as link traversal, progress tracking, link/node budgets, etc. are now included, and all `focus` and `patch` tests are passing.

I'm still mulling over how to integrate with `walk` and will attempt to do that soon, as well as do some testing using an ADL, like you recommended, @rvagg.

I do also see some other failing tests, and will dig into them.

BigLep:
@smrz2001 : is there anything missing before a final review/merge?

Can you please also provide updated benchmarks?  We'd like to include those in the release notes.

2022-07-19 triage conversation:
1. Wondering about the module for "amend".  Maybe it should be in its own module or be under traversal/patch?  This could be a good topic for the community call.

smrz2001:
> @smrz2001 : is there anything missing before a final review/merge?

I don't think so, @BigLep, unless @rvagg or @RangerMauve would like to see something else - I'd be happy to make more changes.

I haven't gotten a chance to add more tests for corner cases but don't think that's blocking. I'll add some with the next PR to this code. I'm planning to put up another one in a bit anyway for adding and exposing plain `Map` and `List` interfaces for better DX around recursive nodes (e.g. `Map.Put(key string, value datamodel.Node)` instead of `MapBuilder.BeginMap()...[do stuff]...MapBuilder.Finish()` or `qp.BuildMap()...[do stuff]` etc.), which has been one of my biggest pain points.

If any specific tests are needed for more confidence though, I can add them, no problem.

To @rvagg's question about integrating with `go-unixfsnode`, I'm definitely having some trouble figuring out how to wire things together correctly and will continue trying to figure that out because I do want to make sure `amend` works correctly when integrated with ADLs.

Once I have a `Map` interface in place, I'm actually also considering rejiggering `go-ipld-adl-hamt` to use that and the update logic so that it becomes more accessible for other, plain map-like uses, while also proving that the layering works correctly. I might just do that before looking at `go-unixfsnode`.

> 2022-07-19 triage conversation:
> 
> 1. Wondering about the module for "amend".  Maybe it should be in its own module or be under traversal/patch?  This could be a good topic for the community call.

I tried really hard to keep the packages clean and separate 😭 but had to give up because of circular dependencies while integrating `amend` with `focus`. More stuff will need to be shuffled around to get that to work (e.g. the things in `traversal/common.go` and `traversal/fns.go` are required in both `traversal/focus` and a separate `amend` package, which means the latter cannot be imported by the former). I'm not a Go expert by any means, so I'm not sure if there's some obvious, idiomatic solution I'm missing here.

I actually ended up going backwards and pulling all `amend` files out into the main `traversal` package in the current version in order to integrate them.

I feel that `amend` (i.e. updating IPLD nodes) is a more generic concept than `patch` (apply JSON patch sets), which is a specific type of update, but I'm open to suggestions. This makes me think that both `traversal` and `patch` should be consumers of the `amend` API. The integration with `focus`, `Progress`, etc. makes that difficult though.

cc @aschmahmann, since you expressed some interest in this PR.

RangerMauve:
> Once I have a Map interface in place, I'm actually also considering rejiggering go-ipld-adl-hamt to use that and the update logic so that it becomes more accessible for other, plain map-like uses, while also proving that the layering works correctly. I might just do that before looking at go-unixfsnode.

That would be cool as heck!

> I feel that amend (i.e. updating IPLD nodes) is a more generic concept than patch (apply JSON patch sets), which is a specific type of update, but I'm open to suggestions. This makes me think that both traversal and patch should be consumers of the amend API. The integration with focus, Progress, etc. makes that difficult though.

Honestly, I agree with this reasoning. Would it make sense to put `amend` in the top level? Or would this even be something that falls under `go-ipld-prime/node/amend`? cc @rvagg 

Might be useful to talk about this with higher bandwidth at the next IPLD community call. 🤷

RangerMauve:
Didn't work under traversal due to circular dependencies.

RangerMauve:
Would it make sense to have a call with @smrz2001 and @rvagg (maybe @aschmahmann) and just devote some time to finalizing what we need to merge this so that it doesn't sit in limbo for too long?

RangerMauve:
The concept that this is a "writable ADL" is really curious though. 😅

smrz2001:
> Would it make sense to have a call with @smrz2001 and @rvagg (maybe @aschmahmann) and just devote some time to finalizing what we need to merge this so that it doesn't sit in limbo for too long?

This would be great, @RangerMauve!

smrz2001:
Moving the "writable ADL" [discussion](https://github.com/smrz2001/go-ipld-prime/pull/4#issuecomment-1201716312) here:
>> I also wanted to pass another thought by you, @RangerMauve.
>> 
>> I can't get the idea out of my head that what `amend` really is, is an updatable ADL. It takes a source `Node`, allows the accumulation of updates, then "presents" a new `Node` to consumers.
>> 
>> I could easily turn `Amender.Build` into `Reify` and add `Substrate` to expose the source `Node`. 
>>
> The concept that this is a "writable ADL" is really curious though. 😅

Yeah, that thought's been coming to me for a while now, although it's still a bit nebulous.

If an ADL is a transformation applied to a "raw" `Node` in order to turn it into a new, "synthesized" `Node`, then `amend` does that as well. In fact, that's the reason this code is working with mostly surgical changes to the rest of the code. Each `Amender` implementation (`mapAmender`, `listAmender`, `linkAmender`, `anyAmender`) masquerades as a `Node` with all the accumulated updates applied.

Node contents are read via `Node.LookupBy*`, `Node.As*`, or iteration (for recursive nodes), so all I had to do was make sure that lookups and iteration incorporated the latest state of the node in a manner consistent with the underlying `Node` implementation. For example, `mapAmender` works seamlessly with `plainMap` because they're both essentially `LinkedHashMap`-based implementations (iterating over `plainMap` returns elements in their order of insertion).

What this also means is that `amend` will not work out-of-the-box with inconsistent "raw" data layouts. Using `mapAmender` with `go-ipld-adl-hamt` will "work", i.e. updates will be present in the "synthesized" `Node`, but encoded data will not always be in the right order.

In fact, I think `go-ipld-adl-hamt` is a halfway `Amender` already. It starts with an empty base `Node` and accumulates updates that it reports during lookups and iteration.

RangerMauve:
I was just off a call with the WNFS team and we were talking about how ADLs map from raw data into a complex Node, but how it wasn't obvious how the opposite could happen.

This made me think that mutability with ADLs is something we should talk about more. e.g. Should we have something in the raw datamodel for transforming a node? The stuff sketched up in #451 seems to allude to a need to be able to modify the keys in a Map or add/remove items in a List (and then get back a new node).

This also seems like an alternative to using a Builder node for interfacing with the data in a more high level way. Is this something we want within the repo or should we maybe look at having a higher level interface for IPLD?

rvagg:
> This also seems like an alternative to using a Builder node for interfacing with the data in a more high level way. Is this something we want within the repo or should we maybe look at having a higher level interface for IPLD?

Yes, I think that's fine. We have bindnode in here already and there's methods in the codegen that can mess with internals too (if they're local anyway: e.g. https://github.com/ipfs/go-ipld-git/blob/master/ipldsch_types.go#L61).

I think we want to be rigid with making sure that all the various layers and implementations can present both:

1. a proper `Node` interface to the data; and a
2. a `Assembler` `Builder` interface that works across the data structure concerned

As long as those things hold true then the stack should line up. But it doesn't preclude offering alternative interfaces to the same data for usability and/or performance purposes. Performance tends to constraint us a bit and has influenced some of the design decisions here. But I think experimentation with alternative interface mechanisms for the different levels is fine as long as we offer a coherent stack that can work transparently when a user/API doesn't know or care what's underneath.

rvagg:
So the next problem is: does this work with the layering that traversal needs to? I think the answer is no, unfortunately. If you look at the current implementation of `FocusedTransform()`, you'll see a `nb := n.Prototype().NewBuilder()`, with the traversal operating on that builder to make a new `Node`. With amend, the original prototype is ignored because it's operating on a raw datamodel form and it returns its own forms of `Node` objects instead. `WalkTransforming()` is similar but it uses the prototype when traversing the complex types, otherwise it just relies on the caller's `Node` to do the transforming.

So, I think if you were to run a traversal using this current implementation on anything but a datamodel layer (e.g. basicnode) `Node` then you might get into trouble. Some examples:

* A `TypedNode` which is a `struct Foo map { ... } representation tuple` - so as a `Node` it presents as if it's a map, but at the base it's really a list. When you pass that off to a codec then it'll get `Representation()` called on it so you encode the list.
* An ADL - take the `adl/rot13adl` demo ADL as a very simple example - it gets and sets strings encoded as ROT13, so the user just sees the unencoded strings but underneath they're encrypted. Pass it off to a codec and you should get the encrypted form.

In both cases, (and perhaps a third one where you layer one on the other) if you pass one of these to the current transform functions and do a simple transformation, it makes the changes through the `NodeAssembler` interface which takes care of making sure the underlying _thing_ gets updated properly. So you can transform a ROT13 ADL `Node` as if it were unencrypted but it'll still do the encrypting for you. You can mess with fields of a tuple-representation Map but it'll still be a List underneath; plus you get all of the other things that come along with a `TypedNode` like a strict list of fields that can be set for a `struct` and what types they can be.

Amend is hijacking that and returning an `Amender`, which can act like a `Node` but loses the layers. If I pass in a HAMT ADL and mess with some of the entries with a transform then I expect the operations to occur on the HAMT. But instead my HAMT is converted to an `Amender`, which has the HAMT underneath it but the transformations won't be occuring as they should via the `NodeAssembler` pattern. e.g. https://github.com/ipld/go-ipld-adl-hamt/blob/9004dbd839e0e73ec52d9b1b47ddbd0627d74b92/builder.go#L274-L285 is what happens when you add a new key to a HAMT - `Node#insertEntry` has all the ADL magic, including all of the operations to create and store new datamodel layer blocks in the LinkSystem.

I'm not sure exactly what to do about this though. Some kind of "commit" operation would be nice to add to the traversals but that would have to be implicit, leaving the user with no option to receive an `Amender` if they want one for the efficiency gains. But perhaps that's the key - allow the user to _opt in_. If it receives an `Amender` because the user made one with `NewAmender` then just work on that and return it. If not, then we'd need to _commit_ operations before returning, and those committals would need to be recursive and operate on the prototypes of the base `Node`s at each point.

But then there's the question of whether accumulating all of this in memory is even what the user wants? With the current form, I could pass in a _huge_ HAMT with strings in it and use the transformers to append some character to each entry and it'd update the nodes along the way, without needing to buffer it all in memory before doing it at once (one of the nice features of ADLs). So perhaps accumulating should be opt-in too, which means perhaps we should preserve some or all of the exsting traversal infrastructure and allow the user to opt-in to an amending traversal?

smrz2001:
> OK, I tried pulling this up into a `node/amend` package but have to concur with @smrz2001 that the interdependency is a bit too much. I started pulling some traversal pieces up into the root package, a `TraversalProgress` interface, the `TraversalFn` type and the `ErrBudgetExceeded` error but there's so much of `Progress` that's used inside amend that it essentially needs all of the fields exposed as functions on the interface, which is quite annoying.
> 
> So, if "where does this belong" from a purely API-sense perspective then I'd try and get it into `node` as its own package. But it's also true that this is essentially a form of transformation which, at the moment belongs under `traversal`.

Thanks for experimenting with this, @rvagg. This is where I end up each time I try to move `amend` out, and I've tried and given up a few times.

> So the next problem is: does this work with the layering that traversal needs to? I think the answer is no, unfortunately. If you look at the current implementation of FocusedTransform(), you'll see a nb := n.Prototype().NewBuilder(), with the traversal operating on that builder to make a new Node. With amend, the original prototype is ignored because it's operating on a raw datamodel form and it returns its own forms of Node objects instead. WalkTransforming() is similar but it uses the prototype when traversing the complex types, otherwise it just relies on the caller's Node to do the transforming.

Agreed, this is what I was trying to allude to in my [earlier](https://github.com/ipld/go-ipld-prime/pull/445#issuecomment-1201870607) comment:
> Node contents are read via Node.LookupBy*, Node.As*, or iteration (for recursive nodes), so all I had to do was make sure that lookups and iteration incorporated the latest state of the node in a manner consistent with the underlying Node implementation. For example, mapAmender works seamlessly with plainMap because they're both essentially LinkedHashMap-based implementations (iterating over plainMap returns elements in their order of insertion).
>
> What this also means is that amend will not work out-of-the-box with inconsistent "raw" data layouts. Using mapAmender with go-ipld-adl-hamt will "work", i.e. updates will be present in the "synthesized" Node, but encoded data will not always be in the right order.

Any amendment logic _must_ take into account the underlying `Node` implementation, e.g. the current `mapAmender` will work with `plainMap` but cannot be assumed to work with any other "map-type" `Node` (e.g. HAMT).

I have a few ideas on how to move forward, let me know what you (and @RangerMauve) think:
- Revert the integration of `amend` and `focus`. I feel like the current code successfully validates that performant traversal and modification is _possible_ using `amend`, but also that `traversal` doesn't _have_ to be `amend`-ified as long as alternative lookup/modification mechanisms are available.
- `amend` becomes its own package inside `node` with its own (limited) version of `Progress` and `Budget` because those are definitely useful concepts. Though it'll be inside `node`, it'll be an alternative to `traversal` that can opted-in to that is also closer to the data model, allowing for more performant operations.
- `Amender` implementations are explicitly tied to the underlying `Node` implementation, e.g. `mapAmender` should be tied to `plainMap` (maybe even renamed to `plainMapAmender`?), HAMT will have it's own `Amender` implementation, etc. This coupling could perhaps be implemented via an updated version of this existing interface:
```
// NodePrototypeSupportingAmend is a feature-detection interface that can be
// used on a NodePrototype to see if it's possible to build new nodes of this style
// while sharing some internal data in a copy-on-write way.
//
// For example, Nodes using an Advanced Data Layout will typically
// support this behavior, and since ADLs are often used for handling large
// volumes of data, detecting and using this feature can result in significant
// performance savings.
type NodePrototypeSupportingAmend interface {
	AmendingBuilder(base Node) NodeBuilder ======> AmendingBuilder(base Node) Amender
	// FUTURE: probably also needs a `AmendingWithout(base Node, filter func(k,v) bool) NodeBuilder`, or similar.
	//  ("deletion" based APIs are also possible but both more complicated in interfaces added, and prone to accidentally quadratic usage.)
	// FUTURE: there should be some stdlib `Copy` (?) methods that automatically look for this feature, and fallback if absent.
	//  Might include a wide range of point `Transform`, etc, methods.
	// FUTURE: consider putting this (and others like it) in a `feature` package, if there begin to be enough of them and docs get crowded.
}
```
`traversal` can then possibly check this feature flag and have an alternative path that takes advantage of any performance gains, but can also fallback to the more generic implementation if no `Amender` is available.
- `Amender` takes its [original](https://github.com/ipld/go-ipld-prime/commit/24469a6bfbc3de80b8558d68faa0b388c2188f82) form with `Add`, `Remove`, `Update` methods instead of `Transform`. The latter makes things extra complicated because we don't know the nature of the operation until _after_ the transformation is applied (i.e. whether it was a deletion, modification, or addition).

smrz2001:
> But then there's the question of whether accumulating all of this in memory is even what the user wants? With the current form, I could pass in a _huge_ HAMT with strings in it and use the transformers to append some character to each entry and it'd update the nodes along the way, without needing to buffer it all in memory before doing it at once (one of the nice features of ADLs). So perhaps accumulating should be opt-in too, which means perhaps we should preserve some or all of the exsting traversal infrastructure and allow the user to opt-in to an amending traversal?

With the approach I outlined above, updating nodes along the way could perhaps be implemented via `hamtAmender` specific configuration.

`Amender` would establish an API for modifying nodes, with reference implementations for existing complex types like `plainMap`, `plainList`, and `plainLink`. ADLs would be able to choose if/how their `Amender` should be implemented, and also be able to use it with `traversal`.

rvagg:
Yes! Mostly that all sounds good. Some thoughts and questions:

* Instead of `plain*` in the language, maybe frame this as "basicnode" since that's become the common parlance for how we refer to these base level things, the basicnode `Node` implementations happen to be implemented as `plain*` but that's really just internal naming. What you're suggesting is building basicnode-level tooling. I'm not really sure what that means in API terms, though. I think you could put some amend-related interfaces in the root of `datamodel` and then do some stuff in a sub-package, or even consider working inside the basicnode package? Or basicnode/amend? Is that too much? I'll let you think on this and come up with ideas, I don't have solid ones at the moment.
* You're getting quite close to the `NodeBuilder` interface here, and as you've suggested already you're almost (or maybe are) building an ADL. So maybe its worth considering whether an `Amender` interface should just extend `NodeBuilder`? If that's possible it would provide some really interesting API possibilities when passing things around, and amender nodes can return proper prototypes that divert building code to use new amenders too. Since the intention is to have consistent build interfaces to a diversity of implementations, maybe we can just go the full way with that.
* We _could_ consider extending all of the basicnode Builders (e.g. `plainMap__Builder`) from an amender by default. So .. for example, if I decoded a graph directly from some bytes, without a typed interface or an ADL in the way, just their basicnode components, then I end up with a graph that's got amenders underneath and I automatically get the amend behaviour when using traversal on it. We could even work up the stack and give the same treatment to the `TypedNode` codegen components. Just a thought, and I'm not sure whether it'd actually be worth it or not, or what user expectations might be when `Build()` is called (are they OK with a change-tracked object? we're already dealing with memory issues up through the go-graphsync->go-data-transfer->go-fil-markets stack because of stacking `Node` problems)
* Can you tell me about `trackProgress` and what it's intended to do? I can't quite figure it out from the code and maybe it's not fully doing what it's intended to do?
* Likewise, `created` doesn't ever seem to be set, it's only ever false from what I can see. Can you talk about the intention there? Or am I looking at it wrong?

smrz2001:
Thanks for the excellent suggestions, @rvagg! 🙏🏼

> * Instead of `plain*` in the language, maybe frame this as "basicnode" since that's become the common parlance for how we refer to these base level things, the basicnode `Node` implementations happen to be implemented as `plain*` but that's really just internal naming. What you're suggesting is building basicnode-level tooling. I'm not really sure what that means in API terms, though. I think you could put some amend-related interfaces in the root of `datamodel` and then do some stuff in a sub-package, or even consider working inside the basicnode package? Or basicnode/amend? Is that too much? I'll let you think on this and come up with ideas, I don't have solid ones at the moment.

I really like the idea of putting `amend`-related interfaces in the root of `datamodel` then working inside `basicnode/amend`, if I can make that work without any dependency issues. I'd love to keep `Amender` implementations as close as possible to their respective raw layouts, so somewhere inside `basicnode` would be perfect.

> * Can you tell me about `trackProgress` and what it's intended to do? I can't quite figure it out from the code and maybe it's not fully doing what it's intended to do?

`trackProgress` is a vestige of the integration with `focus`, which allows _not_ storing metadata while traversing a graph for `Get` operations, presumably to save memory when traversing large graphs. Currently, it's only being passed from `focus`:
```
func (prog Progress) Focus(n datamodel.Node, p datamodel.Path, fn VisitFn) error {
	n, err := prog.get(n, p, true)
	if err != nil {
		return err
	}
	return fn(prog, n)
}
```
```
func (prog Progress) Get(n datamodel.Node, p datamodel.Path) (datamodel.Node, error) {
	return prog.get(n, p, false)
}
```
```
func (prog *Progress) get(n datamodel.Node, p datamodel.Path, trackProgress bool) (datamodel.Node, error) {
	prog.init()
	return NewAmender(n).Get(prog, p, trackProgress)
}
```
Here's how it's used in `mapAmender` (and similarly in the other `Amender`s):
```
func (a *mapAmender) storeChildAmender(seg datamodel.PathSegment, n datamodel.Node, k datamodel.Kind, create bool, trackProgress bool) Amender {
	childAmender := a.opts.newAmender(n, a, k, create)
	if trackProgress {
		a.mods.Put(seg, childAmender)
	}
	return childAmender
}
```
`mods`, which stores the `mapAmender` state, doesn't get updated if `trackProgress = false`.

> * Likewise, `created` doesn't ever seem to be set, it's only ever false from what I can see. Can you talk about the intention there? Or am I looking at it wrong?

Sorry about that 😬 - I was being pedantic and named the variable controlling `created` to the verb `create`. Here's an example from `mapAmender` of how it's set:
```
			// While building the nested amender tree, only count nodes as "added" when they didn't exist and had to be
			// created to fill out the hierarchy.
			create := false
			if childVal == nil {
				a.adds++
				create = true
			}
			a.storeChildAmender(childSeg, newChildVal, newChildVal.Kind(), create, true)
```
`mapAmender.storeChildAmender` will pass this value along to `AmendOptions.newAmender` that will return the appropriate `Amender` with `created` set correctly. This is because I need to know if a child in the graph was present in the base `Node` or "added" to the new one. Currently, this matters more in `mapAmender` because of the way its state is maintained, and is checked via `isCreated`.
```
				// If the child node being removed is a new node previously added to the node hierarchy, decrement
				// `adds`, otherwise increment `rems`. This allows us to retain knowledge about the "history" of the
				// base hierarchy.
				if amd, castOk := childVal.(Amender); castOk && amd.isCreated() {
					a.adds--
				} else {
					a.rems++
				}
```
If this feels a little circuitous, that's because it is lol. If there's a less complicated way to make this work, I'd love to hear your thoughts.

smrz2001:
> * You're getting quite close to the `NodeBuilder` interface here, and as you've suggested already you're almost (or maybe are) building an ADL. So maybe its worth considering whether an `Amender` interface should just extend `NodeBuilder`? If that's possible it would provide some really interesting API possibilities when passing things around, and amender nodes can return proper prototypes that divert building code to use new amenders too. Since the intention is to have consistent build interfaces to a diversity of implementations, maybe we can just go the full way with that.
> * We _could_ consider extending all of the basicnode Builders (e.g. `plainMap__Builder`) from an amender by default. So .. for example, if I decoded a graph directly from some bytes, without a typed interface or an ADL in the way, just their basicnode components, then I end up with a graph that's got amenders underneath and I automatically get the amend behaviour when using traversal on it. We could even work up the stack and give the same treatment to the `TypedNode` codegen components. Just a thought, and I'm not sure whether it'd actually be worth it or not, or what user expectations might be when `Build()` is called (are they OK with a change-tracked object? we're already dealing with memory issues up through the go-graphsync->go-data-transfer->go-fil-markets stack because of stacking `Node` problems)

I'd love to more deeply integrate `Amender` into `basicnode` this way - `Amender` extends `NodeBuilder` and `basicnode` `Builder`s extend `Amender`. Will play around with this and report back.

I've also been worried about memory issues or broken invariants when using mutable `Node`s. Would like to keep chatting about better ways to integrate (or whether we just shouldn't).

rvagg:
ok, I grok `created` now, it was the naming that was throwing me and I see now how you're using it to modify behaviour

re integration - I don't know if it's worth getting too hung up if you can't find a clear path forward, maybe just landing _something_ is better than circling around endlessly. But, I have a vague picture of this being a useful tool that is composable in the stack of other composable things. My mind wants to reach for odd examples like: I have a bunch of blocks to decode, they're all just plain maps `{String:String}` and I want to mush all of them together into a single map, removing duplicates. So I use an Amender and pass that (somehow) as my prototype to `DecodeUsingPrototype()`; or perhaps I have to use the codec directly and call `dagcbor.Decode(amender, reader)`—basically I want a `NodeAssembler` at this point but I want it to be assembling on a _pre existing_ `Node`. Now, I think I can already do this with a basicnode assembler, which I can mess with in different ways until I call `Build()`, but perhaps this is something I don't want a finality to, I want to be able to read from it while also appending to it. Maybe I have a stream of blocks coming in that are adjusting my stored Map which somehow updates current state, so I just keep decoding into the same `NodeAssembler` but that's attached to a live `Node` that I can use to read from.

Basically, I'm thinking of this as a potential generic read/write `Node` with some nice efficiency properties. Maybe that's too much of a reach and maybe the usecases aren't clear enough, but it seems like an interesting way to push the stack and layering forward? Certainly IPLD Patch is in there as a usecase, but we come back to the original problem that IPLD Patch should operate on an arbitrary layer (datamodel, typed, ADL), it's the interfaces that matter to make it work.

smrz2001:
> ok, I grok `created` now, it was the naming that was throwing me and I see now how you're using it to modify behaviour

Cool, I'll also rename the `create` variable to `created` to make the code more readable 😅

> re integration - I don't know if it's worth getting too hung up if you can't find a clear path forward, maybe just landing _something_ is better than circling around endlessly. But, I have a vague picture of this being a useful tool that is composable in the stack of other composable things. My mind wants to reach for odd examples like: I have a bunch of blocks to decode, they're all just plain maps `{String:String}` and I want to mush all of them together into a single map, removing duplicates. So I use an Amender and pass that (somehow) as my prototype to `DecodeUsingPrototype()`; or perhaps I have to use the codec directly and call `dagcbor.Decode(amender, reader)`—basically I want a `NodeAssembler` at this point but I want it to be assembling on a _pre existing_ `Node`. Now, I think I can already do this with a basicnode assembler, which I can mess with in different ways until I call `Build()`, but perhaps this is something I don't want a finality to, I want to be able to read from it while also appending to it. Maybe I have a stream of blocks coming in that are adjusting my stored Map which somehow updates current state, so I just keep decoding into the same `NodeAssembler` but that's attached to a live `Node` that I can use to read from.
> 
> Basically, I'm thinking of this as a potential generic read/write `Node` with some nice efficiency properties. Maybe that's too much of a reach and maybe the usecases aren't clear enough, but it seems like an interesting way to push the stack and layering forward?

No, absolutely, this is a great direction, IMO 🙌🏼 A "generic read/write `Node` with some nice efficiency properties" was what I vaguely had in mind as a long-term objective when I started this activity, just didn't quite understand what it would look like and where it would fit until our recent discussions.

In fact, I was also considering how such a `Node` implementation could further be integrated into the codecs and be _truly_ copy-on-write by storing start/stop offsets/pointers (or offset/pointer + content length) from the serialized bytes instead of always copying data out. For content that never gets explicitly retrieved, the offsets are enough until `Encode()` is called, at which point data gets copied from the original byte stream.

Yet another optimization could be to not decode anything at all, just plug the bytes into the `Node`. If/when data is requested, the `Node` _then_ unmarshals the bytes, but only enough to retrieve what was requested. At `Encode`, all unread blobs are copied directly into the serialized form.

Just a thought - there are definitely more efficiencies we could achieve but they might not provide enough ROI. I'd be up for digging deeper though, once the initial implementation has solidified.

One more thought (and I'm just thinking aloud here) - regarding finality, we could perhaps give applications a choice to `Build` or `BuildFinal` (or `BuildTracked` / `Build`), where the r/w `Node` (and probably recursively every r/w child `Node`) gets marked "final" and can no longer be updated. That way, if an application knows that the `Node` mustn't be updated after a certain stage, we can prevent broken invariants by actually locking the state of the `Node`. Maybe all `Build`s are "final" unless explicitly stated otherwise...

>Certainly IPLD Patch is in there as a usecase, but we come back to the original problem that IPLD Patch should operate on an arbitrary layer (datamodel, typed, ADL), it's the interfaces that matter to make it work.

Yes, agreed 👍🏼

rvagg:
> Yet another optimization could be to not decode anything at all, just plug the bytes into the `Node`.

Yes, maybe. This kind of optimisation gets brought up a relatively frequently, but I've not seen concrete suggestions for where it might provide gains. Data elements in flexible codecs, mainly dag-cbor, tend to be quite small anyway so decoding them isn't too costly and setting up deferred decoding is going to add unnecessary cost. The most common large chunks of data coming out of IPLD codecs is UnixFS file data, which we see coming up through dag-pb in the `Data` element of a block as Bytes and then that's passed through UnixFS (our modern stack does it as an ADL) which further decomposes it into file/directory properties/metadata _and_ file data. But each step along that way uses subslicing of the original bytes. i.e.:

* https://github.com/ipld/go-codec-dagpb/blob/05fe10cbb5d30b4ec93c6000fa7f5fe48a7b27d3/unmarshal.go#L77 ->
* https://github.com/protocolbuffers/protobuf-go/blob/v1.28.1/encoding/protowire/wire.go#L440 ->
* https://github.com/ipfs/go-unixfsnode/blob/94986a76305a01f1f21eb86fb1ed41e5478367f2/file/wrapped.go#L10-L19 ->
* https://github.com/ipfs/go-unixfsnode/blob/94986a76305a01f1f21eb86fb1ed41e5478367f2/data/unmarshal.go#L25

So even though we have a bunch of layers, we're just progressively slicing off smaller pieces of the input `[]byte` from the codec, I don't believe there's any copying involved here.

There are other cases of large chunks of data, but I think they mostly come through as Bytes and I think we end up doing this same pattern each time unless we need to transform it.

One area where we are having real performance concerns is when we stack up layers in a way that's not up-front composable. This is becoming increasingly important for us to deal with because it's causing memory bloating in the Filecoin data transfer stack. Check out the writeup @ https://github.com/ipld/go-ipld-prime/issues/443 on that one. I don't know if anything in here is going to help over there, but that's a case of having too many pieces floating around in memory because we had to plug one thing into another into another just to transform it into the end state, but we still have all of those intermediate pieces hanging around. Probably a distraction from here, but since your head is roughly in this space perhaps it's interesting to think about.

smrz2001:
> > Yet another optimization could be to not decode anything at all, just plug the bytes into the `Node`.
> 
> Yes, maybe. This kind of optimisation gets brought up a relatively frequently, but I've not seen concrete suggestions for where it might provide gains. Data elements in flexible codecs, mainly dag-cbor, tend to be quite small anyway so decoding them isn't too costly and setting up deferred decoding is going to add unnecessary cost. The most common large chunks of data coming out of IPLD codecs is UnixFS file data, which we see coming up through dag-pb in the `Data` element of a block as Bytes and then that's passed through UnixFS (our modern stack does it as an ADL) which further decomposes it into file/directory properties/metadata _and_ file data. But each step along that way uses subslicing of the original bytes. i.e.:
> 
> * https://github.com/ipld/go-codec-dagpb/blob/05fe10cbb5d30b4ec93c6000fa7f5fe48a7b27d3/unmarshal.go#L77 ->
> * https://github.com/protocolbuffers/protobuf-go/blob/v1.28.1/encoding/protowire/wire.go#L440 ->
> * https://github.com/ipfs/go-unixfsnode/blob/94986a76305a01f1f21eb86fb1ed41e5478367f2/file/wrapped.go#L10-L19 ->
> * https://github.com/ipfs/go-unixfsnode/blob/94986a76305a01f1f21eb86fb1ed41e5478367f2/data/unmarshal.go#L25
> 
> So even though we have a bunch of layers, we're just progressively slicing off smaller pieces of the input `[]byte` from the codec, I don't believe there's any copying involved here.
> 
> There are other cases of large chunks of data, but I think they mostly come through as Bytes and I think we end up doing this same pattern each time unless we need to transform it.

That's a great pattern!! I can see how it eliminates the need for copying at each step.

I've built a TLV (tag-length-value, like protobuf) protocol parser in the past that required random access to elements at any point in the buffer, and, in that case, it made sense to use deferred retrieval/copy. Until such a use case is needed with large chunks of data where the layering pattern above *cannot* be used, I agree it doesn't seem worth investing in this type of optimization.

> One area where we are having real performance concerns is when we stack up layers in a way that's not up-front composable. This is becoming increasingly important for us to deal with because it's causing memory bloating in the Filecoin data transfer stack. Check out the writeup @ #443 on that one. I don't know if anything in here is going to help over there, but that's a case of having too many pieces floating around in memory because we had to plug one thing into another into another just to transform it into the end state, but we still have all of those intermediate pieces hanging around. Probably a distraction from here, but since your head is roughly in this space perhaps it's interesting to think about.

I couldn't fully follow the `bindnode` parts of the discussion, but I strongly feel like the interface between layers can be optimized (e.g. not constantly copying data into new `basicnode` objects when going from one layer to another, maybe by using references to existing objects, or creating new nodes where their data is actually shared instead of duplicated).

> **Ideally**, when the `codec/dagcbor` is pushing out nodes into a builder, that builder isn't relying on `basicnode` at all, or it's doing so minimally, and that the nodes are landing directly into their final destination Go structs fronted by `bindnode`.

To this point from your comment in that thread, and I'm really thinking aloud here, perhaps _this_ is actually a good use case for deferred decoding? As a generic example, reifying ADLs today results in a full `Node` being built that holds the whole graph in memory. Composing this `Node` in another ADL will result it yet another full graph being built at reification and held in memory, and so on. If even smaller pieces of information weren't actually decoded from the byte stream, and only metadata about them passed forward through the various transformations until the last one, where they were actually decoded, then each intermediate copy of that data could be eliminated. In other words, it isn't the size of data that justifies deferred decoding but instead avoiding multiple copies of small pieces of data hanging around.

(I'm super out-of-my-depth here so please ignore if this makes no sense 😅)

PS: Wasn't a distraction, was a very fascinating discussion for me!
PPS: I've got a couple of other tasks on my plate to wind up but am making slow yet steady progress on the `basicnode`, `amend` items we discussed above. I won't have any update for today's call (and probably need to skip it), but should have updates at the next one!

rvagg:
> a good use case for deferred decoding

yes .. but pulling that off with our current stack is tricky; and yet it seems increasingly desireable - there's even now a discussion about pushing past the current block size limits (hard, soft, or otherwise) to larger blocks, and what this might mean for our stack. Currently for many of the IPLD pieces it's going to be a lot of pain - even though we do a lot of streaming operations, instantiation of data model layer nodes does currently need it all to be slurped up and reified. Maybe that's not so great in some circumstances. It's not something we've been forced to deal with directly but I think we're edging towards needing to adapt to memory constraints on a number of fronts!

smrz2001:
Sorry for the delay, @rvagg. Had a big release coming up that took most of my spare time the last month.

I've started down the path of `NodeAmender` extends `NodeBuilder` -> `plainMap_Builder` implements `NodeAmender` (and so on for the other `basicnode` types). I'm liking where this is going so far but let's see. Please feel free to point out anything that seems obviously non-ideal.

I decided to do a reset of the code and implement `Amender` afresh using the learnings from the previous version. Please note that the new version is draft code and quite rough at the moment. I'm going to flesh it out, clean it up, add more comments, etc. soon.

rvagg:
Looks interesting!

The main thing that occurs to me is that since you're touching some existing base functionality then you need to be very sensitive to performance and memory consumption as any change is likely going to have widespread impact since these basicnode builders are used a lot. If you're adding additional properties to them then we need to be thinking about memory footprint. If you're adding extra code to existing paths then we need to be sensitive to increased costs for the common cases.

smrz2001:
> Looks interesting!
> 
> The main thing that occurs to me is that since you're touching some existing base functionality then you need to be very sensitive to performance and memory consumption as any change is likely going to have widespread impact since these basicnode builders are used a lot. If you're adding additional properties to them then we need to be thinking about memory footprint. If you're adding extra code to existing paths then we need to be sensitive to increased costs for the common cases.

Thanks for the note, @rvagg, I'll keep that in mind 🙏🏼

FYI, I'm still making progress, just slowly. It's definitely trickier to make such intrusive changes to the builder code, especially with the stricter performance requirements, but also because I'm having to build a deeper understanding of some core IPLD concepts. I'm done with my first draft of changes to `plainMap`, now working on `plainList`.

I was able to take fairly big liberties with the initial version of `amend` that I can't this time but, so far, it feels like I haven't run into any major roadblocks, so I'll keep working in this direction. Should I run into any insurmountable obstacles, we can chat and re-evaluate how to proceed.

smrz2001:
Hey @rvagg, after some thought, I was able to significantly simplify my implementation by short-circuiting a bunch of the original `amend` ideas. Calls to `NodeAmender` will actually mutate the underlying `Node`, and not just return a mutated copy. It seemed reasonable to assume that anyone using a `NodeAmender` is explicitly signing up to modify the `Node` they're passing in.

If this is a good approach, the next hurdle (that I would appreciate some input on) is to figure out how to allow `Link`s to be loaded during traversal from within `basicnode`. I don't want to muck with the layering too much, but also cannot get traversal to work properly without passing through some link-related config. Once this last piece is done, `traversal.Focus*` can be simplified as before but this time with node transformation built into `basicnode`.

These latest updates allowed me to eliminate a lot of the code I had carried over from the previous implementation and all but a couple of map test cases are passing. I'll add some amend-specific test cases once we get closer to a final approach.

smrz2001:
Ok, so these changes are not complete and not fully working, but I whipped up a proof-of-concept for what I'm thinking so I could get your input, @rvagg, and see if something like this could work once fleshed out. Let me know what you think!

PS: If this stuff feels like it's going off into the weeds, please let me know 😄 I'd hate to waste your time lol.

smrz2001:
Alternatively, the original approach was feature complete and unit/performance tested. If it makes sense to put in as an efficient and _much_ less intrusive option for modifying nodes, I wouldn't mind switching back. Just need to make sure that we can figure out a good home for the code, and that the API is clear.

The `basicnode` modifications make more long-term sense but they're definitely intrusive.

smrz2001:
Hey @rvagg, just wanted to see if it makes sense to continue work on this PR. I'd love to take it over the finish line but a lot has changed on the PL side, and I don't want to create work that might not be essential.

rvagg:
Hey @smrz2001 I finally had a chance to look over this! I'm so sorry for being slow on this, I know you've put a lot of effort into it and the level of the attention I've been giving to it doesn't match the amount of work, nor the potential benefits from this work.

Anyway, I think this approach has excellent merit and if you're inclined to get it over the line we should try and proceed. I the current `NodeAmender` interface and it'd be interesting explore what it would look like for an ADL (top of mind is what it would look like used in https://github.com/ipld/go-ipld-adl-hamt).

The use in the traversal package looks reasonable, and you're touching some of the less used parts of that API so the churn concern is quite minimal there and I can imagine it has some great perf benefits used there anyway.

The main bits that will be of concern getting this over the line is the use inside basicnode's `plainMap`, `plainList`, and `plainLink` since those are very hot pieces of code. We'd want to make sure that typical codepaths that use these aren't impacted, or are impacted minimally. The accessor methods are the main concern but I _think_ there's just a new layer of indirection via `Build()` which should result in what's pretty much a noop, so I guess it's not going to be very measurable that there's a new layer in there. But that'd be worth looking at to make sure the impact is minimal.

I think next step, if you want to continue with this, would be to expand on the docs in `datamodel/amender.go` (and maybe other places that might be introducing new things) so we can expand the conversation about this to a couple more people without requiring them to grok everything that's going on. Also it would be good to get the assertions that are currently commented out in `traversal` cleared up so it doesn't look like this is a breaking change—or if it is, that the breakage is minimal. Benchmarks to make the case would be helpful too of course, especially for the current common code paths to show that accessors on the basicnode types is minimally impacted.

smrz2001:
Hey @rvagg, thanks for taking a look! I'm glad this code is still useful/relevant 😊 I'd definitely love to get it over the line.

>I the current `NodeAmender` interface and it'd be interesting explore what it would look like for an ADL (top of mind is what it would look like used in [ipld/go-ipld-adl-hamt](https://github.com/ipld/go-ipld-adl-hamt)).
> 
Yes!!

> The main bits that will be of concern getting this over the line is the use inside basicnode's `plainMap`, `plainList`, and `plainLink` since those are very hot pieces of code. We'd want to make sure that typical codepaths that use these aren't impacted, or are impacted minimally. The accessor methods are the main concern but I _think_ there's just a new layer of indirection via `Build()` which should result in what's pretty much a noop, so I guess it's not going to be very measurable that there's a new layer in there. But that'd be worth looking at to make sure the impact is minimal.
> 
Yes, agreed. With this version of the code, the typical codepaths should remain minimally affected but benchmarking will confirm.

> I think next step, if you want to continue with this, would be to expand on the docs in `datamodel/amender.go` (and maybe other places that might be introducing new things) so we can expand the conversation about this to a couple more people without requiring them to grok everything that's going on.
>
Yep, I can do that. I'll start cleaning up the code, and adding more documentation and tests.

> Also it would be good to get the assertions that are currently commented out in `traversal` cleared up so it doesn't look like this is a breaking change—or if it is, that the breakage is minimal.
>
I actually needed a bit of input regarding those commented assertions.

They're all related to the state of a traversal in `traversal.Progress`, which is used to configure and analyze a traversal. It did not make sense to me to add `Progress` to `basicnode`, at least not in the form it's in today, which is why the assertions checking the traversal state are failing when using `basicnode` for amendment. The assertions regarding `Node` contents are all passing.

Do we want something `Progress`-like in `basicnode`? Or, can the `traversal` APIs be simplified to not have/use `Progress`? That would be a big change to them though.

Alternatively, we could leave `traversal` as-is, without `amend`, especially since you mentioned those are less-used parts of the API. This latter option would be my preference unless you have other ideas about how to do this that are not too complicated or intrusive.

> Benchmarks to make the case would be helpful too of course, especially for the current common code paths to show that accessors on the basicnode types is minimally impacted.
>
Yes, I'll collect some benchmarks and post them here.

rvagg:
> I actually needed a bit of input regarding those commented assertions.

Ahh, so `focusedTransform` is no longer being used and you're offloading all of the internal traversal to the amender which doesn't have the recursive state tracking, so you can't provide the same feedback.

This is tricky and it might be something to get feedback on from people who might use this, I'm thinking specifically of the ADL usage, Will & Masih might be good for that when we get there. My inclination is to say that it we can drop the extra info for a transform, it seems less useful than for a selector-based traversal, but even there we're already finding it (a) only occasionally useful and (b) sometimes even quite wasteful and a problem for performance. Specifically check out https://github.com/ipld/go-ipld-prime/issues/518 and note the bit about IPNI performance concerns—updating the `Progress`'s path information has allocation costs that are being felt for a use case that just doesn't need it.

rvagg:
> Should we consider dropping the state tracking entirely? Is it useful enough that removing it would prevent any common use-cases from being satisfied?
> 
> traversal.Focus also has the notion of trackProgress. Adding something like that to traversal.Walk might be another way to limit the amount of state tracking.

Yes, I've been thinking that maybe a bool in `Config` to turn on and off progress. The complication is `StartAtPath` and `Budget` which need some amount of state. Maybe it could default to `false` but turn on when those features are needed, or when it's explicitly turned to `true`.

smrz2001:
> > traversal.Focus also has the notion of trackProgress. Adding something like that to traversal.Walk might be another way to limit the amount of state tracking.
> 
> Yes, I've been thinking that maybe a bool in `Config` to turn on and off progress. The complication is `StartAtPath` and `Budget` which need some amount of state. Maybe it could default to `false` but turn on when those features are needed, or when it's explicitly turned to `true`.

From what I can tell, `trackProgress` in `Progress.Get` is mainly used to control updates to `Progress.LastBlock` and `Progress.Path` while allowing `StartAtPath` and `Budget` to always be used/updated. Maybe that's the way to do it elsewhere since `Progress.Path` is presumably the main source of allocations during traversals?

smrz2001:
It had been a while since I'd looked back at my own code, and now that I have, I've realized that the changes to `focus` are incomplete. Even in the `basicnode` implementations, `NodeAmender` operations assume that children are one of the ("amendable") `basicnode` types, which is definitely not always going to be the case.

These changes were primarily meant to discuss and validate the current high-level approach, compared with the previous one where `amend` was at the level of `traversal` and `patch`, and not deep inside `basicnode`. Since you've had a chance to review the code, @rvagg, and feel the approach has enough merit to keep exploring its possibilities further, I'll work on fleshing out the implementation (by feature-detecting `NodePrototypeSupportingAmend`), and to see if/how it can then be integrated into existing flows cleanly.

Before that though, if we take a step back, does adding to a `Node` implementation the ability to recursively operate on its children seem like a layering violation? I think it is sufficient for `plainMap` or `plainList` to allow the easy addition/removal/replacement of immediate children but allowing operations on children at or beyond 2 levels of nesting seems like it doesn't belong in `map` or `list`-like objects.

`NodeAmender` could be simplified to just having a `Transform` function that can update the value of any type of `Node` (including replacing the root of a recursive `Node`) or update the immediate children of recursive `Node`s.

What do you think?

I was in the process of adding more `NodeAmender` documentation but then paused and thought I'd get a little more clarity here first.

smrz2001:
Hey @rvagg, I just pushed some reworked code that simplifies things considerably. All tests are passing (except on Windows for some reason haha), so there shouldn't be any regressions in the code, at least not any major ones.

There aren't any new test cases for `NodeAmender.Transform` or any benchmarking results, but I will look into both soon. I did remove two asserts from the `map_test.go` that don't quite apply in the new approach.

I'm going to restart the changes to `focus` and `walk` with this new foundation, which treats `plainMap` and `plainList` as amendable recursive types whose roots or immediate children can be transformed (added, modified, deleted).

Both `focus` and `walk` will remain complicated pieces of code (soon with additional amend feature checks) because not all node implementations support amendment, but we might be able to optimize the code considerably for the types that do.

I feel like recursive types have always been the source of complexity and inefficiency when attempting to modify nodes. The new code makes it easier to modify the basic map/list types while also improving the DX a little bit, i.e. calling `Transform` to modify nodes in-place instead of using `BeginMap`/`BeginList` -> `Finish` to create new nodes with transformed contents.

smrz2001:
Just pushed a version of `walk` with amenders. All existing tests are passing on MacOS, I'm surprised they're failing like they are in CI 😕

I'll work on adding some more tests and benchmarks soon, though would love some feedback regarding the approach if you get a chance to take a look, @rvagg.

smrz2001:
I went ahead and pulled in my map/list interface changes as well here, IMO further improving the DX for the amend API. I feel like now there's probably enough of an API around this to have a fruitful discussion on the path forward, @rvagg.

Since the API will very likely evolve some more before it settles, I feel like I should incorporate feedback before adding more unit/benchmarking tests.

`*Amender` remains an optional, no/low-cost abstraction over `NodeBuilder`, so I don't foresee a huge hit in performance, especially for critical paths. I will of course get some numbers and add correctness tests once we're further along.

rvagg:
This is looking pretty clean, I like the way the traversal code feature detects and splits that out; much less scary than having it all tangled up.

Could you work on adding some basic tests to the map and/or list amenders to demonstrate usage? I feel like that would be a good path to describing what the purpose of this all is.

smrz2001:
Just pushed some tests demonstrating usage for `MapAmender`. The logic for `ListAmender` is trickier to keep optimized, so I'm still working on that and its tests.

Will also figure out why the other tests are failing.

smrz2001:
Ok, just pushed some fixes and tests for `ListAmender` demonstrating its usage. Now there are tests for both `MapAmender` and `ListAmender`. Please let me know what you think when you get a chance to look at the latest code, @rvagg.

It's a little bizarre that all the tests are passing for macOS but not for Windows or Ubuntu for Go 1.19.x/1.20.x, though they pass for Ubuntu on Go 1.18.x. I'll see if I can figure out what's going on - is this something you've run into before?

The Go 1.17.x tests on Ubuntu are failing because I've included an experimental package for some list operations that I did not want to code by hand, though I can implement them to maintain compatibility with older Go versions.

smrz2001:
Ok, so the tests were failing because transformations are now changing underlying nodes by default.

This is probably a good stage to discuss how to deal with im/mutability when transforming nodes. As I see it, we have a couple of options.

A.

By default, the code in its current state will now transform nodes being amended, even after they have been "built" and are presumably immutable. This is the most space/performance efficient because it needs no additional metadata regarding modifications. However, this _could_ lead to broken invariants in other parts of the code assuming any constructed nodes to remain unchanged.

Maybe that's ok because this only happens if a user explicitly requests an amending builder for a node and knows what to expect. If they need a base node to remain immutable, they'll need to use `datamodel.Copy` to create a deep copy that can then be modified. This loses some of the space advantage of this approach but does retain the better DX of amending builders and allows unlimited modifications of the copied node.

B.

I can reuse some of the logic I previously wrote to maintain additional state for modifications while keeping the base node unmodified. This will definitely increase the complexity of the code and have an impact on space/performance because of the need to store metadata regarding modifications separately from the base node. This approach would have the advantage of ensuring that base nodes are never modified from under code expecting them to remain immutable.

This approach would be cleaner but I'm not sure how heavily to weigh that against the significant complexity/space/performance costs.

smrz2001:
Hey @rvagg, have you had a chance to look at some of the latest updates? There are now examples of how to use the new interfaces.

I think we're definitely closer to a clean API though there are still some usability details to iron out, some of which will have an impact on performance.

It would be good to understand the most common ways in which this code might be used in order to choose the right approach, or add some configuration options with sane defaults.

Issue #259, opened 2021-09-27
rename and deprecate the multicodec package


mvdan:
To avoid confusing conflict with https://github.com/multiformats/go-multicodec, which came first, and really can't have another name than "multicodec". Whereas the multicodec package is a registry, so we can call it multiregistry or codecregistry.

mvdan:
One more detail: we shouldn't simply move the package, as that would break downstreams immediately. We can move the package and add a "forwarding" package in the old location, forwarding the APIs, and being marked as deprecated in favor of the new one.

rvagg:
turns out to be too disruptive; I'm going to say this isn't worth it, see also #467

rvagg:
Reopened based on feedback @ https://github.com/ipld/go-ipld-prime/pull/467#issuecomment-1341762950 but putting this in the backlog as it's hard to justify as a priority with limited bandwidth at this stage.

PR #467, opened 2022-09-07
feat: rename multicodec pkg to codecregistry, leaving redirect

rvagg:
Fixes: https://github.com/ipld/go-ipld-prime/issues/259

I'm not sure this is possible since `DefaultRegistry` is not a pointer, so we can't simply provide a redirect for it without breaking code that uses it. I've left the dependent code intact here for now to demonstrate breakage.

If `Registry` were an interface, and/or `DefaultRegistry` were a pointer then this could be cleaner, but as it is I think we can't do it without breaking, so perhaps it's not even worth it and #259 should just be closed.

Thoughts or more creative approaches from anyone?

### DIFF
--- a/codecregistry/defaultRegistry.go
+++ b/codecregistry/defaultRegistry.go
@@ -1,0 +1,117 @@
+package codecregistry
+
+import (
+	"github.com/ipld/go-ipld-prime/codec"
+)
+
+// DefaultRegistry is a codecregistry.Registry instance which is global to the program,
+// and is used as a default set of codecs.
+//
+// Some systems (for example, cidlink.DefaultLinkSystem) will use this default registry,
+// which makes it easier to write programs that pass fewer explicit arguments around.
+// However, these are *only* for default behaviors;
+// variations of functions which allow explicit non-default options should always be available
+// (for example, cidlink also has other LinkSystem constructor functions which accept an explicit codecregistry.Registry,
+// and the LookupEncoder and LookupDecoder functions in any LinkSystem can be replaced).
+//
+// Since this registry is global, mind that there are also some necessary tradeoffs and limitations:
+// It can be difficult to control exactly what's present in this global registry
+// (Libraries may register codecs in this registry as a side-effect of importing, so even transitive dependencies can affect its content!).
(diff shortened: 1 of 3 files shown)


willscott:
Consider making the `canonical` `DefaultRegistry` in `codecregistry` a pointer back to the concrete instance in the multicodec package for now?

BigLep:
2022-11-15: @rvagg is planning to close this, but he will do it.

burdiyan:
No sure whether to post this here or in #259.

I often want to import both IPLD `multicodec` package and multiformat's `multicodec` package. It ends up being very annoying having to alias the package name all the time.

Another thing that is annoying (although orthogonal to this particular issue) is that `github.com/multiformats/go-multicodec` exposes codec values as `type Codec uint64`, but almost everywhere in the ecosystem a simple `uint64` is taken. So, besides having to do named imports, you have convert the types in many places, so overall developer experience becomes pretty clumsy.

IMHO, it's worth putting the effort to resolve the package name issue. The library `go-ipld-prime` is not `v1` yet, so doing deprecation and redirects to some other package should be good-enough solution, even if some small number of other code might break. As far as I understand, things would break only if the actual exported `DefaultRegistry` variable s used somewhere. Using other functions for registering and looking up codecs should work with redirecting to another package.

There has been a number of breaking changes in the IPFS ecosystem of libraries, so I argue that addressing this now might be better than postponing and possibly missing the chance entirely.

rvagg:
That's great feedback @burdiyan but I fear you're right that this is going to take more breakage that I was willing to pull off with this PR. I'll reopen https://github.com/ipld/go-ipld-prime/issues/259 but I'm going to put it on the backlog for now unless (a) there's additional signal that it's worth prioritising, or (b) we get more bandwidth to do things lower down the priority list. Or alternatively someone else, like yourself, could propose a change and help drive it through to completion.

mvdan:
For some historical context, I added the new go-multicodec implementation with the Codec type fairly recently, when the rest of the packages already used `uint64` types. There should be a transition to the named type, but it's also just a nuisance to convert types, so it's hard to argue that we should break dozens of APIs all at once.

PR #473, opened 2022-10-13
Prepare v0.19.0

rvagg:
Probably should have done an 0.19 back when we bumped Go 1.18, but here we are.


BigLep (approved):
Looks good to me @rvagg .  One question: should we update to go 1.19 if we're already doing a go version update?  (Feel free to disregard, just making sure it was intentional to only go to 1.18 currently.)

rvagg:
@BigLep nope, we're going with the IPDX team's choice on this one: https://github.com/protocol/.github/pull/364 - second-to-latest in go.mod, testing for latest and second-to-latest. Will get an auto-bump by unified-ci to 1.19 when 1.20 is released.

PR #477, opened 2022-11-01
gengo: add placeholders for Any

terorie:
### Problem

go-ipldtool hasn't updated its go-ipld-prime dependency in over a year.
This patch fixes go-ipld-prime so it works in go-ipldtool again.

### Description

Fixes gengo compatibility with schemadmt, which generates Any.

Adds placeholders for Any which panic on use at runtime for most functions.

### DIFF
--- a/schema/gen/go/genAny.go
+++ b/schema/gen/go/genAny.go
@@ -1,0 +1,177 @@
+package gengo
+
+import (
+	"io"
+
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/schema"
+)
+
+type anyGenerator struct {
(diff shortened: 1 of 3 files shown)


BigLep:
2023-01-24 maintainer conversation: 
We're juggling priorities at the moment.  @terorie: could you please share more of your usecase that is motivating this work?  We're thankful for you contributing, and just want to know what it's helping unlock for you so we can prioritize accordingly.  Thanks!

terorie:
@BigLep Sorry, I don't remember what I needed this for. I think it was something along the lines of supporting the `Any` type in IPLD schemas. go-ipld-prime partially supports this but not in the `schemadmt` package but not in the code generator.
I don't depend on this, so I'll close for now

PR #480, opened 2022-12-05
Fix hash mismatch error on matching link pointer

masih:
On load, the default IPLD Link System checks the hash of loaded data against the given hash to assure that they match. The actual hash is calculated using the given link's `LinkPrototype.BuildLink`, and then is directly compared with the given link via `==`.

Because the comparison is done by value, if the given link is a pointer the comparison fails which will result in false-positive hash mismatch error.

To address this issue, instead of directly comparing objects using `==` check that their `Binary` value, i.e. their densest possible encoding, is equal. Add tests that assert the issue is fixed and works with existing link representations.

### DIFF
--- a/linking/functions.go
+++ b/linking/functions.go
@@ -11,7 +11,7 @@ import (
 // This file contains all the functions on LinkSystem.
 // These are the helpful, user-facing functions we expect folks to use "most of the time" when loading and storing data.
 
-// Varations:
+// Variations:
 // - Load vs Store vs ComputeLink
 // - Load vs LoadPlusRaw
 // - With or without LinkContext?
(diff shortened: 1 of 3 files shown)


ehsan6sha (approved):

linking/functions.go
```diff
@@ -205,7 +205,7 @@ func (lsys *LinkSystem) Fill(lnkCtx LinkContext, lnk datamodel.Link, na datamode
 	// (Then do a bit of a jig to build a link out of it -- because that's what we do the actual hash equality check on.)
 	hash := hasher.Sum(nil)
 	lnk2 := lnk.Prototype().BuildLink(hash)
-	if lnk2 != lnk {
+	if lnk2.Binary() != lnk.Binary() {
```
willscott, line 208: should this be a `bytes.Equal` check?

masih: How come?
.Binary returns string.

rvagg (approved):
👌

PR #483, opened 2022-12-14
Additional access to schema/dmt package; schema concatenation feature

warpfork:
This is a small diff which introduces a bit more ability to introspect and operate on `schema/dmt` data.

Making the type system accessible makes it easier to manipulate the schema DMT _as IPLD_, which is a very neat thing to be able to do.  (Some of this is already possible because `schemadmt.Type.Schema` is exported specifically; but I see no reason not to export the entire typesystem.)

Then, a function is introduced which can conjoin two parsed schema DMT documents.  (This stops just shy of introducing an "include" syntax to IPLD Schemas, but one can see how it could easily be involved in producing such a feature.)

### DIFF
--- a/schema/dmt/operations.go
+++ b/schema/dmt/operations.go
@@ -1,0 +1,27 @@
+package schemadmt
+
+import (
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+)
+
+// ConcatenateSchemas returns a new schema DMT object containing the
+// type declarations from both.
+//
+// As is usual for DMT form data, there is no check about the validity
+// of the result yet; you'll need to apply `Compile` on the produced value
+// to produce a usable compiled typesystem or to become certain that
+// all references in the DMT are satisfied, etc.
+func ConcatenateSchemas(a, b *Schema) *Schema {
+	// The joy of having an intermediate form that's just regular data model:
+	// we can implement this by simply using data model "copy" operations,
+	// and the result is correct.
+	nb := Type.Schema.NewBuilder()
+	if err := datamodel.Copy(bindnode.Wrap(a, Type.Schema.Type()), nb); err != nil {
+		panic(err)
+	}
+	if err := datamodel.Copy(bindnode.Wrap(b, Type.Schema.Type()), nb); err != nil {
+		panic(err)
+	}
+	return bindnode.Unwrap(nb.Build()).(*Schema)
+}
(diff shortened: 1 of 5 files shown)


schema/dmt/operations.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// ConcatenateSchemas returns a new schema DMT object containing the
+// type declarations from both.
+//
+// As is usual for DMT form data, there is no check about the validity
+// of the result yet; you'll need to apply `Compile` on the produced value
+// to produce a usable compiled typesystem or to become certain that
+// all references in the DMT are satisfied, etc.
+func ConcatenateSchemas(a, b *Schema) *Schema {
```
willscott, line 15: would be great to have a basic test/example so that there's an ongoing validation that concatenation of two basic schemas works

warpfork: In general I'm pro-testing but in this case I don't know what would provide useful coverage in exchange for the additional volume of test code.  We really are just essentially conjoining two maps here.

If there were more interesting and complex logical operations here, I would certainly want to test it, but what this does is so trivial I don't know if it's worth it.

warpfork:
Worth mentioning that I see this as an experiment-enabler moreso than a feature we'd want to spotlight.

It'll be sufficient for me to start playing with what a UX of schemas split across multiple files would feel like, _in a downstream repo_, _without_ pushing any opinionations about that into this repo where they'd become a maintenance commitment.  (Then, I can start taking notes on the experiences that produces, and bring those around for any future discussions about making this a more broadly supported and defined feature of the schema system.)

I'm intentionally refraining from going any further than that at this time.

willscott (approved):
the test would mostly be so that if we change something about the schema interface we don't forget this use in dmt and break it because we've forgotten about it.

that said, i don't feel strongly on that point.

warpfork:
I'll hope that this experiment leads to a more full-fledged feature in the future, and meanwhile, if it breaks, yep, it'll be me who regrets it.  Fair nuff.

rvagg:
@warpfork we're now getting failures while trying to use the latest master because the new doc.go has `package dmt` while the rest are `package schemadmt`. e.g. https://github.com/ipld/codec-fixtures/actions/runs/3761811578/jobs/6393915373

warpfork:
Oof.  That was not from this PR, but was my bad for thinking I could push a docs change without waiting for testing :sweat:   I believe I've fixed it again now.

PR #491, opened 2023-02-10
fix(datamodel): add tests to Copy, make it complain on nil

rvagg:


willscott (approved):

rvagg:
fixed lint errors, will merge and then cut a release after this, there's a bit of a backlog of minor things

PR #493, opened 2023-02-14
Use standard not found error (probably no merge)

hannahhoward:
# Goals

The memstore should return the standard "not found" error for stores that don't have a thing. Sadly, this is currently legacy go-ipld-format ErrNotFound.

# implementation

Import go-ipld-format's ErrNotFound, use it. Obviously, this is terrible.

I think the proper solution is to move that error to this repo, then alias it in go-ipld-format.

Mainly wanted to raise this to provoke the discussion cause I was surprised for find format.IsNotFound not working with memstore in a test. I'll work around.


rvagg:
Yeah, you're right, this is not pleasant and I certainly don't want to pull that in here, especially when there's a good chance it's going to be sucked into the go-libipfs vortex.

But I proposed an alternative in https://github.com/ipld/go-car/blob/master/v2/storage/notfound.go and suggested that the implementation there be brought over here. Instead of using the concrete type we should be doing what the `IsNotFound()` does there:

```go
	if nf, ok := err.(interface{ NotFound() bool }); ok {
		return nf.NotFound()
	}
```

Which would also work for go-block-format—the next step would be to update that error there with that. But it'd technically be a "breaking" change because it's not quite as strict, so it's not perfect and also means anyone using older go-block-format would not match the newer form because of the strictness.

BigLep:
Closing in favor of https://github.com/ipld/go-ipld-prime/pull/494

PR #500, opened 2023-03-07
fix: make StartAtPath work properly for matching walks

rvagg:
The test for this was too easy and doesn't pass if you give it other things to match _prior_ to the `StartAtPath`. New test added for the non-matching case, which is where this functionality is actually used (graphsync), and that works properly already.


hannahhoward (approved):
LGTM, think you've correctly identified the fail

PR #506, opened 2023-03-30
fix: an Any datamodel.Node receiver should accept datamodel.Null

rvagg:
This is a draft because I'm not sure if this is what we want. The main change in behaviour here is in `AssignNull()` where if the field is an `Any` that has a `datamodel.Node` receiver then you should get a `datamodel.Null` set in it. 

The current behaviour always relies on pointers, so right now you'd get a `nil` for your `datamodel.Node` for a `{"foo":null}`. I ran into this and it felt weird, but maybe it's just how it has to be with bindnode since "null" is a special concept anyway? The downside of this is that I can't just take my `Foo` field and pass it back through go-ipld-prime as a `datamodel.Node`, I need an additional `if nil` check on it to operate on a `datamodel.Null` in that case.

Comments welcome. I'll think on this more myself.

### DIFF
--- a/node/bindnode/node.go
+++ b/node/bindnode/node.go
@@ -147,6 +147,16 @@ func actualKind(schemaType schema.Type) datamodel.Kind {
 	return schemaType.TypeKind().ActsLike()
 }
 
+var datamodelNodeIface = reflect.TypeOf((*datamodel.Node)(nil)).Elem()
+
+// for the Any values where we support either *datamodel.Node or datamodel.Node
+func nodeVal(val reflect.Value) datamodel.Node {
+	if !val.Type().Implements(datamodelNodeIface) {
+		val = nonPtrVal(val)
+	}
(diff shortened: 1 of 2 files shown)


node/bindnode/node.go
```diff
@@ -785,15 +795,23 @@ func (w *_assembler) AssignNull() error {
...
+		// if we have a "nullable", then an AssignNull should set the zero value
+		// for the underlying type, if we don't have a "nullable" but we have an
+		// "Any", then we should set the value to datamodel.Null. Otherwise, we
+		// should return an error.
+		if w.nullable {
+			if isAny {
+				w.createNonPtrVal().Set(reflect.ValueOf(datamodel.Null))
+			} else {
+				w.val.Set(reflect.Zero(w.val.Type()))
+			}
```
rvagg, line 807: this is the main changed logic that I'm hesitating on

PR #514, opened 2023-04-20
feat(bindnode): support listpairs struct representation

rvagg:
I want order-preserving structs without a pre-defined strict order (tuple), so this implements listpairs which is in the schema spec but has sparse support through go-ipld-prime.

```ipldsch
type Blop struct {
  foo optional Int
  bar optional String
  baz optional String
} representation listpairs
```

```go
type Blop struct {
  Foo *int
  Bar  *string
  Baz *string
}
```

```json
[ "foo", 100, "baz", "bing" ]
```

_(edit: actually should be this:)_

```json
[ [ "foo", 100 ], [ "baz", "bing" ] ]
```

But also this should decode to the same struct (see `out-of-order` test):

```json
[ "baz", "bing", "foo", 100 ]
```

_(edit: actually should be this:)_

```json
[ [ "baz", "bing" ], [ "foo", 100 ] ]
```

### DIFF
--- a/node/bindnode/repr.go
+++ b/node/bindnode/repr.go
@@ -176,11 +176,9 @@ func (w *_nodeRepr) LookupByIndex(idx int64) (datamodel.Node, error) {
 		return reprNode(v), nil
 	case schema.StructRepresentation_ListPairs:
 		fields := w.schemaType.(*schema.TypeStruct).Fields()
-		if idx < 0 || int(idx) >= len(fields)*2 {
+		if idx < 0 || int(idx) >= len(fields) {
 			return nil, datamodel.ErrNotExists{Segment: datamodel.PathSegmentOfInt(idx)}
 		}
-		isName := idx%2 == 0
-		idx = idx / 2
 		var curField int64
 		for _, field := range fields {
 			value, err := (*_node)(w).LookupByString(field.Name())
@@ -191,10 +189,7 @@ func (w *_nodeRepr) LookupByIndex(idx int64) (datamodel.Node, error) {
 				continue
 			}
 			if curField == idx {
-				if isName {
-					return basicnode.NewString(field.Name()), nil
-				}
-				return reprNode(value), nil
+				return buildListpairsField(basicnode.NewString(field.Name()), value)
 			}
 			curField++
 		}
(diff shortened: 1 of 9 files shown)


node/bindnode/repr.go
```diff
@@ -974,14 +1055,24 @@ func (w *_listStructAssemblerRepr) AssembleValue() datamodel.NodeAssembler {
 		}
 		entryAsm = assemblerRepr(entryAsm)
 		return entryAsm
+	case schema.StructRepresentation_ListPairs:
+		idx := w.nextIndex
+		w.nextIndex++
+		if idx%2 == 0 {
+			asm := (*_structAssembler)(w).AssembleKey()
+			// ???
```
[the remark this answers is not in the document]
rvagg: oh, embarrassing ... that's there cause I don't actually know why I would want to put the next line there; I intended to figure this out

rvagg:
marked as draft cause I just realised after writing out the example above that listpairs is actually `[[k1,v1],[k2,v2]]`, so a list of k/v tuples; so I need to update the repr to look like that!

rvagg:
OK, fixed, this now handles `[[k1,v1],[k2,v2]...]`. It's required a bit more code because of the need for new intermediate assemblers for the inner lists that are strictly length of two, so we have `_listpairsFieldAssemblerRepr` and `_listpairsFieldListAssemblerRepr` to set that up properly.

I also extended the test a bit more so we're not just setting strings as values but also adding a recursive kind on one of the fields, so we end up doing a `[["foo","0"],["bar","1"],["baz","2"],["qux",["3","4"]]]` where the `qux` field is a list of strings.

rvagg:
Made the parser handle "listpairs" for structs in the DSL, including enabling the test fixture that's been waiting to be implemented: https://github.com/ipld/ipld/pull/281

node/bindnode/repr.go
```diff
@@ -992,6 +1076,116 @@ func (w *_listStructAssemblerRepr) ValuePrototype(idx int64) datamodel.NodeProto
...
+func (w _listpairsFieldAssemblerRepr) AssignLink(datamodel.Link) error {
+	return datamodel.ErrWrongKind{
+		TypeName:        w.parent.schemaType.Name(),
+		MethodName:      "AssignLink",
+		AppropriateKind: datamodel.KindSet_JustList,
+		ActualKind:      datamodel.Kind_Map,
+	}
+}
+
+// TODO: support this if it's a list?
```
rvagg, line 1151: turns out I need to do this, so another blocker to merge

rvagg: done

willscott (approved):
🚀

rvagg:
good thing I held off until I had a chance to properly try and use this, it borked on complex nested values—trying to do a listpairs within a listpairs errored out; very tricky to get the "level" right with these builders, representation builders use the schema builders but always need to force them to drop back down to representation level when returning sub-builders to the user

hopefully fixed now, still trying to put it to use

PR #516, opened 2023-04-24
Make traversal.WalkTransforming() work

EdSchouten:
It looks like the current implementation never calls Selector.Decide(), meaning it always attempts to transform the root node, even if it shouldn't be matched.


rvagg:
Thanks for the contribution @EdSchouten! Would you mind having a look at the tests to see if you can either add a new one or adjust an existing one to exercise this?

EdSchouten:
Sure thing! I've just added some basic testing coverage, based on the tests we already have for WalkMatching(). Hope that helps!

rvagg (approved):
👌 beautiful! thanks so much for finding and fixing this!

PR #524, opened 2023-06-13
fix: return error on LookupBySegment failure

rvagg:
Not quite sure what to do with this yet. This probably isn't the "fix", but I need to get this up before I forget the details of how I landed here.

Testing out selectors for our Trusted Gateway walks, I happen to have a DAG where I know it's incomplete—the wikipedia `/wiki/` directory in fact, I don't have many of the pages and I'm expecting an error to be propagated out of the traversal through the sharded directory where it finds it can't load a block. The error occurs in go-unixfsnode/hamt/shardeddir.go's `LookupBySegment()`, and it's a not-found error from the blockstore, but it gets to here and there's a `continue`. i.e. the selector has an interest in this path segment, but it doesn't happen to be in the node in question.

So, our current Trusted Gateway validator (in Lassie) doesn't even reject the malformed DAG because of this. Where I expect a 5 block DAG, but give it a 3, it'll still pass as if everything's fine.

This goes back to the code that started using `Interests()` here: https://github.com/ipld/go-ipld-prime/commit/31c9bb76cc13be01cb2a761f1afe915236e19a36

And it kind of aligns with the stated intention of `Interests()` as a hint, rather than a directive that it _must_ be explored: https://github.com/ipld/go-ipld-prime/blob/f9fb3bd4c6870657dc21d5c917f438ec7784973f/traversal/selector/selector.go#L39-L51

### DIFF
--- a/traversal/walk.go
+++ b/traversal/walk.go
@@ -267,7 +267,7 @@ func (prog Progress) walkAdv(ph phase, n datamodel.Node, s selector.Selector, vi
 	// specific interests, recurse on those.
 	for _, ps := range attn {
 		if v, err := n.LookupBySegment(ps); err != nil {
-			continue
+			return err
 		} else if err := recurse(v, ps); err != nil {
 			return err
 		}


rvagg:
The problem here exists in the nuance between "I don't have that field" (_jedi hand-wave_) and "I have the field but there was an error loading it".

We do have a mechanism for slicing this salami though, `datamodel.ErrNotExists` is an obvious error to check for, and if it's not, then return the error, otherwise continue. The problem is that this is not universally used. For instance we also have a `schema.ErrNoSuchField` that's used for typed nodes, go-unixfsnode will return this for the case of "I don't have that field" in fact. So that kind of sucks.

We could tackle this by doing an error inheritance, or custom `Is()`, or checking for both error types and then strictly stating that a contract for `LookupByX()` should return this error type, so every implementation (basic, typed, ADL) has to conform to get this behaviour. Not a small bugfix though!

hannahhoward:
@rvagg rather than try to ignore the errors when it's not a "loading error", I think what we want is to capture the specific errors that come from loading errors. I think that means wrapping the errors in `loadLink` (https://github.com/ipld/go-ipld-prime/blob/bc536b3f5ceae5be02aac603ee7109c01c17fa0f/traversal/walk.go#L421) with something easily identifiable. That code seems much closer and easier to cover comprehensively.

rvagg:
`loadLink` isn't used by ADLs that load their own nodes, in the particular case I'm hitting of pathing through a sharded directory, the link loading happens in go-unixfsnode and the error comes directly in through `LookupBySegment`, the last link load in the selector code is the block that needs to be reified into a sharded directory.

rvagg:
Some next steps here after a brief discussion:

1. We _should_ be able to handle this problem locally where we care about it (i.e. in our validation and other places we have control and can, and want to, signal errors) by wrapping the linksystem used for traversal. Since refiers are passed a linksystem, any failed loads should be signalled through that so we get a side-channel to block load errors.
2. Investigate the possibility of an `ErrNotExists` contract, around `LookupBy*()` methods. At least something we can `errors.Is()` on. Alternatively we have a list of acceptable error types here to ignore, including currently used types. Any other error type will trigger a fatal in a traversal.

Hannah says that graphsync already does something like (1) above and should be aware of bad blocks in a DAG already, regardless of what the traversal does. That means the blast radius of this problem is less than it might otherwise be.

rvagg:
Workaround in lassie for the two main traversal executions: https://github.com/filecoin-project/lassie/pull/338 - watch for load errors on the linksystem and check for them after we've got no errors from the traversal.

PR #529, opened 2023-06-28
fix(selectors): slice matcher fixes & tests

rvagg:
~~Draft for now; this is fine but I need to make sure there's nothing else needed here for this piece of byte range queries (there's also more pieces).~~

* Don't error slice matcher when not string or bytes, silently non-match
* Cache offset in `readerat` so we don't over-all on the underlying read seeker
* Add lots of tests, including various edges
* Fix 32-bit breakage exposed by tests

### DIFF
--- a/testutil/multibytenode.go
+++ b/testutil/multibytenode.go
@@ -1,0 +1,147 @@
+package testutil
+
+import (
+	"io"
+
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+)
+
+var _ datamodel.Node = MultiByteNode{}
+var _ datamodel.LargeBytesNode = (*MultiByteNode)(nil)
+
+// MultiByteNode is a node that is a concatenation of multiple byte slices.
+// It's not particularly sophisticated but lets us exercise LargeBytesNode in a
+// non-trivial way.
+// The novel behaviour of Read() and Seek() on the AsLargeBytes is similar to
+// that which would be expected from a LBN ADL, such as UnixFS sharded files.
+type MultiByteNode struct {
+	bytes [][]byte
+}
+
+func NewMultiByteNode(bytes ...[]byte) MultiByteNode {
+	return MultiByteNode{bytes: bytes}
+}
+
+func (mbn MultiByteNode) Kind() datamodel.Kind {
+	return datamodel.Kind_Bytes
+}
+
+func (mbn MultiByteNode) AsBytes() ([]byte, error) {
+	ret := make([]byte, 0, mbn.TotalLength())
+	for _, b := range mbn.bytes {
+		ret = append(ret, b...)
+	}
+	return ret, nil
+}
+
+func (mbn MultiByteNode) TotalLength() int {
+	var size int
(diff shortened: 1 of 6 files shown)


rvagg:
I've also removed the hard error in the slice matcher when it reaches a node that's not a string or bytes; this is more consistent with approaches of most of the other selector pieces, they just silently ignore non-matching nodes.

rvagg:
Ready for review now, supports basic byte ranges in Trustless Spec now as this is, sans negative ranges (next PR will add that). Summary:

* Don't error slice matcher when not string or bytes, silently non-match
* Cache offset in `readerat` so we don't over-all on the underlying read seeker
* Add lots of tests, including various edges
* Fix 32-bit breakage exposed by tests

traversal/selector/matcher_test.go
```diff
@@ -0,0 +1,307 @@
...
+
+var _ datamodel.Node = (*MultiByteNode)(nil)
+var _ datamodel.LargeBytesNode = (*MultiByteNode)(nil)
+
+// MultiByteNode is a node that is a concatenation of multiple byte slices.
+// It's not particularly sophisticated but lets us exercise LargeBytesNode as a
+// path through the selectors. The novel behaviour of Read() and Seek() on the
+// AsLargeBytes is similar to that which would be expected from a LBN ADL, such
+// as UnixFS sharded files.
+type MultiByteNode struct {
```
willscott, line 182: maybe move this to a utility file rather than in this test file?

traversal/selector/matcher_util.go
```diff
@@ -1,20 +1,28 @@
...
-// as specified by the ReaderAt interface, and so must only be used
-// in non-concurrent use cases.
-func (r readerat) ReadAt(p []byte, off int64) (n int, err error) {
-	// TODO: consider keeping track of current offset.
-	_, err = r.Seek(off, 0)
+// ReadAt provides the io.ReadAt method over a ReadSeeker. It will track the
+// current offset and seek if necessary.
+func (r *readerat) ReadAt(p []byte, off int64) (n int, err error) {
+	if off != r.off {
+		_, err = r.rs.Seek(off, 0)
```
willscott, line 16: instead of `0` use `io.SeekStart`?

willscott (approved):

hannahhoward (approved):
LGTM

testutil/multibytenode.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (mbnrs *mbnReadSeeker) Read(p []byte) (int, error) {
+	var acc int
+	for _, byts := range mbnrs.node.bytes {
+		if mbnrs.offset-acc >= len(byts) {
+			acc += len(byts)
+			continue
+		}
+		n := copy(p, byts[mbnrs.offset-acc:])
```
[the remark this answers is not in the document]
rvagg: `copy` returns the number of bytes copied which is "the minimum of `len(src)` and `len(dst)`", so it's always going to be capped by `len(p)` in this code and we take the returned `n` as the authoritative offset increment. So I don't believe we're breaking the `Reader` contract here (unless I'm mistaken!).

PR #530, opened 2023-07-04
feat(selectors): negative values for slice matcher's From and To

rvagg:
Ref: https://github.com/ipld/ipld/pull/289

Builds on the basic fixes and tests in: https://github.com/ipld/go-ipld-prime/pull/529

Note particularly the nuances in the test cases for negative values as they relate to what I outlined in https://github.com/ipld/ipld/pull/289.

One implication of the way this is implemented is that we can't do the "your range didn't match, so we'll give you the full file". We end up with "didn't match, you get nothing (just the root)". Which seems like logical behaviour to me, but @willscott you did bring up the http spec saying that you could return the whole lot if your range is OOB or invalid in some way.

### DIFF
--- a/testutil/simplebytes.go
+++ b/testutil/simplebytes.go
@@ -1,0 +1,75 @@
+package testutil
+
+import (
+	"github.com/ipld/go-ipld-prime/datamodel"
+	"github.com/ipld/go-ipld-prime/node/basicnode"
+	"github.com/ipld/go-ipld-prime/node/mixins"
+)
+
+var _ datamodel.Node = simpleBytes(nil)
+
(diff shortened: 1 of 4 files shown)


willscott:
Either behavior can be okay - if we can support the semantics of negative range or otherwise have enough confidence that the range requested doesn't match a portion of the file, not returning anything past the root is i believe the expected semantics (and would result in a render returning an http 416 'range not satisfiable' response)

rvagg:
Made "from" allowed to underflow and be reset to 0. Put comments about the expected behaviour with the boundaries and whether things match or not in the spec docs @ https://github.com/ipld/ipld/pull/289

Issue #531, opened 2023-07-06
datamodel Length() not usable with lazy nodes (e.g. ADLs)


aschmahmann:
Noticed while discovering https://github.com/ipfs/go-unixfsnode/issues/54

The contract on `node.Length()` is
https://github.com/ipld/go-ipld-prime/blob/65bfa53512f2328d19273e471ce4fd6d964055a2/datamodel/node.go#L131-L133

This means that if there is any sort of lazy loading or computation involved in working with a node that is definitely a list or map that there is nothing valid to return.
1. Could return `-1` but that's a contract violation
2. Could return some number (e.g. go-unixfsnode's HAMT returns the largest number of entries it can confirm https://github.com/ipfs/go-unixfsnode/blob/364a5496925b291900d67c35b8b898e1a30b7f0a/hamt/shardeddir.go#L262-L289), but that's also a contract violation since the user expects "the number of entries in a map" not "some number <= the number of entries in a map" and the application has no indication that anything has gone wrong.

I suspect the options here are:
1. Change `Length` to return an error
2. Change the contract to allow `-1` to mean any type of error
3. Add a new function (name TBD) that does return an error and deprecate `Length`

While 2 seems the easiest, it has the ability to break a bunch of code without people noticing. Simple compile-time checked fixes like 1 seem a safer option.

Whether 2 or 3 is preferred seems like a matter of taste.

---

Note: the other functions on the Node interface which could potentially run into a similar issue are `Kind()`, `IsAbsent()`, and `IsNull()` However, it seems like these are less likely to be lazily constructed than `Length` and therefore less likely to be a problem.

Similarly, the map and list iterators can return errors once they're in use (and they can be checked in advance with `Kind()`) and so are also less problematic.

rvagg:
I agree that this is a problem, and it would be good to resolve it without too much of a hack.

I know that during development there was an attempt to reign in the verbosity of the API—it's already very verbose but it could be a lot more by allowing everything to error .. but maybe they should in a world where we're building multi-layered abstractions over these things with complex behaviour underneath.

In terms of options:

> 1. Change `Length` to return an error

I'd prefer this, but it's going to be very painful breaking that contract, with everyone needing to make the jump together. Are we prepared to deal with that pain?

> 2. Change the contract to allow -1 to mean any type of error

This could end up being more dangerous; a length is often used for iteration. Silently changing contract on this could leave users exposed to surprising, or damaging, behaviour. It might be best to be noisy about it.

[...]

PR #532, opened 2023-07-07
feat!: Node#Length() now has an error return

rvagg:
_(I wanted to see how difficult this would be; although the difficulty here doesn't account for the difficulty everywhere else that's going to be hit by such a change)._

Closes: https://github.com/ipld/go-ipld-prime/issues/531

ADL and other Node implementations that need to perform non-trivial work in calculating their Length(), such as by loading multiple child blocks, may encounter errors which should not be silently ignored.


willscott:
Having to check an error every time you want to check an object's length is going to be a pretty painful change. I'd prefer to move that up to the constructor / otherwise have the error surface through existing error channels if possible.

PR #550, opened 2023-09-26
chore: add garbage roundtrip tests for dag-json to check refmt changes/fixes

rvagg:
To confirm that new refmt JSON decoding changes (fixes) don't raise any red flags.

Ref: https://github.com/polydawn/refmt/pull/58

Also includes a new `DeepNodeContentsEquals` quicktest checker that does a `DeepEquals` but doesn't require map keys to be strictly ordered.

Draft until those changes are merged upstream and we're linked to that.


rvagg:
Fix to remt got merged, updated the dependency here

PR #569, opened 2024-08-02
feat(schema): expose useful functions for building schemas

### LINKED ISSUE #568: Incremental schema type building
# What

Enable various methods for building schema type systems more incrementally. 

# Why

Right now, if I have two seperate schema DSL text files, and I want to combine them into one type system, I have to concatenate the bytes for each file and then use IPLD.LoadSchemaBytes()

This is presumably quite slow as it involves a full parse to DMT form followed by a compilation. 

Similarly, if I have a schema.Type that doesn't come from a DSL, for example one I received by using bindnode.Wrap's internal inferSchema, I can't combine into DSL schema that references the type

You can generally imagine this as a schema form problem of a go generic. i.e.

```golang
type MyStruct[T any] struct {
   A string
   B *int
   C T
```
mapping to

```ipldsch
type MyStruct struct {
   A String
   B optional Integer
   C T
}
```

with a second file defining `T` for a specific case:

```
type T struct {
   D String
   E String
}
```

# How

It seems like there are a few changes one could make.

schemadmt.Parse currently adds basic types, and calls ValidateGraph. One could a expose a slightly more low level API that doesn't do all this (instead just accumlating types from a schema DMT)

another option would be to enable merging two type systems. In such a case, duplicates would be accepted as long as the kinds matched. this would be great for being to merge one complex schema.Type into another type system.

hannahhoward:
# Goals

fix #568

# Implementation

add more minimal form of schema compilation and useful functions for merging one type system into another. this would enable the various types of schema type system building I discussed in #568.

Note that most of these functions have been written without access to private data members, in part cause until this is merged I'm just going to build on top. The absolute minimal version of this PR would simply make `spawnType` at https://github.com/ipld/go-ipld-prime/blob/master/schema/dmt/compile.go#L135 a public function `SpawnType` (just to avoid a large copy paste). Pretty much everything else could be built outside the library if you want to target the minimal form.

However, one can see that some of the functions would probably be better with private members. Clone for example is not a full depth clone and only works on types that are in a typesystem, cause many of the public functions of various kinds of types return not the type names but actual types looked up from the TypeSystem.

Anyway, FYC, happy to modify to super minimal (i.e. just spawnType -> SpawnType to make it exported) to building a full incremental builder for type systems. Up to reviewer!

### DIFF
--- a/schema/tmpBuilders.go
+++ b/schema/tmpBuilders.go
@@ -149,6 +149,93 @@ func SpawnEnum(name TypeName, members []string, repr EnumRepresentation) *TypeEn
 	return &TypeEnum{typeBase{name, nil}, members, repr}
 }
 
+// Utility function adding default basic types to schema type system
+func SpawnDefaultBasicTypes(ts *TypeSystem) {
+	ts.Accumulate(SpawnBool("Bool"))
+	ts.Accumulate(SpawnInt("Int"))
+	ts.Accumulate(SpawnFloat("Float"))
+	ts.Accumulate(SpawnString("String"))
+	ts.Accumulate(SpawnBytes("Bytes"))
+
+	ts.Accumulate(SpawnAny("Any"))
+
+	ts.Accumulate(SpawnMap("Map", "String", "Any", false))
+	ts.Accumulate(SpawnList("List", "Any", false))
+
+	// Should be &Any, really.
+	ts.Accumulate(SpawnLink("Link"))
+
+	// TODO: schema package lacks support?
+	// ts.Accumulate(schema.SpawnUnit("Null", NullRepr))
+}
+
+// Clone creates a copy of a type that is not in the original's universe (so it can be used elsewhere safely)
+func Clone(typ Type) Type {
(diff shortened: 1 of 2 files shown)


schema/tmpBuilders.go
```diff
@@ -149,6 +149,93 @@ func SpawnEnum(name TypeName, members []string, repr EnumRepresentation) *TypeEn
...
+		return SpawnFloat(kindedType.Name())
+	case *TypeAny:
+		return SpawnAny(kindedType.Name())
+	case *TypeMap:
+		return SpawnMap(kindedType.Name(),
+			kindedType.KeyType().Name(),
+			kindedType.ValueType().Name(),
+			kindedType.ValueIsNullable())
+	case *TypeList:
+		return SpawnList(kindedType.Name(), kindedType.ValueType().Name(), kindedType.ValueIsNullable())
```
rvagg, line 193: hah, we don't store representation strategy for these even though we have space for it in the dmt, TODO at some point I guess

rvagg (approved):

PR #571, opened 2024-09-01
basicnode: don't panic on negative index

MichaelMure:
I'm not too sure what error should be returned, but it's certainly better than a panic().


node/basicnode/list.go
```diff
@@ -31,6 +31,9 @@ func (plainList) LookupByNode(datamodel.Node) (datamodel.Node, error) {
 	return mixins.List{TypeName: "list"}.LookupByNode(nil)
 }
 func (n *plainList) LookupByIndex(idx int64) (datamodel.Node, error) {
+	if n.Length() < 0 {
```
willscott, line 34: do you want if the length is less than 0 here, or if `idx < 0` ?

MichaelMure: oof of course, fixed it

willscott (approved):

rvagg (approved):

PR #573, opened 2024-09-17
Named type converters

hannahhoward:
# Goals

Currently, the bindnode API enables custom conversions, keyed on go types.

This PR enables a new way to key custom conversions: the type name in a schema.

Why is this useful?
- First it enables using standard go types for custom conversions, rather than needing to create a named go type every time you want a converter
- More interestingly, it enables using interface types within structs, lists, and maps -- since the conversion is keyed on the schema type name, the go type isn't needed to setup a custom conversion, and then the convert can create types that satisfy the interface -- this is super useful for more flexible go structs.

# Implementation

- For compatibility, this keeps all typed custom converter APIs working, with schema named converters taking preference. This does mean that the `config` type must become a struct, and therefore a *config where it is passed
- The schema type name is passed to every call for converterFor & converterForType. Not sure what performance implications this may have (mostly in lookups)
- Wrote a test demonstrating equivalent functioning working with named converters, and a test showing how to use it with interfaces (here used to define a closed union type, something I find is useful to represent often)

### DIFF
--- a/node/bindnode/api.go
+++ b/node/bindnode/api.go
@@ -85,26 +85,38 @@ type converter struct {
 	customToAny   func(interface{}) (datamodel.Node, error)
 }
 
-type config map[reflect.Type]*converter
+type config struct {
+	namedConverters map[schema.TypeName]*converter
+	typeConverters  map[reflect.Type]*converter
+}
 
 // this mainly exists to short-circuit the nonPtrType() call; the `Type()` variant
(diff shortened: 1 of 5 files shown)


rvagg (approved):
cool

rvagg:
windows ci failures are known flaky garbage, `build output "ipldsch_minima.exe" already exists and is not an object file`

PR #574, opened 2025-01-07
Make printer more flexible, allow very-compact forms

rvagg:
I'm attempting here to make something that lets me _see_ the forms of a block without looking at the matrix and without it getting too out of hand when the block is complex.

See printer/printer_test.go for examples of what this is doing. There's some breakage here, however, so @warpfork you might want to weigh in if you still care about the `printer` package:

* Always add commas between list/map items (except for the last one)
* Remove the space between the index/key and the value (so it's always like `foo:string{"bar"}` or `10:string{"bar"}`)
* Don't change `Indentation` if it's left `nil`, it just means there's no indentation, you have to set it. The default package-exposed functions set it to `\t` though.
* If `Indentation` is `nil`, then we also don't print newlines; similar to the behaviour of encoding/json and also the JS `JSON.stringify`.
* Add an `OmitScalarValues` option

As per test example, a basic `Config{OmitScalarValues: true}` would give you the most compact form:

`map{string:string,string:string,string:map{string:string,string:string},string:list{0:int,1:int},string:list{0:float}}`

So far, the most complex thing I've printed with this is a Filecoin HAMT root with embedded [`SectorPreCommitOnChainInfo`](https://github.com/filecoin-project/go-state-types/blob/126d7cfaebc91a8eea739ef6fd28b67a3e903a53/builtin/v16/miner/miner_state.go#L155-L160) (tuple representation) values in it:

`list{0:bytes,1:list{0:list{0:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}},1:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}}},1:list{0:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}},1:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}}},2:list{0:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}}},3:list{0:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}}},4:list{0:list{0:bytes,1:list{0:list{0:int,1:int,2:link,3:int,4:list{},5:int,6:link},1:bytes,2:int}}}}}`

It's still a little bit like looking in the matrix but it's more helpful than printing out CBOR diagnostic output and I'm yet to properly build a set of schemas to match these against.

### DIFF
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -15,21 +15,21 @@ import (
 // All printer configuration will be the default;
 // links will be printed, and will not be traversed.
 func Print(n datamodel.Node) {
-	Config{}.Print(n)
+	Config{Indentation: []byte{'\t'}}.Print(n)
 }
 
 // Sprint returns a textual description of the node tree.
 // All printer configuration will be the default;
 // links will be printed, and will not be traversed.
 func Sprint(n datamodel.Node) string {
-	return Config{}.Sprint(n)
+	return Config{Indentation: []byte{'\t'}}.Sprint(n)
 }
 
 // Fprint accepts an io.Writer to which a textual description of the node tree will be written.
 // All printer configuration will be the default;
 // links will be printed, and will not be traversed.
 func Fprint(w io.Writer, n datamodel.Node) {
-	Config{}.Fprint(w, n)
+	Config{Indentation: []byte{'\t'}}.Fprint(w, n)
 }
 
 // Print emits a textual description of the node tree straight to stdout.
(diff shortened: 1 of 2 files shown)


rvagg:
One adjustment that might be worth making here is adding back the space after `:` when there is indenting, otherwise compress it. I'm fine with this as it is but encoding/json and JSON.stringify also have this behaviour I think.

PR #625, opened 2026-04-23
feat(codec): MaxDepth on DecodeOptions & more bounds testing

rvagg:
* Adds more bounds testing following up from #611.
* Adds a `MaxDepth` field to `dagcbor.DecodeOptions` and `dagjson.DecodeOptions` that caps the nesting depth of decoded structures. Exceeding the limit returns a new `ErrDecodeDepthExceeded` error. Defaults to 1024.


willscott (approved):
useful to be able to trace / prevent loops with this

PR #630, opened 2026-06-01
feat(dagcbor): enforce strict decode defaults

rvagg:
Enable stricter DAG-CBOR decode checks by default using refmt's canonical-form options. Reject indefinite-length CBOR in all modes, and reject non-minimal CBOR heads, NaN, infinity, and duplicate map keys by default.

Add RelaxedDecode as an opt-out for legacy non-canonical inputs while leaving known remaining gaps explicit: map key order and non-64-bit float encodings are still accepted for now.

Update bindnode fuzzing to use relaxed DAG-CBOR decode for generated inputs while retaining canonical re-encode checks.


vmx (approved):
I'm not really a Go coder, but that looks good to me.

